PR #76, opened 2022-10-13
Add ability for API to set name of stream

NickM-27:
We are looking to bundle go2rtc within frigate and use the go2rtc API add the streams. Having the ability to specify the name of the stream will make managing them much simpler.


AlexxIT:
Thanks!

PR #72, opened 2022-10-05
Refactor docker image and ci

felipecrs:
Lots of changes, I don't even remember all, but some are:

1. Reduce image size:

```console
$ docker images
REPOSITORY                                                          TAG              IMAGE ID       CREATED         SIZE
ghcr.io/felipecrs/go2rtc                                            latest           8bc71224401e   4 minutes ago   188MB
alexxit/go2rtc                                                      latest           f3c41b944bd3   6 hours ago     249MB
```

2. Upgrade and pin go version, from 1.18.6 (latest from Alpine repos) to 1.19.2
3. Pin ngrok version to 3.1.0
4. Upgrade base image to the community version and pin it to 9.0.1, which provides a variant which comes with Python preinstalled
5. Build go2rtc from the same revision as the Dockerfile instead of latest main branch.
6. Refactor service call to make it look like the rest of the add-ons, like the SSH and VSCode ones. Let S6-Overlay and add-on watchdog handle restarts.

I'd be happy to clarify any change I made.

PR #88, opened 2022-10-23
Auto-restart go2rtc on error pattern

### LINKED ISSUE #39: There is no retry mechanism for timeouts
There is no retry mechanism for timeouts

I'm through FRP camera port public network, such as RTSP: / / 192.168.4.190:554 / ch01, the port 554 through the public network, and then to public server, link through the server, run for a few hours, if the network fluctuations, or FRP penetration is not stable, lead to broken links, Go2rtc, there is no way to re-link, can only restart GO2RTC

err log

[36murl=[0m"ffmpeg:rtsp://127.0.0.1:54189/ch0#channel=1#subtype=0#audio=opus#unicast=true#resolution=1280x720#video=vp9#raw=-vf transpose=2"
[90m09:25:04.397[0m [1m[31mERR[0m[0m [exec] timeout [36murl=[0m"exec:ffmpeg -hide_banner -allowed_media_types video+audio -fflags nobuffer -flags low_delay -rtsp_transport tcp -i rtsp://127.0.0.1:54189/ch0 -vf transpose=2  -codec:a libopus -ar 48000 -ac 2 -rtsp_transport tcp -f rtsp rtsp://localhost:8554/d40d0930ba209bfe3ee7a339de5253e4"

### LINKED ISSUE #87: Auto Restart without Playback Interruption if Problems
Love this project!

Doing testing at home with my own (Tapo) RTSP cams in preparation for using in real-world non-profit environment with (Swann) RTSP cams.  Noticed that occasional glitches (e.g. error="wrong request") can occur, resulting in current view being frozen.  Easy to workaround by re-entering the view -- but wondering if it is possible to:

- Detect this type of problem and restart go2rtc automatically (the entire app, or somehow just the stream that was affected)
     or
- Restart go2rtc on a regular schedule (e.g. every 5 minutes?) 

...all without interruption of any cameras being viewed, so as not to require intervention for those viewing the streams in a web browser far from the actual server running go2rtc?

felipecrs:
I needed to increase the reliability of go2rtc by removing the need of manually restarting it in such error conditions.

- This PR builds on #72
  - It has to be merged only after #72
- Partially fixes #39 and partially fixes #87
  - We may need to add more error patterns
- This is not a definitive solution. Ideally go2rtc should detect the issue by itself and do the needful to restart the stream without requiring to restart the whole app (which cause the Hass API from being unavailable for some seconds -- probably causing any player to error out)
  - Nevertheless this is better than the current approach, which should help until the definitive fix comes in


felipecrs:
chrome_hd9amuvoUH

felipecrs:
I switched the implementation a little, now the watchdog will only restart the go2rtc service rather than restarting the whole add-on/container, which should make the restart a lot faster.

darethehair:
I don't use Docker myself, so looking forward to this enhancement making it to the regular binaries as well at some point :)

felipecrs:
- This should not be needed anymore since #39 got resolved.

PR #100, opened 2022-11-09
Mention local connection candidate for webrtc

felipecrs:
If my understanding is correct, this is the ideal configuration, am I right?

### DIFF
--- a/README.md
+++ b/README.md
@@ -357,6 +357,7 @@ WebRTC usually works without problems in the local network. But external access 
 webrtc:
   listen: ":8555"  # address of your local server (TCP)
   candidates:
+    - 192.168.1.10:8555 # local server IP-address for local connections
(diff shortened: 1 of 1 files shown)


AlexxIT:
Local IP will be candidate automatically. Check SDP (response) in browser console.

felipecrs:
Also when running as a container without `--network=host`?

AlexxIT:
Only with this setting. Because without it, go2rtc will not know the IP address of the host

felipecrs:
Do you want me to mention in the README that, if you are running as a docker container without `--network=host`, you should specify the candidate manually? Otherwise I can close the PR.

AlexxIT:
I wouldn't recommend anyone to use docker without network host. This is wrong architect decision in my opinion.

felipecrs:
I see. Thank you.

PR #149, opened 2023-01-06
Fix RTSP url in webui

skrashevich:
In case placing go2rtc behind reverse proxy, rtsp url in web-ui was wrong.
Now it fixed by using incoming Host header instead of location.hostname

### DIFF
--- a/cmd/api/api.go
+++ b/cmd/api/api.go
@@ -2,25 +2,33 @@ package api
 
 import (
 	"encoding/json"
+	"net"
+	"net/http"
+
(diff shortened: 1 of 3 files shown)


AlexxIT:
I don't like solution with templates on server. I think it's better to have info REST API. With go2rtc version, current RTSP port, etc. So index.html can request it. Just need to think well about such an API.

skrashevich:
Create API handler for return to client "Host"-Header, send on client side,  looks like a stop-gap solution

AlexxIT:
RTSP address and port fixed in latests version

PR #166, opened 2023-01-13
Fix invalid tcpType for UDP candidate

tsightler:
Current v0.1-rc.8 code added UDP candidates for WebRTC but is hard coded with `tcpType passive` which is invalid.  Spec says tcpType for UDP should be null and some implementations appear to fail on this invalid candidate.  This PR fixes this behavior.


AlexxIT:
Thanks!

PR #167, opened 2023-01-14
Match docs with new webrtc udp fixed port

felipecrs:
@AlexxIT, if I understood correctly the rc8, the docs needs these updates.


AlexxIT:
Thanks! But wiki link is wrong. UDP hole punch. Not UPnP.

felipecrs:
Interesting. Thank you!

Issue #143, opened 2023-01-03
Support environment vars for config


kitchung:
Great project and easy to use!

I'd be wonderful if go2rtc.yaml can support use of:

>1. !include to modularize the config like Home Assistant
>>a. https://www.home-assistant.io/docs/configuration/splitting_configuration/
> 2. secrets.yaml file like Home Assistant
>>a. https://www.home-assistant.io/docs/configuration/secrets/
> 3. env vars like Frigate does with {FRIGATE_RTSP_PASSWORD}
>>a. https://docs.frigate.video/configuration/index#full-configuration-reference

AlexxIT:
I don't think the go2rtc config has to be so complicated to support feature 1 or feature 2.
But, supporting the environment makes sense.

PR #168, opened 2023-01-14
Allow to set webrtc candidates via env

felipecrs:
I know go2rtc does not recommend using something other than `--network host`, but this should help running without `--network host`.

With `--network host`, go2rtc can detect (at some extent), the local IP address of the host machine. At some extent, because if you run it within a VM which is within a NAT, it will not be able to detect the proper IP address.

My use for this will be in the Frigate add-on. I plan to do something like this:

```bash

# Get the local IP address of the HA host machine
local_ip=$(curl -s http://supervisor/core/info | jq -r '.local_ip')
webrtc_port=$(curl -s http://supervisor/addons/self/options | jq -r '.webrtc_port')

export GO2RTC_WEBRTC_CANDIDATES="${local_ip}:${webrtc_port}"

exec go2rtc
```

Which should help simplify this setup process by not requiring the user to set its own list of candidates: https://deploy-preview-4055--frigate-docs.netlify.app/configuration/live#webrtc-extra-configuration

/cc @NickM-27

### DIFF
--- a/cmd/webrtc/webrtc.go
+++ b/cmd/webrtc/webrtc.go
@@ -2,15 +2,18 @@ package webrtc
 
 import (
 	"errors"
+	"io"
+	"net"
+	"net/http"
+	"os"
+	"strings"
+
 	"github.com/AlexxIT/go2rtc/cmd/api"
 	"github.com/AlexxIT/go2rtc/cmd/app"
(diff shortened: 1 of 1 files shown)


AlexxIT:
I think it should be done globally in this issue:
https://github.com/AlexxIT/go2rtc/issues/143

felipecrs:
I totally agree.

PR #180, opened 2023-01-17
Reducing the size of the master-hardware container by half.

skrashevich:
It's essential to verify the absence of functional degradation on various hardware platforms! I have only tested on the platforms that are available to me.

### DIFF
--- a/hardware.Dockerfile
+++ b/hardware.Dockerfile
@@ -1,3 +1,4 @@
+# syntax = docker/dockerfile-upstream:master-labs
 # 0. Prepare images
 # only debian 12 (bookworm) has latest ffmpeg
 ARG DEBIAN_VERSION="bookworm-slim"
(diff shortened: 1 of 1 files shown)


AlexxIT:
I think this is bad idea. Because only one driver for Intel. There is also should be second driver for old devices.
And I can't see divers for AMD.

The idea is that the small container is the main container. Hardware container shouldn't be small. It should work in any cases.

felipecrs:
You also have the option of splitting it in many different containers... as long as we can automate the boring tasks of releasing them automatically (I'm specially talking about the Hass add-on).

AlexxIT:
We need different containers only for hw. There is three types - Intel, AMD, NVidia.
Intel (old and new) works in default alpine container.
AMD (old and new) should work in Hardware. I think there will be a couple of percent of this users.
NVidia cannot be run as an addon because it needs additional docker arguments.

So. No need to create a bunch of containers. They will be difficult to maintain and test.

PR #187, opened 2023-01-20
correct loopback detection on middlewareAuth function

skrashevich:
subj


AlexxIT:
I don't think this is necessary, because `IsLoopback()` do same thing but with more calculations inside.

Issue #183, opened 2023-01-18
Allow to disable stream from API (maybe also from UI/YAML config as well)


felipecrs:
I believe this would be a graceful solution for an issue in Frigate:

- https://github.com/blakeblackshear/frigate/issues/4325

The use case is explained in details there, but basically, I want to be able to automate the availability of the stream using some (Home Assistant) automation for a battery based camera/doorbell.

If I just pass its RTSP stream to Frigate, Frigate will maintain a connection to it forever. With go2rtc in the middle, I could enable the go2rtc stream programmatically through some HA automation (like an external motion sensor, or a doorbell button press sensor, or anything else), then disable it automatically after some minutes.

AlexxIT:
I think this is the wrong way to solve the problem. It is up to the end app (Frigate) to request or not to request the stream.

NickM-27:
agreed, frigate would still be requesting the stream which is still a problem. Frigate should know to not request the stream in the first place

felipecrs:
But take Frigate out of the loop, and I still believe this would be useful.

Any client can be consuming your stream, like VLC, a different NVR, or anything. This allows the stream to be blocked from the administrator side for whatever reason, including the battery powered cameras use case that I mentioned but not limited to that. If it's an internal camera for example, I could use this feature to disable the stream when I'm home.

felipecrs:
> I can drop all current connections when someone calls delete API for stream. It could resolve your problem. You will need to add and remove stream via API.

That's true, I'm aware of this possibility. But I don't like the fact that I'd have to provide the connection details every time I want to enable the stream back. I'm not sure if I can set the connection details just like I could from the YAML config. For example, can I add both a `echo:` source and another `ffmpeg:` source?

felipecrs:
> I can drop all current connections when someone calls delete API for stream. It could resolve your problem. You will need to add and remove stream via API.

Oh wait, I thought this would be happening already. If not, then yes, I think it would make sense as well, regardless of this feature suggestion.

AlexxIT:
I can extend the stream creation API, and take json as input.
If you delete stream - it won't drop current connections. Just remove stream from list.

felipecrs:
Alright... so, with these two changes, I'd understand if you close this issue. But I still think it would be nice to have a _disabling/enabling_ abstraction for users. :)

PR #233, opened 2023-02-01
api/streams: added persistency

kottle:
This code updates  streams config file when a  stream is created/deleted with api/streams.

### DIFF
--- a/cmd/api/config.go
+++ b/cmd/api/config.go
@@ -1,11 +1,12 @@
 package api
 
 import (
-	"github.com/AlexxIT/go2rtc/cmd/app"
-	"gopkg.in/yaml.v3"
 	"io"
 	"net/http"
 	"os"
+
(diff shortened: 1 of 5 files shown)


AlexxIT:
This code totally break my existing YAML config. I think streams should be added to go2rtc.json file, like HomeKit cameras. And go2rtc shouldn't change YAML config from code. Because there is no safe way to do it.

felipecrs:
ESPHome also does something similar, they take YAML files as input and generate JSON files to have them persisted.

felipecrs:
This feature would also come in handy for things like:

- https://github.com/AlexxIT/go2rtc/issues/183

So that streams that gets disabled from API/UI are remembered if go2rtc restarts.

AlexxIT:
I think it's possible to store dynamic part inside `go2rtc.yaml`, but it should be special single line in text format. It will be easy to find and replace. For example:

```yaml
db: '{...}'
```

PR #269, opened 2023-02-20
Add specific note about ubiquiti cameras

NickM-27:
I think this would be helpful to have it noted in the docs so we can point users at this when they are using go2rtc with ubiquiti cameras. Feel free to reword or add more info


AlexxIT:
I have this note already in the RTSP section

NickM-27:
> I have this note already in the RTSP section

Ah whoops I missed that, that is perfect then.

PR #273, opened 2023-02-26
Fix double url-encoding in streams DELETE request

skrashevich:
Due to double encoding, the removal of streams whose names contained special characters didn't works

Example 
Before: /api/streams?src=http%253A%252F%252F192.168.88.158%252Fsecretkey%252Flive%252Ffiles%252Fhigh%252Findex.m3u8
After: /api/streams?src=http%3A%2F%2F192.168.88.158%2Fsecretkey%2Flive%2Ffiles%2Fhigh%2Findex.m3u8

### DIFF
--- a/www/index.html
+++ b/www/index.html
@@ -125,7 +125,8 @@
         ev.preventDefault();
 
         const url = new URL("api/streams", location.href);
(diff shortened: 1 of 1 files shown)


AlexxIT:
Thanks!

PR #274, opened 2023-02-27
fix exit code on config save

skrashevich:
... to handle "restart: on-failure"

### DIFF
--- a/www/editor.html
+++ b/www/editor.html
@@ -42,7 +42,7 @@
         }).then(r => {
             if (r.ok) {
                 alert('OK');
-                fetch('api/exit', {method: 'POST'});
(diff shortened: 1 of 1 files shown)


felipecrs:
I have to say I'm against this. Usually services exits with 0 when they signal a graceful restart.

I'd say your docker compose config is wrong. It should be: `always` or `unless-stopped`.

AlexxIT:
Thanks! This make sense. But I'll change code to 100. Hass used this code for handeling restart in Windows. Also code 1 should remain a real error.

https://tldp.org/LDP/abs/html/exitcodes.html

AlexxIT:

PR #280, opened 2023-03-09
Add HW ROTATION support to vaapi engine

horttorrell32:
This change allows to use the transpose_vaapi HW ACC to perform the rotation.


AlexxIT:
Thanks!

PR #332, opened 2023-04-04
Fix ISAPI 2 Way Audio

acardace:
While testing my HikVision doorbell for 2 way audio I noticed that it wasn't working as the isapi stream would never get opened, this patch fixes it by not letting go2rtc close the producer if senders are present as well.

I don't know if this fix is correct but I tested it on my setup and it opens and closes the isapi "session" correctly.

### DIFF
--- a/cmd/streams/stream.go
+++ b/cmd/streams/stream.go
@@ -185,6 +185,11 @@ producers:
 				continue producers
 			}
 		}
+		for _, track := range producer.senders {
+			if len(track.Senders()) > 0 {
+				continue producers
+			}
+		}
 		producer.stop()
 	}
 	s.mu.Unlock()


AlexxIT:
Thanks! I'm also don't know if this fix correct. But we will se.

AlexxIT:
https://github.com/AlexxIT/go2rtc/releases/tag/v1.3.2

pergolafabio:
Hey @acardace @AlexxIT  , sorry to bump into this.. 
I use this isapi already for a long time, I was even the beta tester :-)

Anyway, what I notice, I use the frigate card to toggle two way audio with the mic button, it works perfectly!!

My hikvision device is also an intercom, so what I do, on an incoming call, I send an curl command to the hikvision device to stop the ringing, that way I can start two way audio... This works...

But it only works for 1 time, on the next call , info the same curl, but I'm unable to open the two way audio...

I think the incoming call kills the isapi audio data, I think frigate/go2rtc still believes the audio session is still open, that's why I can't send audio anymore...

I need to refresh the page with the Lovelace card to make it work again... Probably then refresh opens a new audio channel?

Can this be fixed somehow?

AlexxIT:
I don't know how Frigate card works. WebRTC card allow you to setup different stream for mic
https://github.com/AlexxIT/WebRTC#two-way-audio

pergolafabio:
yeah, i now tested the webrtc card, and that does work, switching modes always provides me the two way audio
However switching the mic button on the frigate card (same as swithcing mode) doesnt start a new session

My guesst; when you switch modes on the webrtc card, it does a full refresh of the card maybe? starting a new audio session?

AlexxIT:
Switching stream in the WebRTC card starts new session. This is made for fixing your issue.

Issue #338, opened 2023-04-07
Panic when accessing /api/stream.mp4 with both video and audio tracks enabled


yousong:
I am wondering maybe it's caused by http.ResponseWriter being written concurrently from multiple goroutine.

The source is a `ffmpeg:rtsp://.../stream1` with h264 video and pcma audio track.
- The issue happens when accessing go2rtc `/api/stream.mp4?src=tl_ipc44gw_main&video=h264,h265&audio=aac,opus,mp3,pcma,pcmu`
- No issue when accessing `/api/stream.mp4?src=tl_ipc44gw_main`.  It seems in this case go2rtc defaults to audio=aac and the resulting mp4 stream is has only a video track present.

go2rtc runs as a armv7 container image (alexxit/go2rtc:master, alexxit/go2rtc@sha256:5e890220e10a482bf25032a99d86a3ab441a7def3a7d218aec92f1e0f57f8333).

The issue exhibits itself in two ways in the log.

1. bufio slice bounts out of range.
```
panic: runtime error: slice bounds out of range [:2964] with capacity 2048

goroutine 519 [running]:
bufio.(*Writer).Flush(0x2572700)
        bufio/bufio.go:628 +0x1ac
bufio.(*Writer).Write(0x2572700, {0x3884800, 0x46c, 0x800})
        bufio/bufio.go:672 +0xdc
net/http.(*response).write(0x106a280, 0x46c, {0x3884800, 0x46c, 0x800}, {0x0, 0x0})
        net/http/server.go:1636 +0x214
net/http.(*response).Write(0x106a280, {0x3884800, 0x46c, 0x800})
        net/http/server.go:1594 +0x48
github.com/AlexxIT/go2rtc/cmd/mp4.handlerMP4.func1({0x740e48, 0x256ca50})
        github.com/AlexxIT/go2rtc/cmd/mp4/mp4.go:116 +0x6c
github.com/AlexxIT/go2rtc/pkg/core.(*Listener).Fire(...)
        github.com/AlexxIT/go2rtc/pkg/core/listener.go:16
github.com/AlexxIT/go2rtc/pkg/mp4.(*Consumer).AddTrack.func4(0x2972b40)
        github.com/AlexxIT/go2rtc/pkg/mp4/consumer.go:133 +0xa4
github.com/AlexxIT/go2rtc/pkg/core.(*Sender).HandleRTP.func1()
        github.com/AlexxIT/go2rtc/pkg/core/track.go:139 +0x64
[...]

PR #350, opened 2023-04-14
mp4: do not use ResponseWriter when exiting handler

yousong:
It seems net/http will deinit w on handler return and using it at that time will cause invalid memory access panic.  This can be reproduced when doing page refresh

Ref: https://github.com/AlexxIT/go2rtc/issues/338


AlexxIT:
Fixed with another way

PR #394, opened 2023-04-29
Automate dependencies updates for GitHub Actions

felipecrs:


AlexxIT:
Thanks. I'll better do it manually. Because new versions can break something.

felipecrs:
@AlexxIT, this only creates the PR for you, it does not mean it will be merged automatically. It will always require your review.

PR #414, opened 2023-05-04
Update hardware.Dockerfile for multi-platform support

skrashevich:
...because we can build from hardware.Dockerfile on non-amd64 host, such as Apple M1

### DIFF
--- a/hardware.Dockerfile
+++ b/hardware.Dockerfile
@@ -12,7 +12,13 @@ FROM ngrok/ngrok:${NGROK_VERSION} AS ngrok
 
 
 # 1. Build go2rtc binary
-FROM go AS build
+FROM --platform=$BUILDPLATFORM go AS build
+ARG TARGETPLATFORM
+ARG TARGETOS
+ARG TARGETARCH
+
+ENV GOOS=${TARGETOS}
+ENV GOARCH=${TARGETARCH}
 
 WORKDIR /build
 


AlexxIT:
What's the point to run this image on M1?
This image with drivers for Intel, AMD and Nvidia. They useless on M1

skrashevich:
this command will build go2rtc-hardware container for generic amd64 platform on your arm64 MacBook (for example):
`docker buildx build --platforms=linux/amd64 -t alexxit/go2rtc:master-hardware -f hardware.Dockerfile --push .`
and then you can use it where you want, on any amd64 system

AlexxIT:
What about?

```
ARG TARGETPLATFORM
ARG TARGETOS
ARG TARGETARCH
```

Looks like `TARGETPLATFORM` not used. And `TARGETOS` and `TARGETARCH` not set anywhere.

skrashevich:
> And `TARGETOS` and `TARGETARCH` not set anywhere.

they set by buildkit (it's integrated args, like BUILDPLATFORM

> Looks like `TARGETPLATFORM` not used

Yes, really

AlexxIT:
OK. I hope this doesn't break anything.

skrashevich:
> OK. I hope this doesn't break anything.

I'm too

AlexxIT:

PR #418, opened 2023-05-05
Add trailing slash handling to HandleFunc in api.go

skrashevich:
This pull request enhances the HandleFunc function in api.go by adding basePath handling for pattern strings without a leading slash and trimming trailing slashes, ensuring consistent and reliable pattern registration.


AlexxIT:
This will break API with tailing slash and should be reworked
https://github.com/AlexxIT/go2rtc/blob/e4c7ffd1b40c6745ac80576b615b947705b5fa58/internal/onvif/init.go#L26

PR #435, opened 2023-05-15
Fix AVCtoAnnexB function

SouL-H:


AlexxIT:
Did you realize that `bytes.Clone` does the same thing? :)

SouL-H:
> Did you realize that `bytes.Clone` does the same thing? :)

This command was not supported in version go 1.19. I needed to fix the code because I got an error while compiling. Then I realized that it was a feature that came after 1.20, so I canceled it :)

PR #440, opened 2023-05-21
Add support for daemonization

skrashevich:


AlexxIT:
Lots of code. It needs to be carefully analyzed all of it.

skrashevich:
> Lots of code. It needs to be carefully analyzed all of it.

The Way of the Maintainer :)

skrashevich:
Up 🙄

Nelson-Levi:
I noticed this change would allow the program to run as a backround process. Would this allow the program to permanently stream two-way audio output without user input?

felipecrs:
This doesn't introduce anything new. You can already achieve that with docker or systemd. go2rtc is meant to be kept executing anyway (as a service), not as a one-off execution.

PR #449, opened 2023-05-25
Added auto-reconnect to webrtc.html

miguelangel-nubla:


AlexxIT:
This is pointless. WS connection may be closed right after webrtc connection established. And it's better to listen PC closed event.

`stream.html` support reconnection feature for webrtc.

PR #469, opened 2023-06-10
fix bug :ws service not load origin from go2rtc.yml

galindocode:
fix possible bug when ws api service not load config from go2rtc.yaml and not load origin


AlexxIT:

PR #471, opened 2023-06-12
Fix camera brand typo in README.md

makuser:
Change "Unify" to "Ubiquiti UniFi", so that users using Ctrl+F or the search function to find information they are looking for, actually have a chance to find it.


AlexxIT:
Thanks. That makes sense.

PR #505, opened 2023-07-05
escape shell args in queryToInput && implement some tests

skrashevich:


AlexxIT:
I don't see the shell escape making much sense

PR #528, opened 2023-07-16
Add support for pure-go ngrok tunneling in the API 

skrashevich:


felipecrs:
Maybe you can remove ngrok from the Dockerfile now?

AlexxIT:
This is wrong realisation with proxing only API port. But idea was about proxing WebRTC port.
Also I need to check go2rtc binary size change.
And including so many external dependencies for so not popular feature - not a good idea.
I need to think about this PR.

skrashevich:
> Also I need to check go2rtc binary size change.

+2M

PR #611, opened 2023-09-04
Clarify import from go2rtc to hass generic camera

felipecrs:
The protocol is not set by default. According to my tests, only TCP works.


AlexxIT:
This is not necessary option. Important not to select UDP.

felipecrs:
You are right, I just tested here. Ok, I updated the PR. Can you take a look again, @AlexxIT?

AlexxIT:
Don't think this really necessary, but anyway

felipecrs:
Ok, sorry. I could have closed the PR too. Have a great day/night anyway.

Issue #626, opened 2023-09-10
Fix support HomeKit server on macOS


OskarLebuda:
Hi, 
I’m not familiar with this lib so let me know if there are some answers for my questions 

I found that I can use my homekit cameras (Aqara G2H Pro) into my custom app using this lib. But all of the docs told me that I have to known the `homekit://` url. But... There is no information about "How to get `homekit://` url".

Can someone tell me what should I do to get stream from my Aqara cameras into Go2RTC? 
What should I do?
How to configure it?

I've tried to use docker image on mac to add my other camera (Eufy indoor cam) and I've got an error:
![image](https://github.com/AlexxIT/go2rtc/assets/10373127/6e1bd552-9971-4f6e-802c-8a9b5f6d896e)

my config:
```yml
api:
  origin: "*"
streams:
  eufy1:
    - rtsp://username:password@192.168.1.33/live0
    - ffmpeg:eufy1#video=h264#hardware
    - ffmpeg:eufy1#audio=opus
homekit:
  eufy1:
    pin: 19550224
    name: Eufy cam
    device_id: eufy1
    device_private: eufy1
```

I really don't know if I doing it right so let me know. 

Another problem is that docker on mac does not support `network_mode: host` so I proxy ports like:
```yml
version: "3.7"

services:
  go2rtc:
    container_name: cube-home-go2rtc
    image: alexxit/go2rtc:1.7.1
    privileged: true
    restart: unless-stopped
    ports:
      - 1984:1984
      - 8554:8554
      - 8555:8555
    environment:
      - TZ=Europe/Berlin
    volumes:
      - "./docker/go2rtc:/config"
```

Let me know if it's a problem for this lib

AlexxIT:
1. I'm not sure if HomeKit will work fine without network host.
2. You should pair your camera with Home Assistant or go2rtc to get homekit link
3. What exact model of your eufy camera?

AlexxIT:
You can run go2rtc without docker

AlexxIT:
Interesting. I have another problem on my mac
```
20:53:16.858 INF go2rtc version 1.7.1 darwin/arm64
20:53:16.859 INF [api] listen addr=0.0.0.0:1984
20:53:16.859 INF [rtsp] listen addr=0.0.0.0:8554
20:53:16.859 INF [webrtc] listen addr=0.0.0.0:8555/tcp
[...]

PR #634, opened 2023-09-11
BugFix: mdns listener on macOS 

skrashevich:
Fix for #626

### DIFF
--- a/pkg/mdns/client.go
+++ b/pkg/mdns/client.go
@@ -6,6 +6,7 @@ import (
 	"errors"
 	"fmt"
 	"net"
+	"runtime"
 	"strings"
 	"syscall"
 	"time"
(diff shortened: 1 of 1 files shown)


OskarLebuda:
@AlexxIT will this fix our problem with macOS binary?

AlexxIT:
@OskarLebuda maybe. But it will break linux, because it wrong code for this OS :)

skrashevich:
@AlexxIT I don't agree. This syscall is supported on any posix-compatible system, including, of course, linux

AlexxIT:
Some linux versions has another REUSEPORT constant. Some doesn't has it at all
https://stackoverflow.com/questions/14388706/how-do-so-reuseaddr-and-so-reuseport-differ/14388707

AlexxIT:
Thanks. REUSEPORT added with another way, only for macOS

PR #661, opened 2023-09-30
WebUI Editor: load ace package from unpkg instead of cloudflare

skrashevich:


AlexxIT:
https://github.com/AlexxIT/go2rtc/releases/tag/v1.8.0

PR #691, opened 2023-10-14
fix display of Version in web ui

hgy59:
### Description

The web ui displays an empty Version string like `Version: , Config: /volume1/@appdata/go2rtc/go2rtc.yaml` on the default web page.

reported at https://github.com/SynoCommunity/spksrc/pull/5889#issuecomment-1763186306

### Details
- fix display of ${data.version} in www/index.html
- Info["version"] must be initialized when Version is set

### DIFF
--- a/internal/app/app.go
+++ b/internal/app/app.go
@@ -38,6 +38,8 @@ func Init() {
 		os.Exit(0)
 	}
 
+	Info["version"] = Version
+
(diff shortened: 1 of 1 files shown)


hgy59:
fixed in e367940bd921cf06a5695e636e454e98ed3b3ffb

PR #702, opened 2023-10-18
Add support for L16 codec on backchannel

oNaiPs:


AlexxIT:
I don't think it's a good idea to change the receiver. In all other cases receiver produce codec as is.
There should be a pipeline between the receiver and the sender that converts the codec.

PR #719, opened 2023-10-28
Add ISAPI example for Hikvision NVR

Mincka:
To use ISAPI through NVR, you need to specify the whole URL in the configuration to set the proper channel.
I found it by looking at the code but it could be helpful to update the doc for that.

### DIFF
--- a/README.md
+++ b/README.md
@@ -565,9 +565,14 @@ This source type support only backchannel audio for Hikvision ISAPI protocol. So
 
 ```yaml
 streams:
+  # Directly on camera
   hikvision1:
     - rtsp://admin:password@192.168.1.123:554/Streaming/Channels/101
     - isapi://admin:password@192.168.1.123:80/
+  # Camera through NVR (ex. main stream on channel 7)
+  hikvision2:
+    - rtsp://admin:password@192.168.1.123/Streaming/Channels/701
+    - isapi://admin:password@192.168.1.123:80/ISAPI/System/TwoWayAudio/channels/701
 ```
 
 #### Source: Nest


AlexxIT:
I don't know how that works for you. Because this line:

https://github.com/AlexxIT/go2rtc/blob/b4ed738d17183d431fc1c23ec1d0c312c1875af7/pkg/isapi/client.go#L39

Mincka:
Thank you for the follow-up.

I'll do further testing tonight and indeed check what's the behavior of the NVR if I follow inconsistent URLs such as: `/ISAPI/System/TwoWayAudio/channels/701/ISAPI/System/TwoWayAudio/channels`

Anyhow, rewrite of client.go to natively support NVR would be more beneficial.

Maybe it could be "simplified" and made more generic by just allowing/requiring the users to specify the full ISAPI URL they wish to use. This would be more explicit and could open the card to additional scenarios like combining the ISAPI audio of one camera, (or a specific audio input of a camera) with any video feed.

AlexxIT:
No problem. But I need to know the exact working links

Mincka:
I really don't get what I did.
Now I get the sound without specifying ISAPI...
Direct on camera for main, sub and third streams with basic setup like this:
```
streams:
  hikvisionGarage:
    - rtsp://xxx:yyy@10.0.0.141/Streaming/channels/101
```
Through the NVR:
```
  hikvisionJardinArriereMarche:
    - rtsp://xxx:yyy@10.0.0.71/Streaming/channels/302
```
* Channel 1: _The type provided ('video/mp4; codecs="hvc1.1.6.L153.B0,flac"') is unsupported._ Works fine on VLC with sound.
* Channel 2: No sound. Also no sound on VLC.
* Channel 3: Not accessible through NVR

I think that my setup just worked without ISAPI URL, so using a wrong was pointless and useless.
I guess I have another issue on the NVR with the missing sound on the sub-stream.

I close this PR since it's wrong anyway.

PR #730, opened 2023-11-05
Refactor device ID calculation in `calcDeviceID` function

skrashevich:
... to generate a MAC address with locally administered and unicast bits set using every second byte from the SHA512 hash of the seed

### DIFF
--- a/internal/homekit/server.go
+++ b/internal/homekit/server.go
@@ -246,8 +246,18 @@ func calcDeviceID(deviceID, seed string) string {
 		// 2. Use device_id as seed if not zero
 		seed = deviceID
 	}
+	var macParts []string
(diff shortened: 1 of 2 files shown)


AlexxIT:
Why we needs this? Just for fun? Any changes to deviceid will break all current homekit setups.

skrashevich:
> Any changes to deviceid will break all current homekit setups.

it will not break current setups because of
image

skrashevich:
> Why we needs this?

Incorrect mac addresses (for example, from the multicast range) will cause difficult-diagnosable and hard-reproducible errors in some network environments

AlexxIT:
This will break setup for all users who not set device_id in yaml or used short device_id.

PR #759, opened 2023-11-17
fix: updating ngrok readme

russorat:
PM from ngrok here. This is just a quick update for the ngrok section of the readme to link to the ngrok docs and to lowercase the name ngrok. thanks for the cool project!


AlexxIT:
Thanks! Your project is very cool too. It helps a lot in development and testing.

I have a PR about implementing ngrok inside go2rtc https://github.com/AlexxIT/go2rtc/pull/528. I like this idea, but ngrok has a lot of dependencies. And increase binary size from 500KB to 2MB. I want to keep the project as light as possible.

Maybe your team has some lite code for implement proxy only for TCP socket?

PR #785, opened 2023-11-29
Ensure exit code is within valid range

skrashevich:
The exitHandler function now properly validates the exit code provided in the query string. It checks for conversion errors and ensures the code is within the valid range of 0 to 125. If the validation fails, it responds with an HTTP 400 Bad Request error. This prevents potential misuse of the exit endpoint by restricting the exit codes to expected values.


AlexxIT:
Thanks!

PR #798, opened 2023-12-04
Fix CORS support in WHEP/WHIP API

MPTres:
Browsers don't allow WHIP/WHEP API calls  to the go2rtc server when they are directed to  a different domain, even when all origins are allowed in go2rtc configuration.
```
api:
  origin: "*"  
```

The problem it that when browsers detect a POST to another domain, they usually send first  a _Preflight Request_. This is an OPTIONS request  to verify with the server if that's  allowed. The server needs to reply with a 'No content' status , and some headers with information about what's allowed.  Right now, go2rtc just returns a "Method not allowed" response, which prevents the browser from sending the actual POST.

This change adds an OPTIONS handler and some allowed headers, so that the cors WHEP/WHIP requests can be accepted by the browser.

### DIFF
--- a/internal/api/api.go
+++ b/internal/api/api.go
@@ -211,7 +211,7 @@ func middlewareCORS(next http.Handler) http.Handler {
 	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
 		w.Header().Set("Access-Control-Allow-Origin", "*")
 		w.Header().Set("Access-Control-Allow-Methods", "GET, POST, PUT, DELETE, OPTIONS")
-		w.Header().Set("Access-Control-Allow-Headers", "Authorization, X-PINGOTHER, Content-Type")
+		w.Header().Set("Access-Control-Allow-Headers", "Authorization, Content-Type")
 
 
 		next.ServeHTTP(w, r)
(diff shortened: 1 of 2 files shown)


AlexxIT:
There you get `X-PINGOTHER` header?

skrashevich:
it's example of non-standard header from Mozilla docs :)

MPTres:
> `X-PINGOTHER`

i saw that on the reference docs in mozilla,  and added it too. I see now that  it's not a standard, just an example. I should remove that.  The content-type is the important one to allow contents other that plain text in the request.

AlexxIT:
@MPTres also you have added unnecessary empty line after new header

MPTres:
> @MPTres also you have added unnecessary empty line after new header

Fixed

skrashevich:
> The content-type is the important one in order to allow json or sdp.
nope
https://developer.mozilla.org/en-US/docs/Glossary/CORS-safelisted_request_header

MPTres:
> > The content-type is the important one in order to allow json or sdp.
> 
> nope https://developer.mozilla.org/en-US/docs/Glossary/CORS-safelisted_request_header

that says that the safelisted content values are: `application/x-www-form-urlencoded, multipart/form-data, or text/plain`
That is consistent with my tests with json and sdp types, where the browser refused the connection if the server did not explicitly allow the content header to be sent .

Also, this is a quick fix  (but somehow consistent with the server just allowing everything from cors "*" ). The proper way would be checking the values received in `Access-Control-Request-Headers` and allow or reject the OPTIONS request based on that.

PR #811, opened 2023-12-08
ffmpeg binary validation

MPTres:
Adds a validation to check if the configured custom ffmpeg binary actually exists in the file system.
If the binary does not exists, logs an error message and closes the app.

A small convenience to find early a typo in the config or a missing ffmpeg file, instead of waiting for a stream request to arrive to detect the problem.

### DIFF
--- a/internal/ffmpeg/ffmpeg.go
+++ b/internal/ffmpeg/ffmpeg.go
@@ -2,6 +2,7 @@ package ffmpeg
 
 import (
 	"net/url"
+	"os"
 	"strings"
 
 	"github.com/AlexxIT/go2rtc/internal/app"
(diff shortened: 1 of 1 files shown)


AlexxIT:
1. Bad idea with `os.Exit` because ffmpeg just an option. go2rtc can do a lot without it.
2. Won't work with env `PATH`, for example ffmpeg with different name...

MPTres:
> 1. Bad idea with `os.Exit` because ffmpeg just an option. go2rtc can do a lot without it.
> 2. Won't work with env `PATH`, for example ffmpeg with different name...

good points. There are other uses cases i missed.

Issue #667, opened 2023-10-03
Choppy camera sound in HomeKit


VladNord:
I have Hikvision camera with built-in microphone. 
When I watch the stream directly from go2rtc- flawless, but exported to HomeKit camera has choppy sound, picture is ok
Sound codec g711ulaw (tried another- no luck)

Here is config:

Thanks for help!

```
streams:
  home_cam_outdoor:
    - rtsp:/*****/ISAPI/Streaming/Channels/101
    - ffmpeg:home_cam_outdoor#audio=opus  

homekit:
  home_cam_outdoor:            
    pin: 12345678         
    name: home_cam_outdoor      
```

AlexxIT:
Known problem

civita:
It seems like this is related to incorrect timestamp increment (ts incr). According to [RFC 7587](https://datatracker.ietf.org/doc/html/rfc7587), ts incr of a 48k sampling rates and 20ms frame size will be `960`, or ts incr = `frame size(ms) * sampling rate(k)` I'm adding this into `srtp/session.go`. However, depends on connection type of iOS devices (Cellular/Wi-Fi), frame size will be other than 20 (60 on Cellular connection). Mismatched frame size results in no sound.

civita:
Update: I am able to get HomeKit Opus Audio working on Cellular connection (frame size 60ms) by repacketize three 20ms frames into one packet (with some hard-coded blocks). It would be better if ffmpeg args can be modified after `SelectedStreamConfig` is retrieved from iOS devices.

AlexxIT:
As said in the RFC 7587 - your timestamp wrong. But maybe Apple also doesn't respect this RFC.

```
   Opus supports 5 different audio bandwidths, which can be adjusted
   during a stream.  The RTP timestamp is incremented with a 48000 Hz
   clock rate for all modes of Opus and all sampling rates.  The unit
   for the timestamp is samples per single (mono) channel.  The RTP
   timestamp corresponds to the sample time of the first encoded sample
   in the encoded frame.  For data encoded with sampling rates other
   than 48000 Hz, the sampling rate has to be adjusted to 48000 Hz.
```

civita:
[...]

PR #843, opened 2023-12-24
Fix choppy audio sounds when opus is used with HomeKit as a consumer

civita:
HomeKit cameras have choppy sounds (https://github.com/AlexxIT/go2rtc/issues/667) in Wi-Fi and no sound in cellular network. This is due to incorrect timestamp increment and mismatched frame sizes.

It looks like iOS devices only request either 20ms (Wi-Fi) or 60ms (Cellular) opus frame sizes, and because I don't know how to dynamically modify ffmpeg args after SelectedStreamConfig is retrieved from iOS devices, merging three 20ms frames into one rtp packet is implemented to deal with mismatched frame sizes. For the timestamp increment, it will be frame size(provided by client) * sampling rate(16k with `opus/16000`). It looks like default `opus` also works.

I put these in `pkg/srtp` and I'm not a golang expert. Please let me know if there's better approaches.


AlexxIT:
https://github.com/AlexxIT/go2rtc/releases/tag/v1.8.5

Thank you! Your support has been very helpful. Rewrote your code a lot, so I can't merge this. Please do some more PR to get on the list of contributors :)

Issue #723, opened 2023-10-31
Nest camera stops working after 5 minutes


Inrego:
After 5 minutes, the nest stream stops working. I've set it up manually with the parameters (not through HA).
So something about the reconnection seems to not be working correctly.

Nothing seems to be logged around the time when the stream stops working, so I'm not sure what else I can provide.

heroash88:
I think this is a limit Google enforced in their API unless you stream it on the Google Home App.

Inrego:
Another fun thing I experienced recently, was that when the stream stops working, then Frigate suddenly used all remaining 20 GB RAM on my machine within seconds, rendering the machine unresponsive due to lack of RAM.

AlexxIT:
go2rtc not using extend API. There are no such plans in the near future.
The fact that the stream is not reconnecting is a bug. It shouldn't be.

Inrego:
I'm sorry, but I'm having a hard time understanding what you mean by your second line

> The fact that the stream is not reconnecting is a bug. It shouldn't be.

Should it be reconnecting automatically, or not?

AlexxIT:
It should be reconnecting. But I haven't this source for testing.

Inrego:
> Hello how exactly do I get the 'device_id=' and the 'refresh_token' ?

It's kind of off topic, but you can use this site:
https://device-access-sample.web.app/

Inrego:
> It should be reconnecting. But I haven't this source for testing.

I might have a go at making a PR for it at some point when I have the time.
Would you accept if I implement calling the stream extension api, rather than reconnecting (while still having reconnect as a fallback if connection is lost)?

AlexxIT:
If the code will be correct and logical - I will accept it :)

Inrego:
I'll do my best. I've no experience in Go, but I've many years of professional experience in C#

Inrego:
I'm afraid I won't have the time to get the proper understanding of the workings, to fix the issue. But I'll share my sparse findings here:
When the stream expires, the webrtc state changes to disconnected, and then closed.

When I manually close the stream by closing VLC, then the disconnected even is absent, and it only closes.

I believe the first step should be to fix the issue of reconnection. Because even if the extend stream is implemented, there could be edge cases where it's not fired in time, or for other reasons fail.

Inrego:
After the stream expires, frigate shows the stream as full of flickering artifacts, and it uses up all my RAM (~30 GB) in seconds.

[...]

PR #855, opened 2024-01-03
Nest extend stream

Inrego:
This solves the issue of the stream expiring after 5 minutes.
However, this does not yet make the nest integration stable. If the stream for some reason should expire, any streaming clients will lose connection. And in the case of Frigate, it will consume all available ram within seconds of reconnecting. I've added this disclaimer to the readme.

Locally, I've fixed this as described here:
https://github.com/AlexxIT/go2rtc/issues/723#issuecomment-1872591484

But I don't feel confident committing this change, as I'm unsure if it will break other WebRTC sources.

### DIFF
--- a/README.md
+++ b/README.md
@@ -579,8 +579,7 @@ streams:
 
 Any cameras in WebRTC format are supported. But at the moment Home Assistant only supports some [Nest](https://www.home-assistant.io/integrations/nest/) cameras in this fomat.
 
-The Nest API only allows you to get a link to a stream for 5 minutes, but a call to extend the stream is made before it expires. If for some reason the stream expires anyway, any streaming clients will lose connection.
-Note: Do not use this with frigate. If the stream expires, Frigate will consume all available ram on your machine within seconds.
+**Important.** The Nest API only allows you to get a link to a stream for 5 minutes. Do not use this with Frigate! If the stream expires, Frigate will consume all available ram on your machine within seconds. It's recommended to use [Nest source](#source-nest) - it supports extending the stream.
 
 ```yaml
 streams:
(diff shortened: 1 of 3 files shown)


Inrego:
Just want to add, that with this PR and the local changes described - my Nest camera has been running rock solid since then on Frigate.

T3chArmy:
This issue has been plaguing me for a while, thanks for your work on this! Hopefully I can stop restarting my frigate server so often and get a bit more life out of my nest cameras.🤞

NodeJSmith:
@AlexxIT just wanted to check on the status of this, I'm having a similar problem with my nest cameras and this seems like it may be the solution

T3chArmy:
I generated a build of 1.8.5 with this cherry picked, and have been running for several weeks. Frigate has been very stable since. 👍

Inrego:
Yeah, same. It's working perfectly.

But the quality is apparently not enough to be merged. The maintainers should remove any mention of nest support until this is fixed imo, since nest does not work in the current state

AlexxIT:
@Inrego Thanks! I made some refactoring to your PR. Unfortunately I don't use Nest to do the testing myself. I hope nothing broke.

I find the Nest APIs to be just plain awful. The whole thing about getting access for $5 and a bunch of tokens is terrible.

AlexxIT (approved):

T3chArmy:
I have 3 left, and I agree. They're a PITA. They're not perfect, and there are frequent artifacts(mostly due to wifi strength I'm pretty sure). I'm slowly moving away from them, but having this has been a nice stopgap to get a little extra life out of them.

NodeJSmith:
Just switched to the master branch in HA to see if this PR improved things and things are working so much better now, thanks @Inrego and @AlexxIT !

AlexxIT:

PR #867, opened 2024-01-09
Clean HASS Urls that come from UniFi Protect

### LINKED ISSUE #81: UniFi RTSPS cameras added via yaml result in packet error
Added a number of UniFi cameras via yaml as below:
```
streams:
 front: rtspx://192.168.1.1:7441/E1oZmxXGX4dAzBsz?enableSrtp
```

When I open go2rtc to view the stream, I either get:
`07:42:38.656 WRN [streams] start error="rtcp: invalid packet version" url=rtspx://192.168.1.1:7441/E1oZmxXGX4dAzBsz?enableSrtp`
or
`07:43:30.654 WRN [streams] start error="rtcp: packet too short" url=rtspx://192.168.1.1:7441/E1oZmxXGX4dAzBsz?enableSrtp`

Doesn't matter if I use rtsps or rtspx

ViViDboarder:
This automatically applies the recommended configuration for UniFi cameras. This could possibly be moved to the Home Assistant component when passing the URL, but I'm starting here to get your thoughts.

Closes #81

### DIFF
--- a/internal/hass/api.go
+++ b/internal/hass/api.go
@@ -5,12 +5,17 @@ import (
 	"encoding/json"
 	"net"
 	"net/http"
+	"regexp"
 	"strings"
 
 	"github.com/AlexxIT/go2rtc/internal/api"
 	"github.com/AlexxIT/go2rtc/internal/streams"
 	"github.com/AlexxIT/go2rtc/internal/webrtc"
 )
+
+// UniFi streams will start with rtsps and end with ?enableSrtp
+// this is not compatible with go2rtc and requires some manipulation
+var UnifiMatch = regexp.MustCompile(`rtsps://.+[?]enableSrtp`)
 
 func apiOK(w http.ResponseWriter, r *http.Request) {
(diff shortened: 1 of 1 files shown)


AlexxIT:
I don't think this is good solution.
1. RTSPX should't be necessary, because latest go2rtc won't check TLS for IP-address as host.
2. Removing enableSrtp necessary. But I don't think this should be automatically. It's better to add SRTP support, but I don't have access to any UniFi camera

ViViDboarder:
It didn’t seem that you were interested in adding SRTP support because it was only useful for UDP and go2rtc only used TCP. Is the scenario changed? I am not very familiar with the protocol, but I can poke around at it.

AlexxIT:
You right. I think it's unnecessary to use SRTP over HTTPS. But looks like this is default option for UniFi.

AlexxIT:
Probably doesn't make sense anymore
https://github.com/home-assistant/core/pull/129852

ViViDboarder:
Yes. Thanks! I was about to get to work to write a patch like that and remove the parameter there but saw this first.

PR #871, opened 2024-01-10
Feature: Make kill signal configurable

### LINKED ISSUE #870: Feature-Request: Make kill method of process configurable
I'm trying to stream my canon eos 600d feed with this config:

```yaml
streams:
  canon: exec:gphoto2 --capture-movie --stdout
```

This works exactly one time. But the process is killed immediately (calling `os.Process.Kill`) by go2rtc after closing the stream without giving gphoto2 enough time for a proper cleanup.

gphoto2 is waiting for the SIGINT signal, therefore my request is to make that configurable.

dadav:
This makes it possible to end `exec` commands gracefully.

Also fixed a typo on the way :)

Fixes #870


dadav:
This also makes it possible to use `gphoto2` as a source.

Like this:

```yaml
exec:
  signal: sigint

streams:
  canon: exec:gphoto2 --capture-movie --stdout
```

AlexxIT:
I don't think this is good idea to use same kill signal for all exec sources. It should be `#param` like ffmpeg and other sources.

dadav:
> I don't think this is good idea to use same kill signal for all exec sources. It should be `#param` like ffmpeg and other sources.

Good point, I'm gonna fix it

skrashevich:
The process may not terminating upon receipt of sigint. It's important to implement some timeout, after which sigkill will be sent. Otherwise, zombie processes may appear

dadav:
Better? @AlexxIT

skrashevich:
What happens if anyone set #killtimeout to zero or negative number? ;)

dadav:
> What happens if anyone set #killtimeout to zero or negative number? ;)
> 
> UPD: oops, looks like, that nothing bad

killtimeout=-1 can be used to disable the second signal with SIGKILL. So, it's a feature? :)

dadav:
Found another problem. When the main process is killed, the child process is not killed correctly. I did some tests in #879 which showed that child processes receive the same signals as parents, so I thought this would be no problem, but it is somehow.

dadav:
Is there anything I can work on to make this PR more acceptable?

AlexxIT:
Sorry. Don't have time to review. Fixing my other projects after Home Assistant 2024.1 update...

dadav:
Just realized this could make problems on windows/darwin. Need to invest some time in os specific code..

dadav:
This should work..

AlexxIT (approved):

AlexxIT:
Thanks! I've rewritten the code to make it simpler.

felipecrs:
I would probably change the default behavior, instead of killing by default, sending SIGTERM and waiting some arbitrary time like 5 seconds before killing the process.

AlexxIT:
For example, ffmpeg will fail in any way, because stdout will be closed before kill signal.

AlexxIT:

PR #878, opened 2024-01-13
fix(clipboard): fix copy to clipboard functionality

skrashevich:
Added a `copyTextToClipboard` function to handle text copying across different browsers and fallback scenarios. This function utilizes the Clipboard API when available, providing an asynchronous method to copy text securely. For browsers where the Clipboard API is not available or the page is not served over a secure context, a fallback method using a temporary textarea element and `document.execCommand` is employed. Replaced direct use of `navigator.clipboard.writeText` with this function in the 'shareget' click event listener to enhance cross-browser support and error handling.

### DIFF
--- a/www/links.html
+++ b/www/links.html
@@ -170,6 +170,57 @@ Telegram: rtmps://xxx-x.rtmp.t.me/s/xxxxxxxxxx:xxxxxxxxxxxxxxxxxxxxxx</pre>
         document.getElementById('sharedel').style.display = 'none';
     }
 
+    function copyTextToClipboard(text) {
+        // Modern approach with the Clipboard API
+        if (navigator.clipboard && window.isSecureContext) {
+            // Navigator clipboard is available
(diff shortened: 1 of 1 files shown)


skrashevich:
> added the [doubt](https://github.com/AlexxIT/go2rtc/labels/doubt)  label [7 hours ago](https://github.com/AlexxIT/go2rtc/pull/878#event-12563219471)

reason?

AlexxIT:
What browsers has problems with one line clipboard function?

skrashevich:
> What browsers has problems with one line clipboard function?

[Safari on non-https pages](https://webkit.org/blog/10247/new-webkit-features-in-safari-13-1/) and [Firefox](https://developer.mozilla.org/en-US/docs/Web/API/Clipboard/write#browser_compatibility)

AlexxIT (approved):

AlexxIT:
Thanks!

PR #930, opened 2024-02-06
Feat: Expressions anywhere in the configuration

skrashevich:
This PR introduces the ability to preprocess configuration data to evaluate expressions contained within it.

The ability to preprocess and evaluate expressions within the configuration could simplify dynamic configuration scenarios and reduce the need for external templating tools.

example:
```yaml
streams:
  test:
    - http://192.168.88.158/live/index.m3u${{ 4+4 }}
```
will be pre-processed to `http://192.168.88.158/live/index.m3u8`

Roadmap:
- [ ] test for regressions and backward compatibility
- [ ] assess the impact on security
- [ ] add support for this syntax to the configuration editor

### DIFF
--- a/internal/app/app.go
+++ b/internal/app/app.go
@@ -9,6 +9,7 @@ import (
 	"runtime"
 	"strings"
 
+	"github.com/AlexxIT/go2rtc/pkg/expr"
 	"github.com/AlexxIT/go2rtc/pkg/shell"
 	"github.com/AlexxIT/go2rtc/pkg/yaml"
 	"github.com/rs/zerolog/log"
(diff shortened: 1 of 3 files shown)


felipecrs:
Just wondering, why not add support for go-template the config file directly?

skrashevich:
> Just wondering, why not add support for go-template the config file directly?

What do you mean? Go-template is a template engine, but my idea is in a dynamic calculated configuration file

felipecrs:
If you make the config file a template, and use go-template to render it before consuming from it, it should give a similar result but maybe more powerful.

```yaml
streams:
  test:
    - rtps://{{ env "MY_USER" }}:{{ env "MY_PASS" }}@myhost.com
```

PR #949, opened 2024-02-17
Fix "no response" error when viewing cameras via apple watch

civita:
Apple watch requires specific video resolution and frame rate, and without it the error message "no response" will be shown when the camera is clicked and watched via apple watch. Already paired cameras won't take effect until they are re-paired.

### DIFF
--- a/pkg/hap/camera/accessory.go
+++ b/pkg/hap/camera/accessory.go
@@ -62,6 +62,7 @@ func ServiceCameraRTPStreamManagement() *hap.Service {
 				VideoAttrs: []VideoAttrs{
 					{Width: 1920, Height: 1080, Framerate: 30},
 					{Width: 1280, Height: 720, Framerate: 30}, // important for iPhones
+					{Width: 320, Height: 240, Framerate: 15}, // apple watch
 				},
 			},
 		},


AlexxIT:
1. Are you sure all models require that attrs?
2. The watch will show the stream even if go2rtc gives a completely different resolution?

civita:
Hi Alex,

Thanks for the suggestion. I was able to verify that it works on my Apple Watch 7. The specific resolution/fps are also mentioned in [HAP-NodeJS](https://github.com/homebridge/HAP-NodeJS/blob/6dbba2f9956d059ddc7976d44e060c030ac9ad06/src/accessories/Camera_accessory.ts#L640C1-L641C1) and [Scrypted](https://github.com/koush/scrypted/blob/d6b9900db5b443cc0325122665efe05e95494d99/plugins/homekit/src/types/camera.ts#L82). The watch will only show thumbnail and won't show real-time stream if go2rtc gives different resolution.

incoming-71281918-8D06-4979-A199-A7E6214B39F5
Thumbnail preview

incoming-7CA8C933-62C1-4112-A798-83E040490F76
Streaming

AlexxIT:
This is because thumbnail can be requested with any resolution

erikpendragon (approved):

AlexxIT (approved):

AlexxIT:

PR #961, opened 2024-02-23
Fix crash with tapo cameras not returning 401

janza:
err is nil and causes nil pointer dereference panic on line [86](https://github.com/janza/go2rtc/blob/14a9763c73dcbe0e46d602d8437aa46baaf94e88/pkg/tapo/client.go#L86)


AlexxIT:
Thanks!

AlexxIT:

PR #963, opened 2024-02-24
feat(app): support daemon mode on non-Windows platforms

skrashevich:
Added a new command-line flag `-daemon` to run the application in the background as a daemon. This option is only available for non-Windows operating systems due to platform-specific process handling. When enabled, the application restarts itself with the same arguments except for the `-daemon` flag, prints the PID of the background process, and then exits the current process.

Simplified replacement for #440

### DIFF
--- a/internal/app/app.go
+++ b/internal/app/app.go
@@ -5,6 +5,7 @@ import (
 	"flag"
 	"fmt"
 	"os"
+	"os/exec"
 	"path/filepath"
 	"runtime"
 	"strings"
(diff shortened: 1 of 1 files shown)


skrashevich:
Why doubt?

AlexxIT:
Why not use simple & symbol?

skrashevich:
> Why not use simple & symbol?

These are completely different things. `&` is a shell function, not a kernel.

For example, a background process started by `&` will not stay alive after the shell session is closed

AlexxIT (approved):

AlexxIT:
Thanks!

AlexxIT:

PR #967, opened 2024-02-25
Modify ISAPI to reliably open connections

f1d094:
Hikvision ISAPI may not accept a new open request if the previous one was not closed (e.g. using the test button on-camera or via curl command) but a close request can be sent even if the audio is already closed. 

Using go2rtc to load microphone, regardless if using dashboard, WebRTC card, or Frigate card, resulted in ISAPI connections that were constantly failing with 401 errors. Testing revealed that manually sending a close request just before opening microphone url allowed it to proceed. 

Modifying func Open in pkg/isapi/client.go to send a 'close' before 'open' seems to be working reliably with Frigate HASS card.

Modified func Close in pkg/isapi/client.go to call `/ISAPI/System/TwoWayAudio/channels/<channel id>/close` instead of `/ISAPI/System/TwoWayAudio/channels/<channel id/close/open`

Made minor mod to pkg/tcp/request.go, changing cnonce to 32 bytes. Still temporary/static and needs randomized.

### DIFF
--- a/pkg/isapi/client.go
+++ b/pkg/isapi/client.go
@@ -86,12 +86,28 @@ func (c *Client) Dial() (err error) {
 func (c *Client) Open() (err error) {
 	link := c.url + "/ISAPI/System/TwoWayAudio/channels/" + c.channel
 
-	req, err := http.NewRequest("PUT", link+"/open", nil)
+   // Hikvision ISAPI may not accept a new open request if the previous one was not closed (e.g.
+   // using the test button on-camera or via curl command) but a close request can be sent even if
+   // the audio is already closed. So, we send a close request first and then open it again. Seems
+   // janky but it works.
+	req, err := http.NewRequest("PUT", link+"/close", nil)
 	if err != nil {
 		return err
 	}
 
 	res, err := tcp.Do(req)
+	if err != nil {
+		return err
+	}
+
+	tcp.Close(res)
+
+	req, err = http.NewRequest("PUT", link+"/open", nil)
+	if err != nil {
+		return err
+	}
+
+	res, err = tcp.Do(req)
 	if err != nil {
 		return
 	}
@@ -124,8 +140,8 @@ func (c *Client) Open() (err error) {
 }
 
 func (c *Client) Close() (err error) {
-	link := c.url + "/ISAPI/System/TwoWayAudio/channels/" + c.channel + "/close"
-	req, err := http.NewRequest("PUT", link+"/open", nil)
+	link := c.url + "/ISAPI/System/TwoWayAudio/channels/" + c.channel
+	req, err := http.NewRequest("PUT", link+"/close", nil)
 	if err != nil {
 		return err
 	}
--- a/pkg/isapi/client.go
+++ b/pkg/isapi/client.go
@@ -84,30 +84,23 @@ func (c *Client) Dial() (err error) {
 }
 
 func (c *Client) Open() (err error) {
-	link := c.url + "/ISAPI/System/TwoWayAudio/channels/" + c.channel
-
    // Hikvision ISAPI may not accept a new open request if the previous one was not closed (e.g.
    // using the test button on-camera or via curl command) but a close request can be sent even if
    // the audio is already closed. So, we send a close request first and then open it again. Seems
    // janky but it works.
-	req, err := http.NewRequest("PUT", link+"/close", nil)
+
+   err = c.Close()
+   if err != nil {
+      return err
+   }
+
+	link := c.url + "/ISAPI/System/TwoWayAudio/channels/" + c.channel
+	req, err := http.NewRequest("PUT", link+"/open", nil)
 	if err != nil {
 		return err
 	}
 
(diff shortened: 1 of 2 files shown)


pergolafabio:
Hey @f1d094 , thats interesting, i was also using the Frigate card to start two way audio on my Hikvision intercom.. When the intercom is ringing, i first send the curl command to stop the ringing, so the speaker is idle again

Then with Frigate, i start two way audio by clicking on the mic button, that works, if i then mute/unmute again, it doesnt work anymore... Gonna try sending the curl with the "close" command to see if it indeed resolves the issue, keep you posted


But, whats strange, if i stop the two way audio, and do a full refsresh of the page/view where the frigate card is located, then i can make it work , then i can again start two way audio  , so that means the channel is somehow closed anyway?

Also, when using the Webrtc Card by AlexxIT instead of the frigate card, i can mute/unmute as many times i want... it stays working... maybe there is something different in those cards? maybe muting/unmuting on webrtc card is just muting thge sound? while clicking on the microphone button on the frigate card is really stopping the audio? ( or tries to)

dermotduffy:
> Why the video should need to reload when I enable the two-way audio, I have no idea. 

The Frigate card when configured with `go2rtc` as the live provider, needs to reset the video stream momentarily to add in the separate microphone stream (unless you use the `always_connected` option, in which case it's included at the start). When I was implementing this, I could not find a way to dynamically do this without resetting the connection. I do plan to revisit this implementation and see if there's a better way, but what you are describing is WAI as coded right now.

pergolafabio:
hey @f1d094  , your solution fixed my issue, sending the close command before starting it again, i have now alwaysb 2way audio!!!

pergolafabio:
> > unless you use the `always_connected` option, in which case it's included at the start
> 
> Aha! This is news I can use. I will try that out this evening and see if that smooths out my setup. Thanks!

i tried the always_connected option, but the audio was not closed for me

pergolafabio:
nice!

are you still proceeding the ISAPI close command, that one i need for the hikvision intercom
Or is possible you can create an ADDON repo so i can try the new fix out?

pergolafabio:
alex, is using this repo for the addons, such an url i need to add:

https://github.com/AlexxIT/hassio-addons

pergolafabio:
or i need to fork this one, and point the url to your github, and remove the image

https://github.com/AlexxIT/hassio-addons/blob/master/go2rtc/config.yaml

pergolafabio:
ok, nevermind, i wait for the fix to be implemented :-)

i now do the isapi curl command before i start the audio, working perfectly :-)

felipecrs:
> @pergolafabio you can just place the binary at /config/go2rtc. The add-on (and Frigate) automatically picks it up instead if it's there.

Ops, sorry. That's not true. It's only true for the Frigate add-on.

pergolafabio:
Yeah , indeed, I'm using only the card

AlexxIT (approved):

AlexxIT:
Thanks!

AlexxIT:

pergolafabio:
Hi @f1d094 , i still have the issue with frigate card? are you using hikvision cameras or doorbell?
I have a doorbell, on an real incoming call to my hikvision indoor panel, i send an answer+hangup command to the hikvision device; so the user outdoor "thinks" i accepted the doorbell ring...
Afterwards, i open HA and browse to the frigate card and i can start two way audio ... this approach works perfectly
On the second call, seems when i turn on the two way audio again, it doesnt work anymore ...

I tought your approach always closed the audio before starting a new session?

What i do now, before starting the MIC on the frigate card, i send first the two way audio close command ... but is there a better/cleaner way?

This only happens on door intercom systems, because if send a fake "answer+hangup" command, i think somehow the the door intercom still thinks the session is open

thnx for the feedback

pergolafabio:
hi, yes, if you can share your card config, that can maybe be helpfull indeed
i have set it like below, i only want to speak/answer when needed 


```
          live:
            microphone:
              always_connected: false
              disconnect_seconds: 0
```

pergolafabio:
like this?

```
          live:
            microphone:
              always_connected: true
              mute_after_microphone_mute_seconds: 90
              # disconnect_seconds: 0
```

will try later to see if there is a diffeerence

pergolafabio:
hmm, tried it , but no difference

in you PR, at what point are you actually sending the close audio command?

pergolafabio:
Well it works, but you need to refresh the card or wait 60 or 90 seconds, somehow the doorintercom keeps the audio session open..

When I test for example an incoming call, I abort the call with hikconnect or with my indoor unit, when I then start two way audio, it doesn't work, only 1 way audio works, the audio session is still open in background somehow...
If first need to close it , before starting a new session ... This is confirmed already by lots of users with same issue :-(

felipecrs:
> If first need to close it , before starting a new session ... This is confirmed already by lots of users with same issue :-(

But this is exactly what this PR aims to do and solve. If it's not doing that, maybe there's a bug in it.

Issue #982, opened 2024-03-04
[Feature Request] Expose Homekit Camera as DoorBell instead of IP Camera.


Minims:
Hello,

Is there a way to expose in Homekit my DoorBell as a DoorBell instead as a standard IP Camera ? 

The device Type is not the same if i pair directly the Device of if i Proxy it with go2rtc.

Thanks.

PR #985, opened 2024-03-05
feat: allow to set deviceCategory exposed in homekit while proxying

Minims:
Hello,

I have open this FR https://github.com/AlexxIT/go2rtc/issues/982 and trying to add it now.

I don't really know what change are need, but as i understand, this PR can be the beginning of the feature.

Waiting for your feedback @AlexxIT 

Thanks.

### DIFF
--- a/internal/homekit/homekit.go
+++ b/internal/homekit/homekit.go
@@ -63,6 +63,7 @@ func Init() {
 		}
 
 		deviceID := calcDeviceID(conf.DeviceID, id) // random MAC-address
+		deviceCategory := calcDeviceCategory(conf.DeviceCategory)
 		name := calcName(conf.Name, deviceID)
 
 		srv := &server{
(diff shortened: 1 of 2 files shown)


AlexxIT:
Are you sure this is enough? HomeKit has a rules for devices. Doorbell should have a button, maybe something else.

Minims:
I’m not sure and i don’t know how to test my code change.

I would expect that proxying expose all original devices and not only camera. 

Maybe it’s possible to do that in go2rtc for every homekit devices ?

AlexxIT:
Oh. This make sense for proxy mode.

Minims:
It seems the current `CategoryCamera` does not proxy only the camera.

I have a "Somfy Indoor Camera" and the 4th entity are available (and working) in both Homekit and HA.
 - Stream
 - Identify
 - Mute Camera
 - Motion Sensor

So I would say it will also work for DoorBell or Bridge Category.

AlexxIT:
Your code will not compile anyway.

AlexxIT:
Thanks. I took your idea and partially reworked it in a separate commit.

AlexxIT:
https://github.com/AlexxIT/go2rtc/releases/tag/v1.9.13

PR #1016, opened 2024-03-22
feat(docker): update Dockerfiles to use ENTRYPOINT instead of CMD

skrashevich:
- Added a new `entrypoint.sh` script for dynamic configuration and validation of environment variables, improving container startup flexibility.
- Included `yq` utility in the base images to facilitate YAML file manipulation within the container, enhancing configuration management capabilities.
- Specified `--platform=linux/amd64` for base images to ensure consistent cross-platform builds, addressing potential compatibility issues on multi-architecture setups.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-# syntax=docker/dockerfile:labs
+# syntax=docker/dockerfile:1.4
 
 # 0. Prepare images
 ARG PYTHON_VERSION="3.11"
@@ -32,7 +32,7 @@ RUN --mount=type=cache,target=/root/.cache/go-build CGO_ENABLED=0 go build -ldfl
 FROM scratch AS rootfs
 
 COPY --from=build /build/go2rtc /usr/local/bin/
-COPY --from=ngrok /bin/ngrok /usr/local/bin/
+COPY --link --from=ngrok /bin/ngrok /usr/local/bin/
 
 
 # 3. Final image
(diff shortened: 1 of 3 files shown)


AlexxIT:
Why we need to create config file? User can set default port just via command line
https://github.com/AlexxIT/go2rtc/wiki/Configuration

skrashevich:
because of without config file:
image


Also, listening to the port given by the `PORT` environment variable is good practice, this is the expected behavior of the applications in serverless environments, for example

AlexxIT:
Config file is not set - error only when user set config as string. It's planned behaviour.

skrashevich:
> Config file is not set - error only when user set config as string. It's planned behaviour.

anyway, other reasons still significant:
- support `PORT` env var (ok, I can refactor entrypoint.sh to use it in cli arg instead of config)
- it's a more correct way, allowing the user in the `docker run` command to add their own cli arguments

AlexxIT:
1. Env vars supported inside config file
2. Config supported inside command line

Both this instruments can be used to set port

skrashevich:
> 1. Env vars supported inside config file
> 2. Config supported inside command line
> 
> Both this instruments can be used to set port

This all requires a separate configuration from user, not a default behavior for go2rtc container

AlexxIT:
No. 2nd can be used in raw docker compose file. Without any file config

PR #1029, opened 2024-04-01
feat(tapo): Add support for H200 hub and child devices (for example battery/sub2G powered D230S1)

aprilmaccydee:
- Allows streaming Tapo devices behind a hub, for example, the D230S1 battery powered doorbell with a H200 parent hub.
- Small fix for authenticating correctly to the H200 hub.

To use, you would use a URL like this:
```
streams:
    camera: tapo://admin:passwordAsUpperCaseSHA256@h200HubIpAddress/?deviceId=childDeviceIdFromPyTapo
```
Basically the same as a non-hub'd tapo camera but with added deviceId query string.

There may be a better way to get the device ID, but as an example to figure out the child device ID, use pytapo and call
```
print(tapo.getMost())
```

### DIFF
--- a/pkg/tapo/client.go
+++ b/pkg/tapo/client.go
@@ -77,11 +77,8 @@ func (c *Client) newConn() (net.Conn, error) {
 
 	query := u.Query()
 
-	deviceId := query.Get("deviceId")
-	if deviceId != "" {
-		q := req.URL.Query()
-		q.Set("deviceId", deviceId)
-		req.URL.RawQuery = q.Encode()
+	if deviceId := query.Get("deviceId"); deviceId != "" {
+		req.URL.RawQuery = "deviceId=" + deviceId
 	}
 
 	req.URL.User = u.User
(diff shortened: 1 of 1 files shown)


pkg/tapo/client.go
```diff
@@ -297,17 +308,33 @@ func dial(req *http.Request) (net.Conn, *http.Response, error) {
...
 	response := tcp.HexMD5(ha1, nonce, nc, cnonce, qop, ha2)
 
-	header := fmt.Sprintf(
-		`Digest username="%s", realm="%s", nonce="%s", uri="%s", qop=%s, nc=%s, cnonce="%s", response="%s"`,
-		username, realm, nonce, uri, qop, nc, cnonce, response,
-	)
+	header := ""
+	if op != "" {
+		header = fmt.Sprintf(
+			`Digest username="%s", realm="%s", nonce="%s", uri="%s", qop=%s, nc=%s, cnonce="%s", response="%s", opaque="%s", algorithm=MD5`,
```
AlexxIT, line 329: Are you sure about `algorithm=MD5` param?

AlexxIT (approved):

AlexxIT:
Thanks!

AlexxIT:

PR #1035, opened 2024-04-09
fix(docker): optimize docker hardware image size by cleaning up apt cache

skrashevich:
This PR optimizes the -hardware Docker image size by including commands to clean up the APT cache after package installation. This change reduces the overall image size by removing unnecessary files and directories that are not needed in the final image, leading to faster download and deployment times.


AlexxIT (approved):

AlexxIT:
Thanks!

PR #1047, opened 2024-04-17
feat(api): add system stats (CPU and RAM usage) to API response

skrashevich:
- Implement GetCPUUsage and GetRAMUsage functions in core package for fetching CPU and memory usage.
- Update apiHandler to include CPU and memory usage in the API response under "stats".
- Ensure thread safety in apiHandler by properly using defer for mutex unlock.
- Extend frontend to display fetched CPU and RAM usage statistics.
- Add tests for new core functionality ensuring correct operation of system resource usage fetching.


AlexxIT:
Can we do same without 8 new dependencies?

AlexxIT:
What if we make support only for the main OS - Linux? Will the dependencies get a lot less complicated?

skrashevich:
> What if we make support only for the main OS - Linux? Will the dependencies get a lot less complicated?

API should return a unified and predictable response, regardless of the OS running

AlexxIT:
Then there will be no such API at all inside go2rtc

skrashevich:
> Then there will be no such API at all inside go2rtc

But the implementation in this PR provides unified and predictable response across OSes

AlexxIT:
I don't want confusing dependencies inside the `go.mod` file. CPU stats are not worth it.

AlexxIT:
Check this. I haven't tested. But it looks like less dependencies
https://github.com/elastic/go-sysinfo

AlexxIT:
I still tend to support Linux only if it's a necessary requirement for light dependencies. CPU and RAM stats is useful feature.

PR #1051, opened 2024-04-18
Integrate WebRTC with RESTful API for Milestone XProtect VMS

ggenny:
This pull request adds a WebRTC module to the Milestone XProtect VMS by integrating it with RESTful API services for authentication, session creation, and session updates. The integration aims to streamline the process of establishing a WebRTC connection with an XProtect VMS via the API Gateway. This implementation adheres to the architectural principles outlined in the [Milestone SDK WebRTC JavaScript Sample](https://github.com/milestonesys/mipsdk-samples-protocol/tree/main/WebRTC_JavaScript).

### DIFF
--- a/internal/mp4/mp4.go
+++ b/internal/mp4/mp4.go
@@ -91,8 +91,7 @@ func handlerMP4(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
-	src := query.Get("src")
-	stream := streams.Get(src)
+	stream := streams.GetOrPatch(query)
 	if stream == nil {
 		http.Error(w, api.StreamNotFound, http.StatusNotFound)
 		return
--- a/internal/webrtc/client.go
+++ b/internal/webrtc/client.go
@@ -49,7 +49,7 @@ func streamsHandler(rawURL string) (core.Producer, error) {
 
 		case "http", "https":
 			if format == "milestone" {
-				return milestoneClient(rawURL, query, "Milestone")
+				return milestoneClient(rawURL, query)
 			} else if format == "wyze" {
 				// https://github.com/mrlt8/docker-wyze-bridge
 				return wyzeClient(rawURL)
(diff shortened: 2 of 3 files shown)


ggenny:
I've noticed that the video is visible but the timestamp I receive is not correct. I'm not sure if this stems from a bug in the Milestone server. It appears to be 100 times smaller than it should be, and needs to be multiplied by a factor. The API includes a video speed factor, but it seems to only pertain to playback and not to live streaming. Is this correct, or am I missing a time factor/time scale somewhere that I haven't noticed in the stream negotiation?


```
diff --git a/internal/webrtc/milestone.go b/internal/webrtc/milestone.go
index f63f49e..947183d 100644
--- a/internal/webrtc/milestone.go
+++ b/internal/webrtc/milestone.go
@@ -229,6 +229,7 @@ func milestoneClient(rawURL string, query url.Values, desc string) (core.Produce
        prod := webrtc.NewConn(mc.PeerConnection)
        prod.Desc = "WebRTC/OpenIPC"
        prod.Mode = core.ModeActiveProducer
+       prod.TimeSpeed = 100
 
        resp, err := createWebRTCSession(mc, details)
        if err != nil {
diff --git a/pkg/webrtc/conn.go b/pkg/webrtc/conn.go
index 6483535..053da5b 100644
--- a/pkg/webrtc/conn.go
+++ b/pkg/webrtc/conn.go
@@ -15,6 +15,7 @@ type Conn struct {
        UserAgent string
        Desc      string
        Mode      core.Mode
+       TimeSpeed float32
 
        pc *webrtc.PeerConnection
 
@@ -99,6 +100,10 @@ func NewConn(pc *webrtc.PeerConnection) *Conn {
                                return
                        }
 
+                       if c.TimeSpeed != 0 {
+                               packet.Timestamp = uint32(float32(packet.Timestamp) * c.TimeSpeed)
+                       }
+
                        if len(packet.Payload) == 0 {
                                continue
                        }

```

AlexxIT (approved):

AlexxIT:
I have made big code refactoring for your PR. I hope nothing breaks, because I can't test this code.

AlexxIT:
I don't know about timespeed. Maybe some problems on milestone side. Because there is no problems with all other clients.

AlexxIT:

PR #1059, opened 2024-04-23
tls-expire(api): add TLS cert expiration check

skrashevich:
- **Refactor of certificate loading into its own function:**
   - Extracted the previous inline certificate loading code from `tlsListen` into a new, dedicated function `LoadCertificate`. This function is responsible for loading a TLS certificate from given file paths or directly from string content, ensuring better testability and separation of concerns.
   - Now, `LoadCertificate` returns both `tls.Certificate` and error, which improves error handling allowing the calling function to take decisions based on the error returned.

- **Enhanced TLS Certificate Loading and Validating in `tlsListen`:**
   - Integrated the new `LoadCertificate` function in the `tlsListen` function.
   - Added error logging right after certificate loading to immediately catch and log issues pertaining to file access or content issues without proceeding further.
   - Performed certificate expiration check after loading the certificate; logs an error if the certificate has expired, otherwise logs the expiry date. This preemptive check helps avoid runtime surprises related to expired certificates.

- **Improved logging:**
   - Added additional log entries such as trace logs for beginning of the listening process and informative logs for successful listening setup including certificate expiry details. This enhancement aids in better observability and debugging capabilities of the listening process.

### DIFF
--- a/internal/api/api.go
+++ b/internal/api/api.go
@@ -2,6 +2,7 @@ package api
 
 import (
 	"crypto/tls"
+	"crypto/x509"
 	"encoding/json"
 	"fmt"
 	"net"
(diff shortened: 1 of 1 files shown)


AlexxIT:
I still think that running TLS on the go2rtc side is not a core functionality at all. And in a normal setup it should be given to special software - nginx or similar.

felipecrs:
@AlexxIT I agree with you, it's not worth implementing these features in go2rtc.

However, @skrashevich implemented already, so it's free. I'd probably take it. 😅

AlexxIT:
Every line of code is not free. It can increase support time. Time of reading code. Time of adding new cider, that can be affected. Time of answering on questions about this functionality.

felipecrs:
That's 100% true.

PR #1068, opened 2024-04-29
fix(config): enhance config loading with fallback and error handling

skrashevich:
- Added a check to skip empty config entries to prevent unnecessary processing.
- Implemented error handling for reading config files to set a flag when a valid config is found.
- Introduced a fallback mechanism to load `go2rtc.yaml` if no other configs are successfully loaded, improving robustness.
- Added a `prepend` function to add the fallback config at the beginning of the configs slice, ensuring it's processed first if used.

### DIFF
--- a/internal/app/app.go
+++ b/internal/app/app.go
@@ -27,6 +27,8 @@ func Init() {
 	var confs Config
 	var daemon bool
 	var version bool
+	configflag := false
+	confs = append(confs, "")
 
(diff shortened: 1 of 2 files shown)


AlexxIT:
The current logic for loading configs works correctly. If the user wants a fallback - let him explicitly specify it in the startup line.

PS. `len(s) == 0` better than `len(s) < 1`

PR #1121, opened 2024-05-18
Switched editor from Ace to Monaco

skrashevich:
Replaced Ace Editor with Monaco Editor in editor.html for an enhanced editing experience.

### DIFF
--- a/www/editor.html
+++ b/www/editor.html
@@ -4,7 +4,7 @@
     <title>go2rtc - File Editor</title>
     <meta name="viewport" content="width=device-width, user-scalable=yes, initial-scale=1, maximum-scale=1">
     <meta http-equiv="X-UA-Compatible" content="ie=edge">
-    <script src="https://unpkg.com/ace-builds@1.33.1/src-min/ace.js"></script>
+    <script src="https://cdnjs.cloudflare.com/ajax/libs/monaco-editor/0.48.0/min/vs/loader.min.js"></script>
     <style>
         body {
             font-family: Arial, Helvetica, sans-serif;
@@ -19,6 +19,10 @@
             width: 100%;
             height: 100%;
         }
+
+        #config {
+            flex: 1;
+        }
     </style>
 </head>
 <body>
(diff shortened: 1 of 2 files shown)


AlexxIT:
4 spaces for tabs in YAML should be fixed.

skrashevich:
image
Implemented configuration validation with guidance prompts.

AlexxIT:
I don't like `monaco-yaml-prebuilt` idea.

skrashevich:
> I don't like `monaco-yaml-prebuilt` idea.

Me too. But this is the only option without local build and storage of the editor bundle. (however, it can be stored on the go2rtc GitHub pages)

AlexxIT:
Why do we need bundle? Is it only for tabs? Or also for yaml support?

skrashevich:
> Why do we need bundle? Is it only for tabs? Or also for yaml support?

It's needed for completion, hints, and validation support for YAML.
Without it the [original package can be used](https://github.com/AlexxIT/go2rtc/blob/166cfc5a5b58c659fd406b4f7857fcefbfa07e38/www/editor.html#L7)

Unfortunately, extended YAML support is only available in [external package](https://github.com/remcohaszing/monaco-yaml), which is designed for use in the node-js environment. I couldn’t make it work without building it into a single package 😔

In any case, the build used has [open source](https://github.com/skrashevich/monaco-yaml-prebuilt) which allows it to be easily rebuilt with new versions, and the current versions used will always be available on [npmjs](https://www.npmjs.com/package/@skrashevich/monaco-yaml-prebuilt) and services like unpkg.

AlexxIT:
I have tested this https://monaco-yaml.js.org/ with `# yaml-language-server: $schema=https://` start line.
It works awesome. We can try to use default monaco bundle and host this JS worker and schema.json on github pages.

skrashevich:
> I have tested this https://monaco-yaml.js.org/ with `# yaml-language-server: $schema=https://` start line. It works awesome. We can try to use default monaco bundle and host this JS worker and schema.json on github pages.

Awesome. I couldn't get it to work in such configuration. I would like to see the code that worked

PR #1128, opened 2024-05-20
homekit-server: add pairing API

mnakada:
Current issues:
1. when HomeHub (AppleTV, HomePod) is being used with HomeKit and the device is deleted from the iOS Home app, the HomeHub pairing information remains and the next registration is not possible.

2. If a device is deleted from the iOS Home app while go2rtc is not running, the device does not exist in the Home app, but the go2rtc pairing information remains, so the next registration cannot be performed.

Suggested correction:
Regarding the first problem,
The current code deletes only the pairing information that has the same id in DelPair in internal/homekit/server.go.
In HomeKit, one stream cannot be registered to multiple homes.
Therefore, when deleting pairing information with permissions=1, modify to delete all pairing information of stream.
If permissions=0, delete only the pairing information of the HomeHub, since it is the HomeHub's information.

Since the second problem cannot be handled automatically, add an API for deleting pairing information.
GET to get the current pairing information, and DELETE to delete the pairing that has been deleted by Query.

- GET api/homekit/pairing
-   get the pairing information.
- DELETE api/homekit/pairing?[stream='stream' | name='mDNSname' | device_id='deviceID' ]
-   remove the pairing that has been deleted by Query.

### DIFF
--- a/internal/homekit/api.go
+++ b/internal/homekit/api.go
@@ -137,3 +137,48 @@ func findHomeKitURLs() map[string]*url.URL {
 	}
 	return urls
 }
+
+type PairingInfo struct {
+	Name         string     `json:"name"`
+	DeviceID     string     `json:"device_id"`
+	Pin          string     `json:"pin"`
+	Status       string     `json:"status"`
+}
+
+func getPairingInfo(host string, s *server) PairingInfo {
+	status := "unpaired"
+	if len(s.pairings) > 0 {
+		status = "paired"
+	}
+	return PairingInfo {
+		Name: s.mdns.Name ,
+		DeviceID: s.hap.DeviceID,
+		Pin: s.hap.Pin,
+		Status: status,
+	}
+}
+
+func apiPairingHandler(w http.ResponseWriter, r *http.Request) {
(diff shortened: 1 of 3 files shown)


internal/homekit/api.go
```diff
@@ -137,3 +137,48 @@ func findHomeKitURLs() map[string]*url.URL {
 	}
 	return urls
 }
+
+type PairingInfo struct {
+	Name         string     `json:"name"`
+	DeviceID     string     `json:"device_id"`
+	Pin          string     `json:"pin"`
+	Status       string     `json:"status"`
```
skrashevich, line 145:
```suggestion
	Name         string     `yaml:"name"`
	DeviceID     string     `yaml:"device_id"`
	Pin          string     `yaml:"pin"`
	Status       string     `yaml:"status"`
```

AlexxIT:
1. Deleting all pairing information when `permissions=1` is incorrect. Real HomeKit doesn't work that way.
2. I don't think there's a need for an API to delete pairing information. This information is stored in the config file, and the user can always delete it manually. This is a rare case.

PR #1138, opened 2024-05-25
Homekit qrcode

mnakada:
HomeKit QR-Code
This request is an additional proposal to the following puru request.
homekit-server: add pairing API #1128

Current problem:
HomeKit's QR-Code registration requires that the DeviceID and SetupHash information be announced in mDNS.
The SetupHash is a string Hash value that combines the SetupID and DeviceID.
In the current code, the SetupHash is a Hash value of DeviceID only.
For this reason, QR-Code reading does not recognize it as a HomeKit accessory.

Suggested fixes:
1. Add setup_id to homekit's option in the config file and add it when SetupHash is calculated.
2. Add API to get string for QR-Code using these values.
3. Add a HomeKit section to the web's links page.

### DIFF
--- a/README.md
+++ b/README.md
@@ -970,7 +970,7 @@ HomeKit module can work in two modes:
 streams:
   dahua1: rtsp://admin:password@192.168.1.123/cam/realmonitor?channel=1&subtype=0
 homekit:
-  dahua1:  # same stream ID from streams list, default PIN - 19550224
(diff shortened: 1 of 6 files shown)


AlexxIT:
Thanks. I took your idea and partially reworked it in a separate commit.

AlexxIT:
https://github.com/AlexxIT/go2rtc/releases/tag/v1.9.13

PR #1155, opened 2024-06-01
feat(jpeg): Add keyframe caching with expiration mechanism to JPEG http handler

skrashevich:
Summary of Content:
This PR introduces an enhanced mechanism for caching keyframes within the (M)JPEG streaming service. The major changes include:

1. **Cache Structure Implementation**:
   - A new `CacheEntry` structure that includes a byte slice for the frame and a timestamp for expiration was created.
   - A global `keyframeCache` with a read-write mutex for thread-safe operations was established, storing the cache entries in a map.
   - A placeholder cache duration (`cacheDuration`) has been defined, which is currently set to one minute but marked with a TODO to make it configurable in the future.

2. **Cache Utilization in Keyframe Handler**:
   - Within the `handlerKeyframe` function, the commit implements a check to serve frames directly from the cache if they are found and haven't expired based on `cacheDuration`.
   - If a keyframe is not in the cache or has expired, it is fetched, transformed, and then updated in the cache with the current timestamp.

3. **Periodic Cache Cleanup**:
   - A new `cleanupCache` goroutine function has been added to periodically clean up the expired keyframes from the cache. The cleanup runs in intervals equal to `cacheDuration`, ensuring that stale entries do not persist in memory. 

4. **Helper Function**:
   - Added `writeJPEGResponse` function to consolidate the response writing part with a correct setup for content-type and content-length in the HTTP header, promoting code reuse and clarity.

These modifications improve the efficiency of the JPEG streaming by reducing the need for frequent keyframe reconstructions, thus potentially lowering the processing overhead and speeding up the frame delivery. The implemented caching strategy enables the application to handle high-concurrency scenarios more effectively by reducing duplicate processing and serving cached results wherever applicable.

### DIFF
--- a/internal/mjpeg/init.go
+++ b/internal/mjpeg/init.go
@@ -6,6 +6,7 @@ import (
 	"net/http"
 	"strconv"
 	"strings"
+	"sync"
 	"time"
 
 	"github.com/AlexxIT/go2rtc/internal/api"
@@ -21,6 +22,24 @@ import (
 	"github.com/rs/zerolog/log"
 )
 
+// CacheEntry represents a cached keyframe with its timestamp
+type CacheEntry struct {
+	frame     []byte
+	timestamp time.Time
+}
+
+// Global cache for keyframes with expiration
+var keyframeCache = struct {
+	sync.RWMutex
+	cache map[string]CacheEntry
+}{
+	cache: make(map[string]CacheEntry),
+}
+
+// Cache duration
+// TODO: Make it configurable
+var cacheDuration = 1 * time.Minute
+
 func Init() {
 	api.HandleFunc("api/frame.jpeg", handlerKeyframe)
 	api.HandleFunc("api/stream.mjpeg", handlerStream)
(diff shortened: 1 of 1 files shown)


skrashevich:
@AlexxIT ~~1 minute should be enough for everyone~~ I corrected the comments from https://github.com/AlexxIT/go2rtc/pull/1967#issuecomment-3764967891

AlexxIT:
There's a problem: the user can change the screenshot size through the query. However, the cache is same for both the original and the modified screenshot.

I don't know if it's worth complicating things and making a different cache for every case.

AlexxIT:
I've simplified your code a bit. But it's still confusing that the cache can be different for different snapshot sizes.

AlexxIT:
Thanks! This is not the best solution, but let it be as it is.

internal/mjpeg/init.go
```diff
@@ -23,75 +23,52 @@ import (
...
+			defer func() {
+				entry = cacheEntry{payload: b, timestamp: time.Now()}
+				cacheMu.Lock()
+				if cache == nil {
+					cache = map[string]cacheEntry{src: entry}
+				} else {
+					cache[src] = entry
+				}
+				cacheMu.Unlock()
+			}()
```
skrashevich: It looks like in case of a stream error, we'll cache an empty frame

AlexxIT:

PR #1168, opened 2024-06-06
fix(app): Refactor daemon initialization and add syscall import

### LINKED ISSUE #1165: daemonization causes process to constantly restart with new pid
Running go2rtc on a RPi5.
Works if I just run `go2rtc_linux_arm64`
However if I run `go2rtc_linux_arm64 -d`, then the process continually restarts with a new pid, making the stream useless.
Note this happens whether I run as root or as unprivileged user.

skrashevich:
**Daemon Initialization Logic Enhanced:**
   - Added a new condition to check if the current process is a child (not started by init or not having a parent). This helps in determining whether to run the process as a daemon.
   - Improved the check for parent process existence and status, ensuring that the daemon status is set correctly based on the parent's state.

**Argument Parsing Simplification:**
   - Previous argument manipulation (clearing daemon flags in the argument list) has been removed.
   - Simplified the way the command for re-running the program is constructed, directly using the unchanged `os.Args[1:]`.

Fix #1165

### DIFF
--- a/internal/app/app.go
+++ b/internal/app/app.go
@@ -45,17 +45,7 @@ func Init() {
 		os.Exit(0)
 	}
 
-	ppid := os.Getppid()
-	if ppid == 1 {
-		daemon = false
-	} else {
-		parent, err := os.FindProcess(ppid)
-		if err != nil || parent.Pid < 1 {
-			daemon = false
-		}
-	}
-
-	if daemon {
+	if daemon && os.Getppid() != 1 {
 		if runtime.GOOS == "windows" {
 			fmt.Println("Daemon mode is not supported on Windows")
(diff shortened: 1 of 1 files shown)


AlexxIT:
Strange logic near
```go
parent, err := os.FindProcess(ppid)
if err != nil || parent.Pid < 1 {
	daemon = false
}
```

AlexxIT (approved):

AlexxIT:
Thanks! I've removed the unclear code. You can get it back in another PR if you prove its point.

AlexxIT:
https://github.com/AlexxIT/go2rtc/releases/tag/v1.9.4

PR #1195, opened 2024-06-16
fix(streams): handle missing codec_name in appendDOT function

skrashevich:
fix `SyntaxError: Syntax error in part "%!q(<nil>), title="sample_r..." (at ` error on network.html

PR #1196, opened 2024-06-16
refactor(webui): enhance network visualization in network.html

skrashevich:
**Commit Summary:**
- `www/network.html`: The visualization implementation using the vis library in `network.html` was significantly enhanced. Changes include introducing physics settings and layout configurations, handling data fetching within a function to update the network graph periodically, and utilizing error handling for fetch operations. The periodic update functionality with intervals was added to keep the visualization current. A new variable `network` was introduced to manage states, supported by better positioning and responsive design adjustments.

- `internal/streams/dot.go`: The `label` function was refactored to use `strings.Builder` for constructing the output string, enhancing performance and readability. Previously concatenated strings were replaced with builder's `WriteString` method.

**Dependency Note:**
This commit should best be merged after #1195, although it does not strictly depend on it.

### DIFF
--- a/internal/streams/dot.go
+++ b/internal/streams/dot.go
@@ -146,19 +146,20 @@ func (c *conn) host() (s string) {
 	return
 }
 
-func (c *conn) label() (s string) {
-	s = "format_name=" + c.FormatName
+func (c *conn) label() string {
+	var sb strings.Builder
+	sb.WriteString("format_name=" + c.FormatName)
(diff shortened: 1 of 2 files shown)


AlexxIT:
Network pan and zoom doesn't stored. And also sometimes networks moves down by 1 pixel on each update.

AlexxIT (approved):

AlexxIT:
Thanks!

Issue #1243, opened 2024-07-04
Hang with exec when use bash script


tsightler:
Hi @AlexxIT,

I'm back with an attempt to provide a more complete description of this issue as well as a simple configuration to reproduce.  I'm still not 100% sure of the cause or why it is more pronounced when using bash, but I'm starting to believe it could happen with any command, given enough cycles.  It's definitely more pronounced with anything that spawns extra processes, because the default behavior is to use SIGKILL, which will leave child processes running since the signal is only sent to the parent.

However, even when the primary exec process properly handles signals and kills children prior to exiting, the problem still occurs.  The simplest way to reproduce the issue is with the following go2rtc config:
```
log:
  level: debug
streams:
  stream1: exec:ffmpeg -re -i /media/BigBuckBunny.mp4 -c copy -rtsp_transport tcp -f rtsp {output}#killsignal=15
  stream2: exec:bigbuckbunny.sh {output}#killsignal=15
```

Obviously stream1 just runs ffmpeg directly, but stream 2 instead uses the following simple bash script:
```
#!/bin/bash
# This script simply runs ffmpeg to stream to RTSP
output=${1}  # URL for publishing RTSP stream

cleanup() {
    # Signal all background processes and wait for them to shutdown prior to exiting
    local pids=$(jobs -pr)
    [ -n "$pids" ] && kill $pids
    wait
    exit 0
}

# Trap signals from go2rtc so that background ffmpeg process can be killed
trap cleanup INT TERM

echo "Starting ffmpeg..." 1>&2
ffmpeg -re -i /media/BigBuckBunny.mp4 -c copy -rtsp_transport tcp -f rtsp ${output} &
wait
cleanup
```

Technically both stream1 and stream2 should behave the same, and mostly they do.  The only difference is that stream2 uses a bash script to spawn ffmpeg in the background, then traps the TERM signal and kills ffmpeg itself prior to existing.  You'll still see in the debug output that ffmpeg gets the kill signal and exits, as does the script, but, about 1 time out of 5, the cmd.wait() call will no return, and thus new connection will not cause the script to be run.  Here's what a successful exit looks like:
```
14:26:27.654 DBG [streams] stop producer url=exec:bigbuckbunny.sh {output}#killsignal=15
14:26:27.658 DBG [exec] frame=   49 fps= 21 q=-1.0 Lsize=N/A time=00:00:02.34 bitrate=N/A speed=1.02x
[...]

PR #1246, opened 2024-07-05
Fix exec race/hang on exit

tsightler:
In go2rtc 1.9.4 it is easy to have hangs on exec processes, especially if they spawn child processes, due to a race condition caused by the fact that closer calls cmd.Wait() for RTSP sessions even though cmd.Wait() has already been called in handleRTSP.  As far as I can tell, callling Wait() twice is bad practice and the behavior of doing so is undefined.  

This small patch fixes the issue by passing in the existing channel from handleRTSP to the closer so that it can share the same wait() instead of attempting to call it a second time.  For the handlePipe case, there is no initial call to cmd.Wait(), so the code behaves as previous.

I've tested various killsignals and killtimeout and this fixes all persistent hang conditions.  There is still a scenario where, sigkill may kill the parent but the secondary process keeps the producer stream alive.  I somewhat wonder if it should always return after sigkill, perhaps after a short timeout, since that should hard kill the parent and clearly the goal at that point is to stop the stream immediately, but I'm guessing this is a rare occurrence.

https://github.com/AlexxIT/go2rtc/issues/1243

### DIFF
--- a/internal/exec/closer.go
+++ b/internal/exec/closer.go
@@ -15,6 +15,7 @@ import (
 type closer struct {
 	cmd   *exec.Cmd
 	query url.Values
+	done  chan error
 }
 
 func (c *closer) Close() (err error) {
@@ -35,5 +36,14 @@ func (c *closer) Close() (err error) {
 		defer timer.Stop() // stop timer if Wait ends before timeout
 	}
 
-	return errors.Join(err, c.cmd.Wait())
+	if c.done != nil {
+		select {
+		case doneErr := <-c.done:
+			err = errors.Join(err, doneErr)
+		}
+	} else {
+		err = errors.Join(err, c.cmd.Wait())
+	}
+
+	return err
 }
--- a/internal/exec/exec.go
+++ b/internal/exec/exec.go
@@ -158,6 +158,8 @@ func handleRTSP(source string, cmd *exec.Cmd, cl io.Closer, path string) (core.P
 		done <- cmd.Wait()
 	}()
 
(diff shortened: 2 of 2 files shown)


dagleaves:
Tested this as related to the similar hang with ffmpeg audio experienced in #1254. It did not resolve the issue in that case, it just now hangs on waiting for done. Likely due to your last statement because the secondary audio stream keeps it alive. Will likely need some additional way to kill all streams.

tsightler:
> Tested this as related to the similar hang with ffmpeg audio experienced in #1254. It did not resolve the issue in that case, it just now hangs on waiting for done. Likely due to your last statement because the secondary audio stream keeps it alive. Will likely need some additional way to kill all streams for that issue

I wouldn't expect the patch I submitted to fix this issue because, if I'm not mistaken, return audio is only supported via pipe, and the code addresses only the hang when using with RTSP exec, not pipe.  I suspect the actual cause is slightly different as well, as the pipe code needs to be modified to send kill to the secondary audio as well, but I'll admit I haven't tried to track that down because my project doesn't current support return audio, although I hope to add it eventually.

AlexxIT:
@tsightler thanks for your point about `cmd.Wait()`. I've completely rewritten this code. I hope it won't be a problem now https://github.com/AlexxIT/go2rtc/commit/6fb59949a24ddd782dbbdebcdc2381a0105fc96e

tsightler:
Thanks @AlexxIT!  I'll give it a spin and see how it goes.

PR #1264, opened 2024-07-17
Install ffplay in container

OnFreund:


felipecrs:
I wonder what's the use for this. Ideally the container should be slim, and ffplay seems more like a tool to be used outside go2rtc, no?

OnFreund:
It's mentioned in the docs as a way to implement backchannel through `exec`, but is missing from the container, and getting it into the container isn't trivial (I have to use a dockerfile based on the image and point my compose file at it).

felipecrs:
Got it. Makes sense.

AlexxIT:
I thought ffplay was already in. I'll have to check

OnFreund:
I only see ffmpeg and ffprobe in the original image. I have a local dockerfile that looks like this:
```
FROM alexxit/go2rtc
RUN apk add ffplay
```

AlexxIT (approved):

AlexxIT:
Thanks!

AlexxIT:

PR #1266, opened 2024-07-18
Refactor /stream API

seydx:
This PR enhances the functionality of the `/streams` API, allowing for more flexible management of multiple sources and improving the overall behavior of the API endpoints.

Key Changes:

1. Multiple Source Management:
   - The API now supports handling multiple sources for streams.

2. `apiStreams` Rewrite:
  - <del> `GET` requests now handle "name" and "src" queries separately:.</del>
     * <del> "name" query searches for an exact stream name match..</del>
     * <del> "src" query returns all streams containing the specified source.</del>
   - `PUT` requests:
     * Allow adding multiple sources (e.g., ?name=test&src=rtsp://localhost:1234/test1&src=rtsp://localhost:1234/test2).
     * Return an error if the stream already exist (PATCH should be used for updating streams)
   - `PATCH` requests:
     * If "src" query is not defined, all sources will be removed.
     * Allow adding/removing multiple sources.
     * Return an error if the stream is not found.
   - `DELETE` requests:
     * Use "name" instead of "src" in the query for consistency with other endpoints.
     * Return an error if the stream is not found.

3. Bug Fixes:
   - Fixed an issue where the config was not updated after a PATCH request.

4. Improved Error Handling:
   - Endpoints now return both a status code and a message for invalid requests, providing clearer feedback.

These changes significantly improve the flexibility and usability of the /streams API, allowing for more complex stream management operations and providing better feedback to API consumers.

P.S. I'm pretty new to Go, so I hope I didn't mess anything up!

### DIFF
--- a/api/openapi.yaml
+++ b/api/openapi.yaml
@@ -13,6 +13,20 @@ servers:
 
 components:
   parameters:
+    stream_name_path:
+      name: name
+      in: path
+      description: Source stream name
+      required: true
+      schema: { type: string }
+      example: camera1
+    stream_src:
+      name: src
+      in: path
+      description: Stream source
+      required: true
+      schema: { type: string }
+      example: "rtsp://user:pass@10.0.0.1/stream1"
     stream_src_path:
       name: src
       in: path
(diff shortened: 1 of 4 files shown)


seydx:
Reverted some changes to minimize the overall impact, eg. the `GET` requests to `/streams` should work as before

The "only" breaking change is for DELETE request (using `name` query instead of `src`)

skrashevich:
🩼

seydx:
https://github.com/AlexxIT/go2rtc/pull/1407

PR #1283, opened 2024-07-28
fix(mdns): macos: handle address reuse and fallback for no interfaces

skrashevich:
### Overview
This PR addresses the error `ERR internal/homekit/homekit.go:128 > error="no interfaces for listen"` on macOS when running the application without root privileges.

### Problem
The mDNS client was failing to bind to network interfaces, resulting in the error mentioned above. This typically occurred when specific network interfaces were not available or the address was already in use.

### Solution
- **Address Reuse Handling**:
  - Added a check for the "address already in use" error, allowing the loop to continue and try other interfaces.
  
- **Fallback for No Interfaces**:
  - Implemented a fallback mechanism to use the system's default network interface if specific interfaces are not available.

### Impact
These changes ensure that the mDNS client can gracefully handle common networking issues on macOS, improving reliability when the application is run without root privileges.

### DIFF
--- a/pkg/mdns/client.go
+++ b/pkg/mdns/client.go
@@ -185,13 +185,21 @@ func (b *Browser) ListenMulticastUDP() error {
 	for _, ip4 := range ip4s {
 		conn, err := lc1.ListenPacket(ctx, "udp4", ip4.String()+":5353") // same port important
 		if err != nil {
+			if strings.Contains(err.Error(), "address already in use") {
+				continue
+			}
 			continue
 		}
 		b.Sends = append(b.Sends, conn)
 	}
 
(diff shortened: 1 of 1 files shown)


AlexxIT:
How can I replicate this error on my mac? I have go2rtc running with no errors. Without sudo.

pkg/mdns/client.go
```diff
@@ -372,3 +380,43 @@ func NewServiceEntries(msg *dns.Msg, ip net.IP) (entries []*ServiceEntry) {
 
 	return
 }
+
+func InterfacesIP4() ([]net.IP, error) {
```
AlexxIT: This is useless

PR #1355, opened 2024-09-16
data race for streams map

michelepra:
If concurrently calling create stream from api we can have error in write into streams map.
See https://go.dev/doc/articles/race_detector

### DIFF
--- a/internal/streams/streams.go
+++ b/internal/streams/streams.go
@@ -61,6 +61,9 @@ func New(name string, source string) *Stream {
 		return nil
 	}
 
+	streamsMu.Lock()
+	defer streamsMu.Unlock()
+
 	stream := NewStream(source)
 	streams[name] = stream
 	return stream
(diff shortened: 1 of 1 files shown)


internal/streams/streams.go
```diff
@@ -61,6 +61,9 @@ func New(name string, source string) *Stream {
 		return nil
 	}
 
+	streamsMu.Lock()
```
AlexxIT, line 64: Better to use:

```go
stream := NewStream(source)
streamsMu.Lock()
streams[name] = stream
streamsMu.Unlock()
return stream
```

michelepra: in golang style is better to defer unlock

```golang
stream := NewStream(source)
streamsMu.Lock()
defer streamsMu.Unlock()
streams[name] = stream
return stream
```

AlexxIT: No. Defer should be used wisely where it's needed. In this case, there is no reason to use it.

AlexxIT (approved):

AlexxIT:
Thanks!

AlexxIT:

PR #1356, opened 2024-09-16
proxy protocol for http api

michelepra:
If we put go2rtc behind proxy (like haproxy) we cannot use some additional feature like network panel with real source ip if stream is consumed with fmp4 over websocket.
[Proxyprotocol](https://www.haproxy.org/download/1.8/doc/proxy-protocol.txt) is made for transmit source ip/port address from proxy chain. Is transparent of application layer due it work in transport layer (tcp and udp).

With this PR can be enabled (default is disabled) and configured with two new api parameter into go2rtc.yaml file

```
api:
  proxy_protocol: <bool>
  proxy_protocol_read_timeout: <string representation of time.Duration>
```

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -16,6 +16,7 @@ require (
 	github.com/pion/srtp/v2 v2.0.18
 	github.com/pion/stun v0.6.1
 	github.com/pion/webrtc/v3 v3.2.40
+	github.com/pires/go-proxyproto v0.7.0
 	github.com/rs/zerolog v1.33.0
 	github.com/sigurn/crc16 v0.0.0-20240131213347-83fcde1e29d1
 	github.com/sigurn/crc8 v0.0.0-20220107193325-2243fe600f9f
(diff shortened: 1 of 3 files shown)


AlexxIT:
go2rtc can handle well known `X-Forwarded-For` header:
https://github.com/AlexxIT/go2rtc/blob/8399edce6ab1e51e43009464a6383e526009c73c/pkg/core/connection.go#L111-L124

michelepra:
Like reported in [wikipedia](https://en.wikipedia.org/wiki/X-Forwarded-For#Alternatives_and_variations) `PROXY protocol` is alternatives of `X-Forwarded-For` (that is supersedes by RFC7239 `forwarded` header)
Main goal of proxy protocol is that is not limited to http due it's based on transport layer and is transparent to application.

With this PR is default disabled and anyone can decide that proxyprotocol can be do the need

AlexxIT:
I see no reason to add an extra library, complicate the project code and config. Without any benefit.
If user already has reverse proxy - he doesn't have any http restrictions.

PR #1365, opened 2024-09-20
feat(api): add pprof endpoint

jrcichra:
This will help debug running go2rtc instances.

Since the default http handler is used for the api, it is automatically registered.


AlexxIT:
I don't think it's good idea of using this without additional setting.

jrcichra:
Okay I can look into putting it behind a setting.

jrcichra:
I had to replace the default mux so pprof wouldn't auto register. When you specify the `pprof` flag it registers on the api mux.

jrcichra:
Hey @AlexxIT does this change look ok?

AlexxIT:
I needs to think about this. I have design concept for access rights settings. But it will take time to implement.

SVilgelm:
The pprof endpoints are really needed for debugging purpose only, they can be enabled via build tags. Add something like this:
`proof.go`
```go
// +build pprof

package api

import "net/http/pprof"
```

I this case the pprof can be enabled by building the binary with `pprof` tag: `go build -tags pprof`

PR #1362, opened 2024-09-19
Build the docker image for linux/arm/v6

edenhaus:
Home assistant wants to use this project to improve the camera experience by offering WebRTC for any camera stream instead to high latency HLS stream type.
For this we need to build this project also for linux/arm/v6

HA PR: https://github.com/home-assistant/core/pull/124410

PR #1366, opened 2024-09-20
Revert "Build the docker image for linux/arm/v6"

edenhaus:
Reverts AlexxIT/go2rtc#1362

The easiest way is to revert it, as I solve the issue directly in HA by using the released binaries.
If a Docker image is available, I always tend to use that one, as Docker automatically selects the right platform.

I use now the following `RUN` command:
```dockerfile

# Get go2rtc binary
RUN \
    case "${BUILD_ARCH}" in \
        "aarch64") go2rtc_suffix='arm64' ;; \
        "armhf") go2rtc_suffix='armv6' ;; \
        "armv7") go2rtc_suffix='arm' ;; \
        *) go2rtc_suffix=${BUILD_ARCH} ;; \
    esac \
    && curl -L https://github.com/AlexxIT/go2rtc/releases/download/v1.9.4/go2rtc_linux_${go2rtc_suffix} --output /bin/go2rtc \
    && chmod +x /bin/go2rtc \
    # Verify go2rtc can be executed
    && go2rtc --version
```

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -159,7 +159,6 @@ jobs:
           platforms: |
             linux/amd64
             linux/386
-            linux/arm/v6
             linux/arm/v7
             linux/arm64/v8
           push: ${{ github.event_name != 'pull_request' }}


AlexxIT:
No problem. I have fixed docker file. But changes will be only with next release. Them only in master version for now.
https://hub.docker.com/r/alexxit/go2rtc/tags

PR #1407, opened 2024-10-22
Extend streams API to allow multiple sources

edenhaus:
In Home Assistant, we use the API to add streams to go2rtc, but currently, the API supports only one source. 
Currently, only the yaml config supports adding multiple sources to one stream.

This PR adds support for adding multiple sources for one stream via the API.
I choose to pass the sources via the body as source URLs can contain sensitive data like username and password, and we should avoid passing them via URL as many tools log/cache the whole URL.

I tested the PR manually as I'm new to go and didn't find any API tests in go2rtc.
Thanks in advance for the review and any feedback :)

### DIFF
--- a/internal/homekit/api.go
+++ b/internal/homekit/api.go
@@ -101,7 +101,7 @@ func apiPair(id, url string) error {
 		return err
 	}
 
-	streams.New(id, []string{conn.URL()})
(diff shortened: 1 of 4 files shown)


AlexxIT (approved):

AlexxIT:

PR #1408, opened 2024-10-22
Add the option to pass ICE servers with an async WebRTC offer

edenhaus:
Add the option to pass ICE servers with an async WebRTC offer, which are merged with the servers defined in the config. I have added the option only for the async WebRTC offer V2, but it can also be added for the sync V2 one if needed.

This is helpful if you are using a TURN server, where the credentials are only valid for a certain time.

Without this PR, the servers can only be set via the config file. Changes in the config file take effect only after restarting the go2rtc binary, which closes running streams.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -5,6 +5,7 @@ go 1.20
 require (
 	github.com/asticode/go-astits v1.13.0
 	github.com/expr-lang/expr v1.16.9
+	github.com/go-viper/mapstructure/v2 v2.2.1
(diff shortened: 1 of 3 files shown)


edenhaus:
Closing in favor of https://github.com/AlexxIT/go2rtc/commit/16e48314990d851368e57ffbe7b1afe5b01b5b15

AlexxIT:

PR #1428, opened 2024-11-02
Fix hanging webrtc streams

seydx:
Since v1.9.4 it is no longer possible to stop webrtc streams (`webrtc:ws://`). Go2RTC attempts to stop them and logs it, but on the client side the Peer Connection never disconnects, making it impossible to stop the stream.

This Pull Request reimplements the stop function for the webrtc producer, which terminates the Peer Connection.

### DIFF
--- a/pkg/webrtc/producer.go
+++ b/pkg/webrtc/producer.go
@@ -47,3 +47,13 @@ func (c *Conn) Start() error {
 	c.closed.Wait()
 	return nil
 }
+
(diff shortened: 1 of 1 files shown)


AlexxIT:
Thanks for pointing to exact place of problem. But I have fixed this with another way.

AlexxIT:
https://github.com/AlexxIT/go2rtc/releases/tag/v1.9.7

PR #1429, opened 2024-11-02
Adds support for Abode IOTA via HomeKit

los93sol:
This adds the ability to discover the Abode IOTA and pair it over HomeKit.  It also fixes the TLV8 parsing that would make this particular camera fail to playback due to it producing 0xff, 0x00 in between it's TLV's.  I updated the TLV8 parsing to handle scenarios like this generically because in this case 0xff is type 255 with length 0 so you can just handle them like normal 0 length TLV's and move on.

### DIFF
--- a/internal/homekit/api.go
+++ b/internal/homekit/api.go
@@ -72,7 +72,7 @@ func discovery() ([]*api.Source, error) {
 		log.Trace().Msgf("[homekit] mdns=%s", entry)
 
 		category := entry.Info[hap.TXTCategory]
-		if entry.Complete() && (category == hap.CategoryCamera || category == hap.CategoryDoorbell) {
+		if entry.Complete() && (category == hap.CategoryBridge || category == hap.CategoryCamera || category == hap.CategoryDoorbell) {
 			source := &api.Source{
 				Name: entry.Name,
 				Info: entry.Info[hap.TXTModel],
--- a/pkg/hap/tlv8/tlv8.go
+++ b/pkg/hap/tlv8/tlv8.go
@@ -5,7 +5,6 @@ import (
 	"encoding/base64"
 	"encoding/binary"
 	"errors"
-	"fmt"
 	"io"
 	"math"
 	"reflect"
@@ -186,57 +185,27 @@ func Unmarshal(data []byte, v any) error {
 }
 
 func unmarshalStruct(b []byte, value reflect.Value) error {
-	var waitSlice bool
+	if len(b) < 2 {
+		return nil // End of payload, nothing left to process
+	}
 
-	for len(b) >= 2 {
-		t := b[0]
-		l := int(b[1])
+	t := b[0]       // Type
+	l := int(b[1])  // Length
+	v := b[2 : 2+l] // Value (can be empty if l == 0)
 
-		// array item divider
-		if t == 0 && l == 0 {
-			b = b[2:]
-			waitSlice = true
-			continue
-		}
+	// Move the cursor past the current TLV item
+	remainder := b[2+l:]
+	tag := strconv.Itoa(int(t))
+	valueField, ok := getStructField(value, tag)
 
-		var v []byte
-
-		for {
-			if len(b) < 2+l {
-				return errors.New("tlv8: wrong size: " + value.Type().Name())
-			}
-
-			v = append(v, b[2:2+l]...)
-			b = b[2+l:]
-
-			// if size == 255 and same tag - continue read big payload
-			if l < 255 || len(b) < 2 || b[0] != t {
-				break
-			}
-
-			l = int(b[1])
-		}
-
-		tag := strconv.Itoa(int(t))
-
-		valueField, ok := getStructField(value, tag)
-		if !ok {
-			return fmt.Errorf("tlv8: can't find T=%d,L=%d,V=%x for: %s", t, l, v, value.Type().Name())
-		}
-
-		if waitSlice {
-			if valueField.Kind() != reflect.Slice {
-				return fmt.Errorf("tlv8: should be slice T=%d,L=%d,V=%x for: %s", t, l, v, value.Type().Name())
-			}
-			waitSlice = false
-		}
-
+	if ok {
 		if err := unmarshalValue(v, valueField); err != nil {
 			return err
 		}
 	}
 
-	return nil
+	// Recursively process the remaining payload
+	return unmarshalStruct(remainder, value)
 }
 
 func unmarshalValue(v []byte, value reflect.Value) error {
--- a/pkg/hap/tlv8/tlv8.go
+++ b/pkg/hap/tlv8/tlv8.go
@@ -189,14 +189,19 @@ func unmarshalStruct(b []byte, value reflect.Value) error {
 		return nil // End of payload, nothing left to process
 	}
 
-	t := b[0]       // Type
-	l := int(b[1])  // Length
-	v := b[2 : 2+l] // Value (can be empty if l == 0)
-
-	// Move the cursor past the current TLV item
-	remainder := b[2+l:]
+	t := b[0]            // Type
+	l := int(b[1])       // Length
+	v := b[2 : 2+l]      // Value (can be empty if l == 0)
+	remainder := b[2+l:] // Move cursor to the next TLV item
 	tag := strconv.Itoa(int(t))
 	valueField, ok := getStructField(value, tag)
+
+	// Accumulate fragments if the value length is 255
+	for l == 255 && len(remainder) >= 2 && remainder[0] == t {
+		l = int(remainder[1])
+		v = append(v, remainder[2:2+l]...)
+		remainder = remainder[2+l:]
+	}
 
 	if ok {
 		if err := unmarshalValue(v, valueField); err != nil {


AlexxIT:
I'm not sure your TLV changes don't break anything.

los93sol:
I’m not either lol but I did run through pairing which has TLVs larger than 255 bytes and that worked fine and I’m pretty familiar with HAP as I originally patched the PDU fragmentation issue that prevented certain BLE devices from functioning properly with the lib so I am fairly confident this change is solid.

AlexxIT:
О. Oh, great. I wrote that code completely blind. Just on a guess as to how it should work.

los93sol:
@AlexxIT I'm looking more into the HomeKit side and wanted to get your thoughts on some things.

My camera has two instances of the CameraRTPStreamManagement service and there actually is a difference in the video and audio formats offered on them.  My thought is it would be nice to be able to specify the iid in the configuration to force it to use the specific service as their streaming statuses do appear to accurately reflect which one is in use.

The other thing I was looking at is allowing the user to specify the max width and max height to try to cap out the highest negotiated resolution.  The idea with this would be for use with Frigate where I may want one stream to be lower resolution for detect and a higher resolution for record.

The last thing I was looking at is two way audio to the homekit device.  It looks like there's a handful of scenarios to differentiate around it

1. Proxying a HomeKit device to HomeKit.  I tested this and it worked fine with Apple Home.
2. Proxying any other device to HomeKit.  I have not tested this, but I do have some onvif cameras that I have two way audio working with already so I am able to test it.
3. Two-way audio between go2rtc, frigate, home assistant, etc. to the HomeKit device.  I found that frigate's UI is looking specifically for recvonly to display the speaker button to listen to the camera, and sendonly to display the microphone to at least start the enablement of two way audio.  I also am able to conditionally add the sendonly media based on the presence of the speaker service on the camera.  I'm a bit confused here though as it seems like the direction should be DirectionSendRecv in this case as the audio codec is the same in both directions so trying to understand if frigate's logic should account for DirectionSendRecv as well or if I should actually add the same media twice.

seydx:
@los93sol 

I'm also interested in two-way audio. I haven't tested it yet, but do native homekit cameras with two-way audio proxied from go2rtc to homekit still have a working talkback button?

In the case of non-native homekit cameras with two-way audio, like a tapo camera, there is no talkback button, right?

AlexxIT:
@los93sol 
1. I have not seen any cameras that have a different CameraRTPStreamManagement. Can you send me the response from the camera? Preferably in raw JSON format.
2. Resolution is a nice feature. I've already added a customizable bitrate. Resolution can be added, it's not complicated.
3. Two way audio from and to HomeKit is not supported yet. I just didn't have time to implement it.

los93sol:
I’ll try to get that payload for you.

in the meantime I’ve been trying to get two way audio going and I am hitting some walls that I’m not sure how to properly address.

The scenario I’m targeting is getting it to work from webrtc to the HomeKit device.  The problem is the protocol does not even support opus/48000/2 and the non standard aac won’t work either.  The best I’ve been able to do is add opus/48000/2 to the producer manually and get the pipelining to actually route the audio packets to me, but then I need to transcode and the only codecs I think can actually be supported are opus formats unless you’ve compiled ffmpeg with fdk_aac.  The other route i tried was to just add support for opus/16000/1 (what my camera supports) to the webrtc consumer and I was able to get the pipelining setup correctly that way, but haven’t been able to get the rtp packets to actually land in the sender’s handler.  It seems like from a compatibility position that what I really need is a layer between similar to how ffmpeg can be used to convert audio from the device to the browser in a supported format, but in this case it needs to go the other way, from browser to the device.  Curious what your thoughts are on the best way to handle this. 



> @los93sol
> 
> 1. I have not seen any cameras that have a different CameraRTPStreamManagement. Can you send me the response from the camera? Preferably in raw JSON format.
> 2. Resolution is a nice feature. I've already added a customizable bitrate. Resolution can be added, it's not complicated.
> 3. Two way audio from and to HomeKit is not supported yet. I just didn't have time to implement it.

I’ll try to get that payload for you.

in the meantime I’ve been trying to get two way audio going and I am hitting some walls that I’m not sure how to properly address.

The scenario I’m targeting is getting it to work from webrtc to the HomeKit device.  The problem is the protocol does not even support opus/48000/2 and the non standard aac won’t work either.  The best I’ve been able to do is add opus/48000/2 to the producer manually and get the pipelining to actually route the audio packets to me, but then I need to transcode and the only codecs I think can actually be supported are opus formats unless you’ve compiled ffmpeg with fdk_aac.  The other route i tried was to just add support for opus/16000/1 (what my camera supports) to the webrtc consumer and I was able to get the pipelining setup correctly that way, but haven’t been able to get the rtp packets to actually land in the sender’s handler.  It seems like from a compatibility position that what I really need is a layer between similar to how ffmpeg can be used to convert audio from the device to the browser in a supported format, but in this case it needs to go the other way, from browser to the device.  Curious what your thoughts are on the best way to handle this.

AlexxIT:
Apple uses non standard OPUS format. You can't use it anywhere without fixing.

AlexxIT:
Maybe this will helps you https://github.com/AlexxIT/go2rtc/blob/master/pkg/opus/homekit.go

los93sol:
Here's the json payload you wanted, this thing has a ton of accessories connected so I cleaned those out for brevity.  If you want to see those as well then just lmk and I can send them

```
{
  "accessories":
  [
    {
      "aid":1,
      "services":
      [
        {
          "iid":1, "type":"3E", "primary":false, "hidden":false, "linked":[],
          "characteristics":
          [
            { "iid":2, "type":"14", "format":"bool", "perms":["pw"] },
            { "iid":3, "type":"20", "format":"string", "value":"abode", "perms":["pr"], "ev":false },
            { "iid":4, "type":"21", "format":"string", "value":"Bridge", "perms":["pr"], "ev":false },
            { "iid":5, "type":"23", "format":"string", "value":"IOTA-[REDACTED]", "perms":["pr"], "ev":false },
            { "iid":6, "type":"30", "format":"string", "value":"[REDACTED]", "perms":["pr"], "ev":false },
            { "iid":7, "type":"52", "format":"string", "value":"2.0.20", "perms": ["pr"], "ev":false },
            { "iid":8, "type":"53", "format":"string", "value":"1.0", "perms": ["pr"], "ev":false },
            { "iid":9, "type":"34AB8811-AC7F-4340-BAC3-FD6A85F9943B", "format":"string", "value":"2.0.1;16A75", "perms": ["pr","hd"], "ev":false }
          ]
        },
        {
          "iid":16, "type":"A2", "primary":false, "hidden":false, "linked": [],
          "characteristics":
          [
            { "iid":18, "type":"37", "format":"string", "value":"1.1.0", "perms":["pr"], "ev":false }
          ]
        },
        {
          "iid":816, "type":"110", "primary":false, "hidden":false, "linked":[],
          "characteristics":
          [
            { "iid":819, "type":"120", "format":"tlv8", "value":"AQEA", "perms":["pr","ev"], "ev":false },
            { "iid":820, "type":"114", "format":"tlv8", "value":"AaUBAQACDAEBAQIBAgMBAAQBAAMLAQKABwICOAQDAR7/AAMLAQIABQIC0AIDAR7/AAMLAQKAAgICaAEDAR7/AAMLAQLgAQICDgEDAR7/AAMLAQJAAQICtAADAR7/AAMLAQIABQICwAMDAR7/AAMLAQIABAICAAMDAR7/AAMLAQKAAgIC4AEDAR7/AAMLAQLgAQICaAEDAR7/AAMLAQJAAQIC8AADAR4=", "perms":["pr"], "ev":false },
            { "iid":821, "type":"115", "format":"tlv8", "value":"AQ4BAQMCCQEBAQIBAAMBAQIBAA==", "perms":["pr"], "ev":false },
            { "iid":822, "type":"116", "format":"tlv8", "value":"AgEA/wACAQH/AAIBAg==", "perms":["pr"], "ev":false },
            { "iid":823, "type":"118", "format":"tlv8", "value":"", "perms":["pr","pw"], "ev":false },
            { "iid":824, "type":"117", "format":"tlv8", "value":"", "perms":["pr","pw"], "ev":false }
          ]
        },
        { "iid":832, "type":"110", "primary":false, "hidden":false, "linked":[],
          "characteristics":
          [
            { "iid":835, "type":"120", "format":"tlv8", "value":"AQEA", "perms":["pr","ev"], "ev":false },
            { "iid":836, "type":"114", "format":"tlv8", "value":"AYcBAQACDAEBAQIBAgMBAAQBAAMLAQIABQIC0AIDAR7/AAMLAQKAAgICaAEDAR7/AAMLAQLgAQICDgEDAR7/AAMLAQJAAQICtAADAR7/AAMLAQIABAICAAMDAR7/AAMLAQKAAgIC4AEDAR7/AAMLAQLgAQICaAEDAR7/AAMLAQJAAQIC8AADAR4=", "perms":["pr"], "ev":false },
            { "iid":837, "type":"115", "format":"tlv8", "value":"AQ4BAQMCCQEBAQIBAAMBAQIBAA==", "perms":["pr"], "ev":false },
            { "iid":838, "type":"116", "format":"tlv8", "value":"AgEA/wACAQH/AAIBAg==", "perms":["pr"], "ev":false },
            { "iid":839, "type":"118", "format":"tlv8", "value":"", "perms":["pr","pw"], "ev":false },
            { "iid":840, "type":"117", "format":"tlv8", "value":"", "perms":["pr","pw"], "ev":false }
          ]
        },
        { "iid":848, "type":"112", "primary":false, "hidden":false, "linked":[],
          "characteristics":
          [
            { "iid":850, "type":"23", "format":"string", "value":"Microphone", "perms":["pr"], "ev":false },
            { "iid":851, "type":"11A", "format":"bool", "value":0, "perms":["pr","pw","ev"], "ev":false },
            { "iid":852, "type":"119", "format":"uint8", "value":50, "perms":["pr","pw","ev"], "ev":false, "unit":"percentage", "minValue":0, "maxValue":100, "minStep":1 }
          ]
        },
        {
          "iid":864, "type":"113", "primary":false, "hidden":false, "linked":[],
          "characteristics":
          [
            { "iid":866, "type":"23", "format":"string", "value":"Speaker", "perms":["pr"], "ev":false },
            { "iid":867, "type":"11A", "format":"bool", "value":0, "perms":["pr","pw","ev"], "ev":false },
            { "iid":868, "type":"119", "format":"uint8", "value":50, "perms":["pr","pw","ev"], "ev":false, "unit":"percentage", "minValue":0, "maxValue":100, "minStep":1 }
          ]
        },
        {
          "iid":512, "type":"7E", "primary":true, "hidden":false, "linked":[],
          "characteristics":
          [
            { "iid":513, "type":"66", "format":"uint8", "value":3, "perms":["pr","ev"], "ev":false, "minValue":0, "maxValue":4, "minStep":1 },
            { "iid":514, "type":"67", "format":"uint8", "value":3, "perms":["pr","pw","ev"], "ev":false, "minValue":0, "maxValue":3, "minStep":1 }
          ]
        },
        {
          "iid":353, "type":"85", "primary":false, "hidden":false, "linked":[],
          "characteristics":
          [
            { "iid":354, "type":"22", "format":"bool", "value":0, "perms":["pr","ev"], "ev":false }
          ]
        }
      ]
    }
  ]
}	
```

los93sol:
I have two way audio working now from webrtc to the camera.  The way I did this is to just slip opus/48000/2 as a sendonly in the producer and implemented AddTrack in the producer so it will match up and basically just slam the packet as is at the audioSession.  The only thing I did have to do is set the PayloadType to 110 and swap the SSRC on the packet with the one that we told HomeKit about when exchanging endpoints.  I didn't see it until I did a packet capture, but despite setting the PayloadType and SSRC in the handler, they were not set on the wire because inside WriteRTP there's....

	clone := rtp.Packet{
		Header: rtp.Header{
			Version:        2,
			Marker:         packet.Marker,
			PayloadType:    s.PayloadType,
			SequenceNumber: packet.SequenceNumber,
			Timestamp:      packet.Timestamp,
			SSRC:           s.Local.SSRC,
		},
		Payload: packet.Payload,
	}
	
	I changed that to just copy it off the packet, but my question really is were those taken off the session intentionally or was that just not a concern previously?

AlexxIT:
I don't understand what part of the code you are referring to.

pkg/srtp/session.go
```diff
@@ -83,10 +83,10 @@ func (s *Session) WriteRTP(packet *rtp.Packet) (int, error) {
 		Header: rtp.Header{
 			Version:        2,
 			Marker:         packet.Marker,
-			PayloadType:    s.PayloadType,
+			PayloadType:    packet.PayloadType,
```
los93sol, line 86: This line and a few lines down is what was preventing me from changing the payload type and ssrc on the rtp packets

AlexxIT:
We set these values when we set up the connection. I don't see why they need to be changed.
https://github.com/AlexxIT/go2rtc/blob/5cafc05e130f279a48b46411868ec92567963a86/pkg/homekit/consumer.go#L82-L127

los93sol:
Ah, it's because the HomeKit producer doesn't call that, from what I see that's only in play when your running the HomeKit module to proxy a device.  I'll take a closer look and see if I can bring the producer inline.

AlexxIT:
Sorry, haven't had time to look into this PR yet.

los93sol:
Have you had a chance to check this out yet?

AlexxIT:
In the queue https://github.com/AlexxIT/go2rtc/pulls/assigned/AlexxIT

PR #1438, opened 2024-11-07
Fix "panic: send on closed channel"

huynhquangtoan:


pkg/core/track.go
```diff
@@ -99,6 +99,11 @@ func NewSender(media *Media, codec *Codec) *Sender {
 	s.Input = func(packet *Packet) {
 		// writing to nil chan - OK, writing to closed chan - panic
 		s.mu.Lock()
+		if s.isClosed {
```
SVilgelm, line 102: I would recommend to refactor function to use `defer s.mu.Unlock()`, so it would be closed anyway and the remove all other `s.mu.Unlock()` instances

AlexxIT:

PR #1450, opened 2024-11-11
Lower codec not matched error for ffmpeg to debug

edenhaus:
In Home Assistant, we add for all streams the following additional line `"ffmpeg:{stream_name}#audio=opus"` to have audio on WebRTC as most cameras are only supporting aac.

If you use a no-audio camera with the above ffmpeg config, you will see anytime you start the stream the following log entry: `09:32:51.973 ERR [rtsp] error="streams: codecs not matched: H264 => ANY" stream=camera.x`

This PR downgrade the level to debug only when ffmpeg is the source of the RTSP stream

### DIFF
--- a/internal/ffmpeg/ffmpeg.go
+++ b/internal/ffmpeg/ffmpeg.go
@@ -223,6 +223,8 @@ func parseArgs(s string) *ffmpeg.Args {
 			s += "?video&audio"
 		}
 		s += "&source=ffmpeg:" + url.QueryEscape(source)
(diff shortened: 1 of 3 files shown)


AlexxIT (approved):

AlexxIT:
Thanks. I've made some edits to the implementation.

AlexxIT:
https://github.com/AlexxIT/go2rtc/releases/tag/v1.9.7

PR #1511, opened 2024-12-16
Accept rtsp client without interleaved parameter

### LINKED ISSUE #1510: Got Unsupported transport with Genetec
Hello,

I have a RTSP client (Genetec, a CCTV commercial software) connecting to go2rtc.
However, it got "461 Unsupported transport".

The SETUP packet sent by Genetec:
```
SETUP rtsp://192.168.1.1:554/Camera1/trackID=0 RTSP/1.0
CSeq: 2
User-Agent: OmnicastRTSPClient/1.0
Transport: RTP/AVP/TCP;unicast
```

It seems the issue is caused by the missing `interleaved` parameter, which is expected by go2rtc ([see here](https://github.com/AlexxIT/go2rtc/blob/8ecaabfce9b563f3fb6a0025531add7786865338/pkg/rtsp/server.go#L152)).

However, the `interleaved` parameter does not seem to be required by the [RFC 2326](https://www.ietf.org/rfc/rfc2326.txt).

Are there any specific reasons for requiring the `interleaved` parameter?
On the contrary, I’d be happy to open a PR to remove this condition.

Thanks

fmcloudconsulting:
Fix #1510

### DIFF
--- a/pkg/rtsp/server.go
+++ b/pkg/rtsp/server.go
@@ -156,7 +156,7 @@ func (c *Conn) Accept() error {
 
 				c.session = core.RandString(8, 10)
 				c.state = StateSetup
-			
+
 				if c.mode == core.ModePassiveConsumer {
 					if i := reqTrackID(req); i >= 0 && i < len(c.Senders) {
 						// mark sender as SETUP
--- a/pkg/rtsp/server.go
+++ b/pkg/rtsp/server.go
@@ -141,17 +141,14 @@ func (c *Conn) Accept() error {
 			}
 
 		case MethodSetup:
-			tr := req.Header.Get("Transport")
-
 			res := &tcp.Response{
 				Header:  map[string][]string{},
 				Request: req,
 			}
 
 			// Test if client requests TCP transport, otherwise return 461 Transport not supported
-			// This allows smart clients who initially requested UDP to fall back on TCP transport.
-			if strings.HasPrefix(tr, "RTP/AVP/TCP") {
-
+			// This allows smart clients who initially requested UDP to fall back on TCP transport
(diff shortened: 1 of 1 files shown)


AlexxIT:
Your changes are incorrect. The client can come for UDP transport.

fmcloudconsulting:
The [existing code](https://github.com/AlexxIT/go2rtc/pull/1511/files#diff-c2c0ef0da16d5b632204ca43fcaa7d3b513da1d9705fa915da2c3faf031d5ce5L151) already expects TCP-only transport, so there is no regression with this PR?

AlexxIT:
But it return `461 Unsupported transport` error, so smart client can change UDP to TCP request. Many clients support this out of box.

fmcloudconsulting:
Got the point. Should be fixed now.

AlexxIT:
It's wrong again. Some clients don't send unicast word. I think only first prefix should be enough.

fmcloudconsulting:
Fixed.
By the way, there was a typo between tr and transport variables, fixed!
I'll build & deploy on my side.

felipecrs:
@fmcloudconsulting did the deploy on your side happened successfully? Is this PR working as expected?

AlexxIT (approved):

AlexxIT:
Thanks!

fmcloudconsulting:
Hello,

After hours of effort we were unable to get Genetec to work in RTSP over TCP.
Only RTSP over UDP seemed possible.
So we used [mediamtx](https://github.com/bluenviron/mediamtx) as restream server.

However, the patch proposed here has been tested for other clients and worked well.
Thanks for merging!

AlexxIT:
It seems that the patch is useless because of the initially bad software.
I was responding in the issue that sw is not implementing the RTSP specification correctly.

AlexxIT:

PR #1522, opened 2024-12-25
Improve RTSP server authentication handling and auditing

subbyte:
In the latest version (v1.9.7), when using the RTSP module/server, a client fails authentication silently, which makes it very vulnerable to [RTSP brute force attack](https://github.com/anpa1200/RTSP-brute-force-tool):
- There is no log of failed RTSP authentication
- An attacker can test multiple RTSP requests with different credentials in a single TCP connection

This RP adds "failed auth" logging for rtsp server. In addition, the PR prevents an attacker packs many brute force authentication attempts into one TCP connection. After one authentication failed, the TCP connection will terminate, forcing the attacker to start a new TCP connection for the next brute force attempt, which makes it possible to be identified and blocked by a firewall like [pf](https://home.nuug.no/~peter/pf/en/bruteforce.html).

Note `ffmpeg` and `ffmpeg`-based players, e.g., `mpv`, have a special authentication behavior---even if username/password is provided in the URL, the first request does not include it, resulting in a failed authentication, while all subsequent requests (in the same TCP connection) add the authentication field in the header, resulting in successful authentications.

### DIFF
--- a/internal/rtsp/rtsp.go
+++ b/internal/rtsp/rtsp.go
@@ -1,6 +1,7 @@
 package rtsp
 
 import (
+	"fmt"
 	"io"
 	"net"
 	"net/url"
@@ -237,7 +238,10 @@ func tcpHandler(conn *rtsp.Conn) {
 	})
 
 	if err := conn.Accept(); err != nil {
-		if err != io.EOF {
+		if err == rtsp.FailedAuth {
+			rAddr := conn.Connection.RemoteAddr
+			log.Warn().Msg(fmt.Sprintf("[rtsp] failed authentication from %s", rAddr))
+		} else if err != io.EOF {
 			log.WithLevel(level).Err(err).Caller().Send()
 		}
 		if closer != nil {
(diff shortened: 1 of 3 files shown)


internal/rtsp/rtsp.go
```diff
@@ -237,7 +238,10 @@ func tcpHandler(conn *rtsp.Conn) {
 	})
 
 	if err := conn.Accept(); err != nil {
-		if err != io.EOF {
+		if err == rtsp.FailedAuth {
+			rAddr := conn.Connection.RemoteAddr
+			log.Warn().Msg(fmt.Sprintf("[rtsp] failed authentication from %s", rAddr))
```
SVilgelm, line 243: please use Msgf to avoid unnecessary string formatting and memory allocations if warn level is disabled.
And since the logger is structured, you can put the address to a separate field, so later the logs can be filtered by address field.
```suggestion
			log.Warn().Str("remote_addr", conn.Connection.RemoteAddr).Msg("[rtsp] failed authentication")
```

subbyte: Thanks! Fixed.

AlexxIT:
Thanks!

AlexxIT:

Issue #1174, opened 2024-06-08
Tapo Hub H200 with D230 Doorbell 


cavefire:
According to the pull request https://github.com/AlexxIT/go2rtc/pull/1029, H200 sub devices should now be supported. However, I have been unable to configure this correctly.

H200 Firmware: 1.2.21 Build 20230731 rel.51223
D230 Firmware: 1.1.15 Build 20240417 rel.61958
go2rtc Version: 1.9.3

My configuration looks like this:

```yaml
hub: 
  - tapo://admin:Sha265HashOfCloudPW-uppercase@10.0.30.215/?deviceId=idOfDevice
```

(Note: I have entered the correct password and device ID in my configuration.)

Below is the HTTP trace from go2rtc to the hub (the hashes are slightly redacted):
```http
POST /stream?deviceId=idOfDevice HTTP/1.1
Host: 10.0.30.215:8800
User-Agent: Go-http-client/1.1
Content-Length: 0
Content-Type: multipart/mixed; boundary=--client-stream-boundary--

HTTP/1.1 401 Unauthorized
Content-Length: 14
Cache-Control: no-cache
Connection: close
WWW-Authenticate: Digest realm="TP-Link IP-Camera",qop="auth",nonce="XXX65511401ff8747a712ce1f3b3cXXX",opaque="XXX047137638bef3555d5c7c6d9fdXXX",algorithm="MD5"
Set-Cookie: TP_HTTP_COOKIE=XXX65511401ff8747a712ce1f3b3cXXX; path=/

HTTP ERROR 401

POST /stream?deviceId=idOfDevice HTTP/1.1
Host: 10.0.30.215:8800
User-Agent: Go-http-client/1.1
Content-Length: 0
Authorization: Digest username="admin", realm="TP-Link IP-Camera", nonce="XXX65511401ff8747a712ce1f3b3cXXX", uri="/stream?deviceId=idOfDevice", qop=auth, nc=00000001, cnonce="XXXaVLv6YDgygdlhBiiGGs6ypw23wXXX", response="XXX0f5bb80a4330fff81ceb1923c5XXX", opaque="XXX047137638bef3555d5c7c6d9fdXXX", algorithm=MD5
Content-Type: multipart/mixed; boundary=--client-stream-boundary--

HTTP/1.1 401 Unauthorized
Content-Length: 14
Cache-Control: no-cache
Connection: close
[...]

PR #1543, opened 2025-01-08
Fix H200 + D230 Doorbell stream

cavefire:
As stated in my issue #1174 streaming the D230 doorbell was not possible.

Running firmware versions:
H200: 1.3.6 Build 20240829 rel.71119
D230: 1.1.17 Build 20240731 rel.53511

This PR is implementing the fix suggested by @aprilmaccydee (as stated in https://github.com/AlexxIT/go2rtc/issues/1174#issuecomment-2247587646). So all credits to them for figuring that out and sharing.

Using "io.Copy" reads the full body before discarding it, while "Body.Close()" only closes the body. I am not quite sure, why that makes a difference, but it does.

### DIFF
--- a/pkg/tapo/client.go
+++ b/pkg/tapo/client.go
@@ -291,7 +291,7 @@ func dial(req *http.Request, brand, username, password string) (net.Conn, *http.
 	if err != nil {
 		return nil, nil, err
 	}
-	_ = res.Body.Close() // ignore response body
+	_, _ = io.Copy(io.Discard, res.Body) // ignore response body
 
 	auth := res.Header.Get("WWW-Authenticate")
 


SVilgelm:
> I am not quite sure, why that makes a difference, but it does.

Because it discards the leftovers in the network buffer. It is pretty common practice in go to use io.Copy and then close the body of a request

pkg/tapo/client.go
```diff
@@ -291,7 +291,7 @@ func dial(req *http.Request, brand, username, password string) (net.Conn, *http.
 	if err != nil {
 		return nil, nil, err
 	}
-	_ = res.Body.Close() // ignore response body
+	_, _ = io.Copy(io.Discard, res.Body) // ignore response body
```
SVilgelm, line 294: This is great, but you still need to close the body
```suggestion
	_, _ = io.Copy(io.Discard, res.Body) // discard leftovers
	_ = res.Body.Close() // ignore response body
```

cavefire: Thanks for the clarification and suggestion.
Still works after the change, so added it to this PR.

AlexxIT:
Looks like your right
- https://pkg.go.dev/net/http#Response
- https://stackoverflow.com/questions/60931535/is-resp-body-close-required-if-i-dont-need-response

SVilgelm:
it would be better to refactor all such cases instead of just this one, but can be done in separate PRs

AlexxIT:
Thanks!

AlexxIT:

PR #1567, opened 2025-01-24
Add Ring camera integration

seydx:
Added Ring camera integration that allows users to connect their Ring cameras. Users can authenticate either via email/password (including 2FA support) or by providing a refresh token. (see below)

go2rtc_ring

After successful authentication, all available Ring cameras are displayed with their corresponding source configuration (`ring:?device_id=XXX&refresh_token=XXX`). These source configurations can then be used to add individual Ring cameras to go2rtc, which handles the WebRTC connection to stream the camera feed. Two-Way audio is also supported.

A separate producer was added for snapshot endpoints (`ring:?device_id=XXX&refresh_token=XXX&snapshot`) to directly obtain a snapshot from the Ring API

The Ring integration implementation is based on the [ring-client-api](https://github.com/dgreif/ring) library

<hr>

Question: When making requests to Ring's API, sometimes a new refresh token is returned. What would be the best approach to handle and persist these updated refresh tokens in go2rtc's configuration @AlexxIT  ?

Regarding to [ring-client-api wiki](https://github.com/dgreif/ring/wiki/Refresh-Tokens#refresh-token-updates), it is only important for push notification, which is not used here..

> @ring-client-api
> It may seem possible to simply re-use the initially generated token multiple times, and indeed it will generally work for authentication even if newer tokens are not saved...

### DIFF
--- a/internal/ring/init.go
+++ b/internal/ring/init.go
@@ -1,0 +1,47 @@
+package ring
+
+import (
+	"net/http"
+
+	"github.com/AlexxIT/go2rtc/internal/api"
+	"github.com/AlexxIT/go2rtc/internal/streams"
+	"github.com/AlexxIT/go2rtc/pkg/core"
+	"github.com/AlexxIT/go2rtc/pkg/ring"
+)
+
+func Init() {
+	streams.HandleFunc("ring", func(source string) (core.Producer, error) {
+		return ring.Dial(source)
+	})
+
+	api.HandleFunc("api/ring", apiRing)
+}
+
+func apiRing(w http.ResponseWriter, r *http.Request) {
+	query := r.URL.Query()
+	refreshToken := query.Get("refresh_token")
+
+	ringAPI, err := ring.NewRingRestClient(ring.RefreshTokenAuth{RefreshToken: refreshToken}, nil)
+	if err != nil {
+		http.Error(w, err.Error(), http.StatusInternalServerError)
+		return
+	}
+
+	devices, err := ringAPI.FetchRingDevices()
+	if err != nil {
+		http.Error(w, err.Error(), http.StatusInternalServerError)
+		return
+	}
+
+	var items []*api.Source
+
+	for _, camera := range devices.AllCameras {
+		query.Set("device_id", camera.DeviceID)
+
+		items = append(items, &api.Source{
+			Name: camera.Description, URL: "ring:?" + query.Encode(),
+		})
+	}
+
+	api.ResponseSources(w, items)
+}
--- a/internal/ring/init.go
+++ b/internal/ring/init.go
@@ -84,9 +84,17 @@ func apiRing(w http.ResponseWriter, r *http.Request) {
     var items []*api.Source
     for _, camera := range devices.AllCameras {
         cleanQuery.Set("device_id", camera.DeviceID)
+
+        // Stream source
         items = append(items, &api.Source{
             Name: camera.Description,
             URL:  "ring:?" + cleanQuery.Encode(),
+        })
+
+        // Snapshot source
+        items = append(items, &api.Source{
+            Name: camera.Description + " Snapshot",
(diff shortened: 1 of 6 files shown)


AlexxIT:
Thanks! You can safely change config yaml, like HomeKit integration do it for pairings:
https://github.com/AlexxIT/go2rtc/blob/fc02e6f4a53bcc8c7409ae68c809caa09b3da5fe/internal/homekit/api.go#L98-L107

gambalaya:
@seydx I have been using this ring camera integration for over a month now (from a nightly build), and it has been working great, thanks! However yesterday for the first time the refresh token that I had been using for the last month became unauthorized and I had to regenerate again via the Add menu. I saw your comment above about reusing the old token shouldn't matter since it is only important for push notifications - but apparently the old token eventually rolls off somewhere... Wondering if there could be a way to auto-refresh so I don't have to redo it manually if it needs it?

seydx:
> @seydx I have been using this ring camera integration for over a month now (from a nightly build), and it has been working great, thanks! However yesterday for the first time the refresh token that I had been using for the last month became unauthorized and I had to regenerate again via the Add menu. I saw your comment above about reusing the old token shouldn't matter since it is only important for push notifications - but apparently the old token eventually rolls off somewhere... Wondering if there could be a way to auto-refresh so I don't have to redo it manually if it needs it?

yeah you re right, experienced the same, will make a new PR to auto update the refresh token

gambalaya:
> yeah you re right, experienced the same, will make a new PR to auto update the refresh token

Thanks @seydx - as a workaround I now automatically update the token by extracting the homebridge token and patching the go2rtc ring source each time (only possible because I am also using the homebridge ring plugin of course):

> token="$(jq -r '.platforms[].refreshToken | values | @uri' /homebridge/config.json)"
curl --request PATCH "http://xxx:1984/api/streams?name=ring&src=ring:%3Fdevice_id=xxx%26refresh_token=$token"

Feierdevil:
Hello,

I dont understand how i can add the ring cam.

The Ring Login are correct and i see the devID with the refreshToken but i dont know what or where i do this.

Can anybody Help me?

Thx

seydx:
> Hello,
> 
> I dont understand how i can add the ring cam.
> 
> The Ring Login are correct and i see the devID with the refreshToken but i dont know what or where i do this.
> 
> Can anybody Help me?
> 
> Thx

copy the url like in the image above, and create a new stream with it in go2rtc

laurenttaieb:
Hi go2rtc team.  I have finally managed to not get an error message by setting up the configuration file this way:
streams:
   CamName: 
       - ring: ?device_id=(deviceID)&refresh_token=(my refresh token)
       -
  
I am using v1.9.9. so the integration shall work.
But the camera shows a black screen. Any idea on what I may have  missed ?

Thanks
Laurent

-> found the issue. The refresh token has changed...

laurenttaieb:
> see

This is not the end of the story.
when I click on stream on the web interface, the actual camera doesn't show up all the time, in more than 50% of cases, I have a back scree, with a MSE sign on the top right instead of the RTC one and when the image truly appears, it's only a snapshot not a live one.. Has anyone the same type of issue (I'm not using the "snapshot" version of the Token.
Thanks

gambalaya:
> > see
> 
> This is not the end of the story. when I click on stream on the web interface, the actual camera doesn't show up all the time, in more than 50% of cases, I have a back scree, with a MSE sign on the top right instead of the RTC one and when the image truly appears, it's only a snapshot not a live one.. Has anyone the same type of issue (I'm not using the "snapshot" version of the Token. Thanks

I started getting a black screen a few weeks back. Maybe it is an H265 vs H264 codec issue? Tsightler's ring-mqtt project continues to work for me though, which I think tells Ring it is a Firefox client so it always delivers H264.

PR #1568, opened 2025-01-25
Ring: Fix snapshot producer MarshalJSON and prevent nil reference during stop

seydx:
In my previous PR, I forgot to correctly execute MarshalJSON on a snapshot producer. As a result, for example, the `/probe` call didn't work on it.

Also, during my testing, I noticed that in rare cases the following error occurs:

```
DBG [streams] start producer url=ring:?device_id=XXX&refresh_token=XXX&snapshot
DBG [streams] stop producer url=ring:?device_id=XXX&refresh_token=XXX&snapshot

[signal SIGSEGV: segmentation violation code=0x2 addr=0x28 pc=0x102d10330]

goroutine 891 [running]:
github.com/AlexxIT/go2rtc/pkg/ring.(*Client).Start(0x5?)
	github.com/AlexxIT/go2rtc/pkg/ring/client.go:504 +0x20
github.com/AlexxIT/go2rtc/internal/streams.(*Producer).worker(0x140004a2cf0, {0x102fd9128?, 0x1400047a540?}, 0xd)
	github.com/AlexxIT/go2rtc/internal/streams/producer.go:161 +0x34
created by github.com/AlexxIT/go2rtc/internal/streams.(*Producer).start in goroutine 879
	github.com/AlexxIT/go2rtc/internal/streams/producer.go:157 +0x1b0
```

This PR should fix this by no longer setting the producer to `nil` during a stop (as is the case with other producers)

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -102,14 +102,14 @@ jobs:
         env: { GOOS: freebsd, GOARCH: amd64 }
         run: go build -ldflags "-s -w" -trimpath
       - name: Upload go2rtc_freebsd_amd64
-        uses: actions/upload-artifact@v3
+        uses: actions/upload-artifact@v4
         with: { name: go2rtc_freebsd_amd64, path: go2rtc }
 
       - name: Build go2rtc_freebsd_arm64
         env: { GOOS: freebsd, GOARCH: arm64 }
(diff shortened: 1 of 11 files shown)


AlexxIT:
Thanks!

AlexxIT:

PR #1589, opened 2025-02-10
fix: ONVIF client

seydx:
This PR fixes two ONVIF camera compatibility issues:

1. Some cameras (Tapo with latest fw) send malformed HTTP responses with garbage/null bytes before the actual XML content. Fixed by:
   - Switching from http.Client to direct TCP connection
   - Looking for "<?xml" marker to find start of actual content
   - Using manual HTTP request writing
   - Setting default timeout of 5 seconds

```
Post "http://192.168.178.131:2020/onvif/device_service": net/http: HTTP/1.x transport connection broken: malformed HTTP status code "/onvif/device_service\x00HTTP/1.1\x00\x00Host\x00"
panic: Post "http://192.168.178.131:2020/onvif/device_service": net/http: HTTP/1.x transport connection broken: malformed HTTP status code "/onvif/device_service\x00HTTP/1.1\x00\x00Host\x00"
```

2. Ensure compatibility with all ONVIF cameras by:
   - Setting standard ONVIF paths as defaults (/onvif/media_service, /onvif/imaging_service)
   - Using advertised XAddr values from GetCapabilities when available

Testing: Verified working with cameras that previously failed with:
- "malformed HTTP status code" errors
- "unsupported service" errors

### DIFF
--- a/pkg/onvif/client.go
+++ b/pkg/onvif/client.go
@@ -3,9 +3,10 @@ package onvif
 import (
 	"bytes"
 	"errors"
+	"fmt"
 	"html"
 	"io"
-	"net/http"
+	"net"
 	"net/url"
 	"regexp"
 	"strings"
@@ -37,13 +38,22 @@ func NewClient(rawURL string) (*Client, error) {
 		client.deviceURL = baseURL + u.Path
 	}
 
+    // Set default media URL before trying to get capabilities
+    client.mediaURL = baseURL + "/onvif/media_service"
+    client.imaginURL = baseURL + "/onvif/imaging_service"
+
 	b, err := client.DeviceRequest(DeviceGetCapabilities)
 	if err != nil {
 		return nil, err
 	}
 
-	client.mediaURL = FindTagValue(b, "Media.+?XAddr")
-	client.imaginURL = FindTagValue(b, "Imaging.+?XAddr")
+    // Update URLs if found in capabilities
+    if mediaAddr := FindTagValue(b, "Media.+?XAddr"); mediaAddr != "" {
+        client.mediaURL = mediaAddr
+    }
+    if imagingAddr := FindTagValue(b, "Imaging.+?XAddr"); imagingAddr != "" {
+        client.imaginURL = imagingAddr
+    }
 
 	return client, nil
 }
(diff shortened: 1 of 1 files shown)


vinicius73:
I have the same issue.
I've tested this code, this PR solve it, thanks.

@AlexxIT could you merge it?

AlexxIT:
PR will be accepted after a careful review.

AlexxIT:
1. I don't like idea support malformed responses. Especially for one buggy camera. Hopefully Tapo will fix it themselves.
2. Support "unsupported service" is OK.

Can you fix or split PR?

felipecrs:
> 1. I don't like idea support malformed responses. Especially for one buggy camera. Hopefully Tapo will fix it themselves.

@AlexxIT you didn't ask but I would like to advocate in @seydx's favor:

I think one of the best things about go2rtc is how fault-tolerant and universal it is.

In the ideal world, all camera manufacturers would implement things following best practices or would release firmware updates fixing stuff.

But real world is very different, in real world you buy a camera and almost never receive a firmware update for it.

And it's amazing how go2rtc works great for the real world.

I don't even have any Tapo cameras, and maybe Tapo is one of the good manufacturers which fixes issues reported by their customers.

But it seems to me that @seydx's fix is generic enough to handle potential issues like this for any future camera that may send malformed responses.

From what I can tell, it doesn't seem that the fix will have a performance penalty.

Also, the code remains simple to me.

Maybe the request logic could be encapsulated in a new class to keep ONVIF client class tidy. And then maybe unit tests could be added to such class, especially to guarantee that it continues to work for previous cameras.

AlexxIT:
There are many fixes for broken protocols in go2rtc.

Just specifically in this case there is just too much code for a fairly useless situation - ONVIF from some fw from some Tapo cameras.

I've had to use a raw HTTP client elsewhere on go2rtc before. Perhaps after refactoring and merging this code - the fix can be returned.

It is even better to write a test on this fix, so that the crooked response from the camera is fixed there.

ostehamster:
I updated my setup from v1.9.7 to v1.9.8 the other day and faced the same issue - or at least I think it is the same.

The changes in commit f601c47, "Improve ONVIF server", removed the optional "Category" field from the "GetCapabilities" requests, which makes my Tapo CS320WS send garbage. Adding back the "Category" field fixes it.

I don't know if it is a better fix, but it is at least less code.

```
diff --git a/pkg/onvif/client.go b/pkg/onvif/client.go
index cb6221e..a54323e 100644
--- a/pkg/onvif/client.go
+++ b/pkg/onvif/client.go
@@ -161,6 +161,8 @@ func (c *Client) GetServiceCapabilities() ([]byte, error) {
 func (c *Client) DeviceRequest(operation string) ([]byte, error) {
        if operation == DeviceGetServices {
                operation = `<tds:GetServices><tds:IncludeCapability>true</tds:IncludeCapability></tds:GetServices>`
+        } else if operation == "GetCapabilities" {
+               operation = `<tds:GetCapabilities xmlns:tds="http://www.onvif.org/ver10/device/wsdl"><tds:Category>All</tds:Category></tds:GetCapabilities>`
        } else {
                operation = `<tds:` + operation + `/>`
        }
```

AlexxIT:
@ostehamster fixed https://github.com/AlexxIT/go2rtc/commit/57cd791348544802cd1135c6e0f3484ba1dc5ed7

AlexxIT:
Thanks!

AlexxIT:

eddively:
Do we know which Tapo cameras this may help?

AlexxIT:
This code works for all ONVIF cameras. It is not linked to Tapo.

PS. Using onvif source for Tapo cameras makes absolutely no sense.

PR #1620, opened 2025-02-27
Correct slight syntax error in example

DRuggeri:
As I chased down why go2rtc seems to go CPU crazy on new substreams, I was examining the sample commands and noticed this one seemed incorrect.


AlexxIT:
Thanks!

PR #1628, opened 2025-03-04
Added details on how to get Nest doorbell refresh token and device ID

dairefinn:
There wasn't a whole lot in the README about adding a nest doorbell directly so I added a guide on how to do it based on my recent experience doing it.

### DIFF
--- a/README.md
+++ b/README.md
@@ -635,7 +635,11 @@ streams:
 
 Currently only WebRTC cameras are supported.
 
-For simplicity, it is recommended to connect the Nest/WebRTC camera to the [Home Assistant](#source-hass). But if you can somehow get the below parameters - Nest/WebRTC source will work without Hass.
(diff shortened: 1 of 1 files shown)


slhad:
Thanks to you @dairefinn I was able to see that my doorbell and nest camera are the oldest ones and are sadly

```json
{
"traits": {
   "supportedProtocols": [
            "RTSP"
          ]
   }
}
```

and enabled me to find https://developers.google.com/nest/device-access/traits/device/camera-live-stream#generatertspstream

I will see if I can mash something together for the RTSP protocol because my doorbell is too expensive to replace

felipecrs:
@slhad in this case you can use my integration:

https://github.com/felipecrs/hass-expose-camera-stream-source

But I think you can upgrade your doorbell by enrolling it in preview updates in Google Home app. I never done so, just heard it works.

felipecrs:
Have you tried this btw?

slhad:
I will keep it in mind if I still get issues, but for now I have 10+ hours of uninterrupted doorbell footage, that's perfect

PR #1632, opened 2025-03-06
feat: x-www-form-urlencoded support

Klutrem:
Previously I was used [RtspToWeb](https://github.com/deepch/RTSPtoWeb)
This pull request implements `application/x-www-form-urlencoded` content-type, which is used by RTSPToWeb in webRTC protocol. With this PR, clients which worked with RTSPToWeb, will work with go2rtc now.


AlexxIT:
Thanks!

AlexxIT:

Issue #1639, opened 2025-03-10
Has anyone ported it to riscv64?


lab37:
```
go build
        package github.com/AlexxIT/go2rtc
        imports github.com/AlexxIT/go2rtc/internal/v4l2
        imports github.com/AlexxIT/go2rtc/pkg/v4l2/device: C source files not allowed when not using cgo or SWIG: 
        videodev2_arch.c
```

AlexxIT:
What's your hardware? Do you need V4L2 or I can just disable it?

lab37:
This hardware:  https://www.bit-brick.com/     
bit_brick_k1 board
os: bianbu
No need V4L2,  but  I have no idea about  how to disable it.
thank you very much.

nakata5321:
Hi, I want to do the same—create a port for RISC-V.

I have a [StarFive VisionFive 2 board](https://www.starfivetech.com/en/site/boards) running Ubuntu 24.04, which I installed from [this source](https://canonical-ubuntu-boards.readthedocs-hosted.com/en/latest/how-to/starfive-visionfive-2/).

I tried building the binary on my board by following the GitHub Actions workflow but encountered the same error:

> ```
> go build -ldflags "-s -w" -trimpath
>         package github.com/AlexxIT/go2rtc
>         imports github.com/AlexxIT/go2rtc/internal/v4l2
>         imports github.com/AlexxIT/go2rtc/pkg/v4l2/device: C source files not allowed when not using cgo or SWIG: 
>         videodev2_arch.c
> ```

To fix this, I set the environment variable:
```
export CGO_ENABLED=0
```
as done [here](https://github.com/AlexxIT/go2rtc/blob/master/.github/workflows/build.yml#L15).

However, I then ran into an issue with the V4L2 package.

From what I understand, in order to build this package, I need to provide a file named `videodev2_riscv64.go`, similar to how it's done for [other platforms](https://github.com/AlexxIT/go2rtc/tree/master/pkg/v4l2/device).

But I’m unsure where to find the necessary information to create this file.
@AlexxIT, could you help me understand where to get the details used in `videodev2_<PLATFORM>.go` files?

I’m not sure if I absolutely need the V4L2 package at this stage, but I’d like to compile a fully functional binary.

P.S. for now to build binary I copy all data from `videodev2_arm.go` to `videodev2_riscv64.go` and it seems fine.

AlexxIT:
Fixed in master version.

nakata5321:
[...]

PR #1651, opened 2025-03-13
add riscv64 platform for v4l2 package

nakata5321:
In continue of issue https://github.com/AlexxIT/go2rtc/issues/1639 add file for riscv64 architecture


vooon:
To make it buildable need to apply this patch: https://github.com/openwrt/packages/blob/ec99b0d71459c12b661f95f0b1d439bc3a814735/multimedia/go2rtc/patches/0005-v4l2-fix-riscv64-build-and-go-fmt.patch

And then enable RV64 in internal/v4l2 https://github.com/openwrt/packages/blob/ec99b0d71459c12b661f95f0b1d439bc3a814735/multimedia/go2rtc/patches/0004-v4l2-eanble-risc-v-64-support.patch

PR #1655, opened 2025-03-14
Add JPEG quality adjustment via quality parameter

streamthing:
This pull request introduces the ability to adjust JPEG quality in the FFmpeg-based processing pipeline by adding a `quality` parameter. The value of this parameter is directly passed to `-q:v` in FFmpeg, allowing users to control the output image compression level.

### 1. Modified `internal/ffmpeg/jpeg.go`
- Added a `quality` variable to parse query parameters.
- Implemented logic to append `-q:v {quality}` to the FFmpeg arguments if the parameter is provided.

### 2. Updated `README.md`
- Added documentation for the new `quality` parameter, explaining that it allows users to change the JPEG quality (e.g., `quality=1`).

## Usage Example
Users can now control the JPEG compression level via a URL query parameter:

```
http://<go2rtc_ip>:1984/api/frame.jpeg?src=camera&quality=1
```

**Where:**
- **quality=1** → Highest quality, larger file size.
- **quality=31** → Lowest quality, smaller file size.

## Why this change?
This update provides greater flexibility in adjusting image quality and file size based on user needs, making the API more versatile.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1203,6 +1203,7 @@ API examples:
   - You can use `width`/`w` and/or `height`/`h` params 
   - You can use `rotate` param with `90`, `180`, `270` or `-90` values
   - You can use `hardware`/`hw` param [read more](https://github.com/AlexxIT/go2rtc/wiki/Hardware-acceleration)
+  - You can use `quality` param to change jpeg quality (for example `1`)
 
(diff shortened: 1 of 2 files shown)


felipecrs:
This is awesome. Maybe in future we can support also `resolution` (or `width` and `height`) to avoid unnecessary data transfer while requesting a small icon, for example.

PR #1658, opened 2025-03-16
fix: Unpairing HomeKit streams

seydx:
During my tests with camera.ui, I encountered the following issue:

After I paired a HomeKit camera using go2rtc, and later deleted it through the API (without unpairing it), I was no longer able to unpair the camera. In HomeKit Discovery, the camera was shown as paired, without the possibility to unpair it.

Refactoring and code simplification (due to circular dependencies)

* Removed the `findHomeKitURLs` and `findHomeKitURL` functions from `internal/homekit/api.go` and `internal/homekit/homekit.go`, respectively, and replaced their usage with the new `streams.FindPrefixURLs` and `streams.FindPrefixURL` functions.
* Added `FindPrefixURL` and `FindPrefixURLs` functions to `internal/streams/helpers.go` to handle URL finding logic based on a prefix.

Dependency updates:

* Added the `hap` package to `internal/streams/api.go` for handling HomeKit unpairing.

Enhancements to stream deletion:

* Updated the `apiStreams` function in `internal/streams/api.go` to unpair HomeKit streams before deletion.

### DIFF
--- a/internal/streams/api.go
+++ b/internal/streams/api.go
@@ -5,6 +5,7 @@ import (
 
 	"github.com/AlexxIT/go2rtc/internal/api"
 	"github.com/AlexxIT/go2rtc/internal/app"
+	"github.com/AlexxIT/go2rtc/pkg/hap"
 	"github.com/AlexxIT/go2rtc/pkg/probe"
 )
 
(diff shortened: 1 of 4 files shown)


internal/streams/api.go
```diff
@@ -5,6 +5,7 @@ import (
 
 	"github.com/AlexxIT/go2rtc/internal/api"
 	"github.com/AlexxIT/go2rtc/internal/app"
+	"github.com/AlexxIT/go2rtc/pkg/hap"
```
AlexxIT, line 8: I don't like this dependency in this module. The streams module does not need to know about the existence of any particular source.

PR #1664, opened 2025-03-19
Onvif ptz

notjulian:
Hi Alex,
Thanks for this amazing project!

I'd like to have the PTZ controls...

So I made some changes. I understand that more features means more code to maintain and more bugs, so I tried to be as minimal as possible. Also, the layout maybe needs a UI/UX review.

Please let me know what you think. 
If it needs improvement, I'm happy to make changes.

PS: at the moment is only ptz, no zoom
PPS: to display the controls the src needs to have the _ptz in the name (happy to improve ...)

### DIFF
--- a/internal/api/ptz.go
+++ b/internal/api/ptz.go
@@ -1,0 +1,126 @@
+package api
+
+import (
+	"encoding/json"
+	"fmt"
+	"net/http"
+	"strings"
+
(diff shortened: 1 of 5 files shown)


AlexxIT:
Why? It can be useful someday.

notjulian:
because is exactly the same of the 
https://github.com/AlexxIT/go2rtc/pull/1665
sorry about the mess ...

PR #1682, opened 2025-03-31
fix(core): potential sender loop deadlock

infastin:
`buf` may become `nil` while Sender goroutine starts, which would cause goroutine to deadlock.
Pass `buf` directly to prevent this from happening.

### DIFF
--- a/pkg/core/track.go
+++ b/pkg/core/track.go
@@ -139,22 +139,29 @@ func (s *Sender) Start() {
 	}
 	s.done = make(chan struct{})
 
-	go func() {
-		// for range on nil chan is OK
-		for packet := range s.buf {
+	// pass buf directly so that it's impossible for buf to be nil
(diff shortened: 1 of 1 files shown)


AlexxIT:
I don't see the point of any of the fixes.
- range on nil - no problem
- reading done and buf without mutex - no problem

infastin:
range on nil blocks indefinitely, if you are fine with that, i will close the pull request.

AlexxIT:
Oh. You right. Did't know that.

infastin:
> * reading done and buf without mutex - no problem
Removed it.

AlexxIT:
Another similar problem - https://github.com/AlexxIT/go2rtc/commit/699a995e8c470a42f0127ecb045ab3fcdc120c33

AlexxIT:

PR #1712, opened 2025-04-24
Fix Stream API for DELETE

Bhlowe:
Quick fix for streams REST Delete Endpoint. Was using src instead of name for the streams map key. 
Also should use Delete method instead of calling delete(streams, name) for lock. 
Great project! Just starting to experiment.

### DIFF
--- a/internal/streams/api.go
+++ b/internal/streams/api.go
@@ -93,7 +93,7 @@ func apiStreams(w http.ResponseWriter, r *http.Request) {
 		}
 
 	case "DELETE":
-		delete(streams, name)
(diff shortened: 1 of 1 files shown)


Bhlowe:
Oops. Disregard this. I think calling Delete(src) instead of delete(streams, src) is a good idea. But I was just confused because sometimes "src" is a URL and name is the public stream ID.. and for delete, src is the stream ID. Sorry!

felipecrs:
Maybe you can send a PR for the `Delete(src)` part?

PR #1716, opened 2025-04-29
 Fix: Handle RTSP cameras requiring fragment-implied config in dynamic snapshot API

gudaja:
This PR fixes an issue where fetching snapshots using the /api/frame.jpeg endpoint with a dynamic src=URL parameter would fail for certain RTSP cameras.

This change ensures that dynamic snapshot requests work correctly for these cameras, allowing snapshots to be taken from 'unconfigured' cameras (via ?src=URL) that previously failed.


AlexxIT (approved):

AlexxIT:
Thanks!

AlexxIT:

PR #1744, opened 2025-05-20
Secrets Management

seydx:
# Secrets Management

## What's this about?
I've added a simple (but effective) secret management system to go2rtc. It lets you store credentials in a dedicated `secrets` section in the config file instead of embedding them directly in stream URLs. It works with existing template system and on top, every values used in secrets, will be auto redacted in log and stream info.

## Why is this useful?
- Keeps sensitive stuff like tokens and passwords organized in one place
- Clients can update credentials on the fly (super handy for tokens that expire)
- Everything stays in the same config file - just better organized
- Works alongside existing config approaches, so nothing breaks
- Redact secrets in log/stream info

#### 1. Reference a secret by name (client would manually query the secret, see example code)

```yaml
# go2rtc.yaml
secrets:
  ring_doorbell:
    camera_id: "12345678"
    device_id: "abc123def456"
    refresh_token: "eyJhbGciOi..."

streams:
  front_door:
    - ring:?secrets=ring_doorbell
```

#### 2. Template system for direct value insertion:
```yaml
# go2rtc.yaml
secrets:
  hikvision:
    username: admin
    password: passw0rd

streams:
  hikvision_4k:
    - rtsp://${hikvision_username}:${hikvision_password}@192.168.178.142/Streaming/channels/102
```

### In the code
```go
// Via query
query := u.Query()
secretsName := query.Get("secrets")

// or own secretsName
secretsName := "ring_doorbell"

var defaultConfig = struct {
  CameraID  string    `yaml:"camera_id"`
  DeviceID  string    `yaml:"device_id"`
  RefreshToken string `yaml:"refresh_token"`
} {
  CameraID:     query.Get("camera_id"),
  DeviceID:     query.Get("device_id"),
  RefreshToken: query.Get("refresh_token"),
}

// Get (or create) a secret with default values
secret, err := core.NewSecret(secretsName, defaultConfig)
if err != nil {
  return
}

// update the config with the secret values
secrets.Unmarshal(&config)

// Use it
cameraID := secret.Get("camera_id").(string)
refreshToken := secret.Get("refresh_token").(string)

// Update it when needed
secret.Set("refresh_token", newToken) 

// Writes back to go2rtc.yaml
secret.Save()
```

## Implementation details
- Thread-safe with mutex locks
- Simple key-value storage with type conversion helpers
- Works with nested values
- Minimal changes to existing code

## Note
I'd like to find a way to directly map stream names to secret names without explicitly specifying the secret in the URL, but haven't found a clean solution yet. I'm open to suggestions and ideas on how to best implement this or other improvements!

The code is pretty straightforward and doesn't change how anything else works, so existing stream configs will continue to work exactly as before.


AlexxIT:
I don't like the idea of new template format. Go2rtc already has support [environment variables](https://github.com/AlexxIT/go2rtc/blob/master/internal/app/README.md#environment-variables).

I don't mind another section in the config file to hold secrets. But it is important that these secrets are in the same format as the environment variables.

seydx:
> I don't like the idea of new template format. Go2rtc already has support [environment variables](https://github.com/AlexxIT/go2rtc/blob/master/internal/app/README.md#environment-variables).
> 
> I don't mind another section in the config file to hold secrets. But it is important that these secrets are in the same format as the environment variables.

Thanks for the feedback! I understand your concerns about both the template format and adding a new section to the config file.

To clarify, the main purpose of this PR is to provide developers with a way to programmatically update credentials at runtime, particularly for integrations like Ring, Tuya, or Nest that require periodic token refreshes.

Environment variables are great for initial configuration, but they're static after the application starts. The proposed system allows:

1. Dynamic updating of credentials during runtime (crucial for OAuth tokens that expire)
2. Writing changes back to the config file
3. A clean API for source integrations to manage their own credentials

Regarding the config structure, I'm flexible on the implementation approach. Instead of a dedicated `secrets` section, we could:

- Integrate this functionality directly into the existing `streams` section
- Use a naming convention within the existing config structure
- Store these dynamic credentials in a separate file entirely

The template syntax was mainly introduced as a convenience, but it's not the core value of the PR and could be removed.

What approach would you prefer that maintains the existing config structure while still providing a way for integrations to handle dynamic credential updates?

felipecrs:
Guys, if you ask me, I think it's a bad idea to store programmatic persistent data in go2rtc.yaml.

Instead, it should be stored in a separate file like `go2rtc.json` (a bit similar to how Home Assistant stores stuff in `.storage`, but it does not need to be that complicated).

Because go2rtc supports running with no configuration file at all.

You can also pass everything through config flags, for example.

Or in case of Frigate, the go2rtc.yaml is dynamically generated on every container startup out of the frigate.yaml, and then stored in memory at `/dev/shm/go2rtc.yml`.

This means changes made by go2rtc would not survive a container restart.

AlexxIT:
I don't mind adding a new section to the config file.

For example, `CAMERA_PASSWORD` can be taken either from `CREDENTIALS` file, from environment variable or from the `env` section of the config file. Does this solve your problem?

```yaml
env:
  CAMERA_PASSWORD: very-secret

streams:
  camera1: rtsp://rtsp:${CAMERA_PASSWORD}@192.168.1.123/av_stream/ch0
```

AlexxIT:
In the first version, go2rtc stored dynamic data in the `go2rtc.json` file.
It seemed like an added complexity to me. So I spent a lot of effort to allow dynamic data (like homekit authorization) to be stored in the `go2rtc.yaml` file.
Considering that go2rtc is able to work with several config files at the same time - it is already possible to configure one of the config files to store dynamic data and the other to store static data.

felipecrs:
> Considering that go2rtc is able to work with several config files at the same time - it is already possible to configure one of the config files to store dynamic data and the other to store static data.

Is there some kind of documentation for that? Or maybe you can share some example. I may be able to change Frigate to leverage this.

AlexxIT:
@felipecrs documentation above (go2rtc support multiple config files): https://github.com/AlexxIT/go2rtc/blob/master/internal/app/README.md

First file in the list will be selected as "dynamic":
https://github.com/AlexxIT/go2rtc/blob/ae8145f26696ffc19a13fea4a21b4a5306fc0e3a/internal/app/config.go#L65-L68

felipecrs:
> I don't mind adding a new section to the config file.
> 
> For example, `CAMERA_PASSWORD` can be taken either from `CREDENTIALS` file, from environment variable or from the `env` section of the config file. Does this solve your problem?
> 
> ```yaml
> env:
>   CAMERA_PASSWORD: very-secret
> 
> streams:
>   camera1: rtsp://rtsp:${CAMERA_PASSWORD}@192.168.1.123/av_stream/ch0
> ```

@seydx you can probably reconcile your approach with this suggestion. I.e. save the data back into `env:` rather than `secret:`.

seydx:
> > I don't mind adding a new section to the config file.
> > For example, `CAMERA_PASSWORD` can be taken either from `CREDENTIALS` file, from environment variable or from the `env` section of the config file. Does this solve your problem?
> > ```yaml
> > env:
> >   CAMERA_PASSWORD: very-secret
> > 
> > streams:
> >   camera1: rtsp://rtsp:${CAMERA_PASSWORD}@192.168.1.123/av_stream/ch0
> > ```
> 
> @seydx you can probably reconcile your approach with this suggestion. I.e. save the data back into `env:` rather than `secret:`.

Are you sure this is implemented? It does not work, i think we need to fix `ReplaceEnvVars` for this

```go
// ReplaceEnvVars - support format ${CAMERA_PASSWORD} and ${RTSP_USER:admin}
func ReplaceEnvVars(text string) string {
	var yamlConfig struct {
		Env map[string]string `yaml:"env"`
	}

	yaml.Unmarshal([]byte(text), &yamlConfig);

	re := regexp.MustCompile(`\${([^}{]+)}`)
	return re.ReplaceAllStringFunc(text, func(match string) string {
		key := match[2 : len(match)-1]

		var def string
		var dok bool

		if i := strings.IndexByte(key, ':'); i > 0 {
			key, def = key[:i], key[i+1:]
			dok = true
		}

		if dir, vok := os.LookupEnv("CREDENTIALS_DIRECTORY"); vok {
			value, err := os.ReadFile(filepath.Join(dir, key))
			if err == nil {
				return strings.TrimSpace(string(value))
			}
		}

		if value, vok := os.LookupEnv(key); vok {
			return value
		}

		if yamlConfig.Env != nil {
			if value, vok := yamlConfig.Env[key]; vok {
				return value
			}
		}

		if dok {
			return def
		}

		return match
	})
}
```

seydx:
> I don't mind adding a new section to the config file.
> 
> For example, `CAMERA_PASSWORD` can be taken either from `CREDENTIALS` file, from environment variable or from the `env` section of the config file. Does this solve your problem?
> 
> ```yaml
> env:
>   CAMERA_PASSWORD: very-secret
> 
> streams:
>   camera1: rtsp://rtsp:${CAMERA_PASSWORD}@192.168.1.123/av_stream/ch0
> ```


Are you sure this (checking config for `env`) is implemented? had to fix `ReplaceEnvVars` for this to work

```go
// ReplaceEnvVars - support format ${CAMERA_PASSWORD} and ${RTSP_USER:admin}
func ReplaceEnvVars(text string) string {
	var yamlConfig struct {
		Env map[string]string `yaml:"env"`
	}

	yaml.Unmarshal([]byte(text), &yamlConfig)

	re := regexp.MustCompile(`\${([^}{]+)}`)
	return re.ReplaceAllStringFunc(text, func(match string) string {
		key := match[2 : len(match)-1]

		var def string
		var dok bool

		if i := strings.IndexByte(key, ':'); i > 0 {
			key, def = key[:i], key[i+1:]
			dok = true
		}

		if dir, vok := os.LookupEnv("CREDENTIALS_DIRECTORY"); vok {
			value, err := os.ReadFile(filepath.Join(dir, key))
			if err == nil {
				return strings.TrimSpace(string(value))
			}
		}

		if value, vok := os.LookupEnv(key); vok {
			return value
		}

		if yamlConfig.Env != nil {
			if value, vok := yamlConfig.Env[key]; vok {
				return value
			}
		}

		if dok {
			return def
		}

		return match
	})
}
```

felipecrs:
> Are you sure this is implemented?

It is **not** implemented. It was a suggestion from @alexxit on how you can implement it in a way that he would accept. :)

felipecrs:
> but this means, we need to add a new section anyway? :) since there is no `env` section atm

Absolutely, that's what he meant. 😅

seydx:
> > but this means, we need to add a new section anyway? :) since there is no `env` section atm
> 
> Absolutely, that's what he meant. 😅


Oh man, i got it now.... I should read everything to the end before i reply >_>

> But it is important that these secrets are in the same format as the environment variables.

👍

seydx:
@AlexxIT 

Hey Alex, I updated the PR. 

I've now added both `env` and `secrets`.

I kept these two separate for a specific reason. The `env` config is a key=value dict and basically serves the same purpose as replacing `${...}` templates in the config.

The `secrets` config, however, is explicitly intended only for sensitive information, but it also replaces the `${...}` templates in the config.

Completely optional, users can put their sensitive information in the `secrets` config so that clients can make use of the new secrets function (update secrets on runtime etc)

The next reason though is that there's also a new `Redact` function that makes use of all available secrets and removes all areas from the log that contain a secret. Stream info now also redacts sensitive information.

What do you think about it?

Example:

```yaml
env:
  CAMERA_IP: 192.168.178.142
secrets:
  hikvision:
    username: admin2
    password: Admin!Admin
   
streams:
  hikvision:
    - rtsp://${hikvision_username}:${hikvision_password}@${CAMERA_IP}/Streaming/channels/101
```

Logs:

```
21:59:52.176 DBG [streams] start producer url=rtsp://*****:*****@192.168.178.142/Streaming/channels/101
22:00:04.194 DBG [streams] stop producer url=rtsp://*****:*****@192.168.178.142/Streaming/channels/101
```

Stream Info

```json
{
  "producers": [
    {
      "id": 3,
      "format_name": "rtsp",
      "protocol": "rtsp+tcp",
      "remote_addr": "192.168.178.142:554",
      "url": "rtsp://*****:*****@192.168.178.142/Streaming/channels/101",
      ....
    }
  ],
  "consumers": [
    ...
  ]
}

```

felipecrs:
Utilizing `secrets` to power the Redact function is very clever in my opinion, and a worthy feature.

Gasken23 (approved):

pkg/core/helpers.go
```diff
@@ -7,6 +7,8 @@ import (
 	"strconv"
 	"strings"
 	"time"
+
+	"github.com/AlexxIT/go2rtc/internal/app"
```
AlexxIT, line 11: Cannot import app into pkg

AlexxIT:
There are things that confuse me:
- There are a lot of places to set these variables: ENV from OS, CREDENTIALS_DIRECTORY, env from yaml, secrets from yaml. I suggest to remove "secrets from yaml" (especially two levels of nesting) and leave only "env from yaml".
- Removing variable values ​​from logs and URLs is a good idea. But this still won't save all users. Many will continue to enter links directly into configs. I already had a fix to solve this problem: https://github.com/AlexxIT/go2rtc/blob/ec08dfee9c0d4c0c6ae620ca1cc64c8d44be223d/pkg/core/helpers.go#L88-L91

There is a problem that some sources have no place to store dynamic data. I agree that it needs to be resolved. A very good example with pairings for homekit.

And here the question is, how will such sources get into the config for the first time? For example, ring. It should go into the sources section. And then immediately get to the env/secrets section. During its operation it will modify the env/secrets section.

I propose this version. I can implement it myself if it doesn't bother anyone. The env section can be used like regular environment variables, as well as dynamic key-value store. But it must be `string: string`.

```yaml
env:
  ring_doorbell: camera_id=12345678&device_id=abc123def456&refresh_token=eyJhbGciOi...
  hikvision_username: admin
  hikvision_password: passw0rd

streams:
  front_door:
    - ring:?env=ring_doorbell
  hikvision_4k:
    - rtsp://${hikvision_username}:${hikvision_password}@192.168.178.142/Streaming/channels/102
```

seydx:
@AlexxIT 

I see your point about too many places for credentials. Using env instead of a separate secrets section makes sense and keeps things simpler. The only reason I went with secrets was to have a clear boundary of what to redact

Regarding the initial setup and auto-save - my approach was to make it opt-in by the user. If the user sets up a secret/env entry and references it in the stream URL, sources can update it. If not, sources just use the values from the URL directly without any auto-save. The user decides whether they want the dynamic update functionality or not.

From your comment it sounds like you want sources to automatically create env entries and modify stream URLs without user intervention? Mine requires the user to set it up first.

AlexxIT:
I don't think the user has a choice. For the ring source, it is forced to use dynamic secrets.

If the source doesn't require dynamic authentication, there's little point in using variables. Perhaps just to hide credentials from logs. I don't think anyone will bother too much with this.

AlexxIT (approved):

AlexxIT:
Thanks! I hope I took all the features during the refactoring. What happened in the end:
- The `env` section has been added to the config. It works similarly to environment variables.
- Everything from `env` section, as well as all variables used in the config, will be encrypted when output to the API and logs.
- Integrations can read and add their own variables to the env section. This will be saved in the config.

I suggest not to provide named links from sources to env sections. Instead, the source can come up with a unique name for the env variable itself. For example: source name + camera ID + something else (`"ring_" + device_id + "_token"`).

```yaml
env:
  ring_ABCD_token: XXX
  dahua_pass: XXX

streams:
  camera1: ring:?device_id=ABCD&refresh_token=XXX
  rtsp-dahua1: rtsp://admin:${dahua_pass}@192.168.1.123/cam/realmonitor?channel=1&subtype=0
```

AlexxIT:

Issue #1060, opened 2024-04-23
Doorbird camera discussion


AlexxIT:
- https://github.com/AlexxIT/go2rtc/issues/323
- https://github.com/AlexxIT/go2rtc/issues/778
- https://github.com/AlexxIT/go2rtc/issues/841

oeiber:
Hi,
how can I help to integrate backchannel audio for doorbird devices?

AlexxIT:
Developing two way without physical access to device is very hard task. There is no feedback.
Theoretically only possible if there is some other camera nearby that I can watch and hear through.

oeiber:
I'll give you access to my device. Here's the link to the api documentation: 
[](https://www.doorbird.com/downloads/api_lan.pdf?rev=0.36)
You can find related information on page 18.

AlexxIT:
I also need yet another camera for feedback.

fergalom:
> I'll give you access to my device. Here's the link to the api documentation: [Doorbird API](https://www.doorbird.com/downloads/api_lan.pdf?rev=0.36) You can find related information on page 18.

@oeiber I can help also as requested

Gis70:
Hello, what do you need exactly, i own a doorbird too.

fransiouz:
I also have a doorbird if I can help. On my side both video and audio receive work well. But not microphone.

fransiouz:
There is a plugin in scrypted.app for doobird which seems to support audio transmit. Can it help ?
https://github.com/koush/scrypted/tree/main/plugins/doorbird

luke17287:
I also have a doorbird and would love this feature.

luke17287:
Dankeschön. An SIP hatte ich auch schon rumgewerkelt. Da ich aber ein Noob bin, habe ich es bisher nicht zum laufen gebracht;)
Ich teste den Guide und melde mich ✌️

luke17287:
Thank you - sound great.
I also tried a bit with SIP but im a noob and was not able to get it running;) .Ill try with your guide.

fransiouz:
Yes, same for me, but I succeed. Forum/discord + chatgpt help...
I can share you some config if needed

luke17287:
@fransiouz: I get crazy with asterisk. Iam not able to get my doorbird registered after several hours. I would appreciate if you could share your asterisk custom config files (extensions.conf / pjsip.conf / sip.conf)?

fransiouz:
[...]

PR #1783, opened 2025-06-24
Fix doorbird overload when backchannel audio is used

oeiber:
Added one connection per second limit as described in api documentation

### DIFF
--- a/pkg/doorbird/backchannel.go
+++ b/pkg/doorbird/backchannel.go
@@ -14,7 +14,10 @@ type Client struct {
 	conn net.Conn
 }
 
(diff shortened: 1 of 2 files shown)


oeiber:
#1060

pkg/doorbird/backchannel.go
```diff
@@ -15,7 +14,11 @@ type Client struct {
 	conn net.Conn
 }
 
+var dialLimiter = time.Tick(time.Second) // 1 tick per second
```
felipecrs, line 17: A comment with a link to the relevant api documentation would be nice.

oeiber (commented):
Link to api documentation added

PR #1794, opened 2025-07-03
Fix doorbird backchannel audio

oeiber:
- proper session handling
- honor http status codes
- prevent device from being flooded by limiting concurrent audio channels

### DIFF
--- a/pkg/doorbird/backchannel.go
+++ b/pkg/doorbird/backchannel.go
@@ -2,6 +2,7 @@ package doorbird
 
 import (
 	"bufio"
+	"errors"
 	"fmt"
 	"net"
 	"net/url"
@@ -64,11 +65,11 @@ func Dial(rawURL string) (*Client, error) {
 		if err == nil {
 			if statusCode == 204 {
 				conn.Close()
-				return nil, fmt.Errorf("DoorBird user has no api permission: %d", statusCode)
+				return nil, errors.New("DoorBird user has no api permission")
 			}
 			if statusCode == 503 {
 				conn.Close()
-				return nil, fmt.Errorf("DoorBird device is busy: %d", statusCode)
+				return nil, errors.New("DoorBird device is busy")
 			}
 		}
 	}
@@ -104,7 +105,7 @@ func (c *Client) GetTrack(media *core.Media, codec *core.Codec) (*core.Receiver,
 
 func (c *Client) AddTrack(media *core.Media, codec *core.Codec, track *core.Receiver) error {
 	if len(c.Senders) > 0 {
-		return fmt.Errorf("DoorBird backchannel already in use")
+		return errors.New("DoorBird backchannel already in use")
 	}
 
 	sender := core.NewSender(media, track.Codec)
--- a/pkg/doorbird/backchannel.go
+++ b/pkg/doorbird/backchannel.go
@@ -1,13 +1,20 @@
 package doorbird
 
 import (
+	"bufio"
 	"fmt"
 	"net"
 	"net/url"
+	"strconv"
+	"strings"
 	"time"
 
 	"github.com/AlexxIT/go2rtc/pkg/core"
 	"github.com/pion/rtp"
+)
+
+var (
+	clt Client
 )
 
 type Client struct {
@@ -16,6 +23,10 @@ type Client struct {
 }
 
 func Dial(rawURL string) (*Client, error) {
+	if clt.conn != nil {
+		return &clt, nil
+	}
+
 	u, err := url.Parse(rawURL)
 	if err != nil {
 		return nil, err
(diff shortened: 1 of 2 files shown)


felipecrs:
@AlexxIT [audio mixing](https://github.com/AlexxIT/go2rtc/pull/1794/commits/a92e04b6e0885bec723b222f5d16d1ffa35a22a1) would be nice for all 2-way audio sources. It would improve 2-way audio reliability and allow for more advanced use cases like keeping some ambient music playing in camera, while still being able to speak to camera. Or mixing TTS announcements, Assist, and what not.

What do you think? I can raise a separate issue for it.

AlexxIT:
Audio mixing will work only for PCM-like formats. Not very universal.
But you can create separate issue.

oeiber:
Is something still open here because it was not merged with version 1.9.10?

AlexxIT:
I released the update quickly. There was no time to review this PR.

AlexxIT:
A lot of work has been done here. But it's very much out of the go2rtc architecture.

- Using a singleton as a client. I understand why this is done. But other go2rtc sources don't work like that. It violates the overall logic.
- Logs inside pkg will create chaos in the overall log output.
- Audio mixing shouldn't be a single client's concern. It should be part of the entire architecture.

This code may work. And it works well. But when go2rtc needs to be improved, this part could become a major problem.

oeiber:
> A lot of work has been done here. But it's very much out of the go2rtc architecture.
> 
> * Using a singleton as a client. I understand why this is done. But other go2rtc sources don't work like that. It violates the overall logic.
> * Logs inside pkg will create chaos in the overall log output.
> * Audio mixing shouldn't be a single client's concern. It should be part of the entire architecture.
> 
> This code may work. And it works well. But when go2rtc needs to be improved, this part could become a major problem.

Do you have any constructive suggestions on how to tackle the problems to meet the 'overall logic'?
Otherwise, I'll delete the merge request and give up on the whole thing...

AlexxIT:
Does the device work with the current release code? If only one WebRTC client is open at a time.

oeiber:
No, it doesn't. Device is getting flooded with audio packets.

AlexxIT:
Does this happen even with a single two-way client? I see your code primarily fix the issue of multiple clients.

oeiber:
Yes, it happens also with a single client.
In your original version, each audio packet sent to doorbird starts a new tcp connection. After a while the device gets unreachable.
To fix the problem, I included a connection handler, which ensures, data sent to the device uses the same tcp connection. There's also a periodic connection check, which starts a new session, after the first one gets broken for any reason:

```
func (c *Client) Start() error {
	if c.conn == nil {
		return nil
	}

	go func() {
		ticker := time.NewTicker(SenderCleanupInterval)
		defer ticker.Stop()
		for range ticker.C {
			c.cleanupOrphanedSenders()
		}
	}()

	// Start a heartbeat goroutine to periodically check connection health
	go func() {
		heartbeat := time.NewTicker(HeartbeatInterval)
		defer heartbeat.Stop()

		for range heartbeat.C {
			c.mu.RLock()
			conn := c.conn
			c.mu.RUnlock()

			if conn != nil {
				// Try to write a small amount of silence to keep connection alive
				silence := make([]byte, 160) // 20ms of silence at 8kHz
				_ = conn.SetWriteDeadline(time.Now().Add(core.ConnDeadline))
				if _, err := conn.Write(silence); err != nil {
					fmt.Printf("DoorBird: Heartbeat write failed: %v\n", err)
					// Don't break here, let the main read loop handle it
				}
			}
		}
	}()

	// The main loop now just monitors for any unexpected data or connection errors
	// DoorBird typically doesn't send data back, so we use a very long timeout
	buf := make([]byte, 1)
	connectionStart := time.Now()
	for {
		_ = c.conn.SetReadDeadline(time.Now().Add(ConnectionReadTimeout))
		n, err := c.conn.Read(buf)
		if err != nil {
			elapsed := time.Since(connectionStart)
			fmt.Printf("DoorBird: Connection failed after %v, error: %v\n", elapsed, err)
			c.cleanup()
			cltMu.Lock()
			delete(cltMap, c.URL)
			cltMu.Unlock()
			return err
		}
		if n > 0 {
			fmt.Printf("DoorBird: Unexpected data received: %v\n", buf[:n])
		}
	}
}

```

AlexxIT:
This code may not work:
https://github.com/AlexxIT/go2rtc/blob/2b5f9429a8fec87ec1150b9379a563b9d9969e10/pkg/doorbird/backchannel.go#L90-L93

It is assumed that Start holds the connection. For the producer, it's simple. The Start receives the data.

In the case of backchannel, Start only needs to maintain the connection. Maybe send pings, as you do.

Maybe it needs to repeat this code:
https://github.com/AlexxIT/go2rtc/blob/2b5f9429a8fec87ec1150b9379a563b9d9969e10/pkg/isapi/client.go#L130-L132

oeiber:
I can give it a try. But there's still the question how to handle multiple baxkchannel sessions.

AlexxIT:
If there is only one line in the config, there will be only one session. Go2rtc simply won't let you run yet another. It's built into the core.

But if you open multiple browsers, they may compete for the right to send their data to the same session. This issue hasn't been debugged in go2rtc.

oeiber:
> This code may not work:
> 
> https://github.com/AlexxIT/go2rtc/blob/2b5f9429a8fec87ec1150b9379a563b9d9969e10/pkg/doorbird/backchannel.go#L90-L93
> 
> 
> 
> It is assumed that Start holds the connection. For the producer, it's simple. The Start receives the data.
> 
> 
> 
> In the case of backchannel, Start only needs to maintain the connection. Maybe send pings, as you do.
> 
> 
> 
> Maybe it needs to repeat this code:
> 
> https://github.com/AlexxIT/go2rtc/blob/2b5f9429a8fec87ec1150b9379a563b9d9969e10/pkg/isapi/client.go#L130-L132

Hi @AlexxIT, seems it is working with the few lines of code adopted into the Client function!

AlexxIT:
It would be great if you could debug your camera and make another PR. It's hard for me to do this because I don't have this camera.

Your code from this PR may be useful when it becomes possible to make a centralized mixer for multiple clients of the same backchannel.

oeiber:
Here we go :-) #1895

PR #1832, opened 2025-08-05
note on issues with tapo protocol

mihailstoynov:


AlexxIT:
Are you sure this makes sense? Does anyone else confirm this? Is there a discussion of this problem somewhere?

mihailstoynov:
> Are you sure this makes sense? Does anyone else confirm this? Is there a discussion of this problem somewhere?

This is my personal experience. I was struggling with setting the tapo protocol for a C210 camera and couldn't make it work (plaintext, md5, sha256). At some point I saw I had 2FA, switched it off, and voila - tapo protocol working - 2-way audo, microphone and the works (plaintext and sha256 worked fine).

All this time rtsp was working fine with a secondary account set up in the app.

While I'm here commenting, can i set the sub-stream (the small one) via the tapo protocol? I couldn't find any documentation on the subject.

AlexxIT:
You can use subtype query param

https://github.com/AlexxIT/go2rtc/blob/34b103bbcba512a07fb1ea8a3c219ae31d2201bf/pkg/tapo/client.go#L114-L120

kdunggttn:
> > Are you sure this makes sense? Does anyone else confirm this? Is there a discussion of this problem somewhere?
> 
> This is my personal experience. I was struggling with setting the tapo protocol for a C210 camera and couldn't make it work (plaintext, md5, sha256). At some point I saw I had 2FA, switched it off, and voila - tapo protocol working - 2-way audo, microphone and the works (plaintext and sha256 worked fine).
> 
> All this time rtsp was working fine with a secondary account set up in the app.
> 
> While I'm here commenting, can i set the sub-stream (the small one) via the tapo protocol? I couldn't find any documentation on the subject.

Thanks man you saved my day. I had the same issue until I read your comment.

cat101:
One of my tapo cameras started to fail. I found that on this model (C520WS), go2rtc was getting "unauthorized" when using tapo protocol. The solution is to go to the tapo app and enable "Me>Tapo lab>Third-party compatibility".

My other tapo cameras have not been affected

README.md
```diff
@@ -1,343 +1,1261 @@
...
+
+*[New in v1.2.0](https://github.com/AlexxIT/go2rtc/releases/tag/v1.2.0)*
+
+[TP-Link Tapo](https://www.tapo.com/) proprietary camera protocol with **two way audio** support.
+
+- stream quality is the same as [RTSP protocol](https://www.tapo.com/en/faq/34/)
+- use the **cloud password**, this is not the RTSP password! you do not need to add a login!
+- you can also use UPPERCASE MD5 hash from your cloud password with `admin` username
+- some new camera firmwares require SHA256 instead of MD5
+- Note: in some cases you might have to disable 2FA on your account.
```
salvacorts, line 536: I can confirm this with an extra step.
```suggestion
- Note: in some cases you might have to disable 2FA on your account and restart the camera.
```

README.md
```diff
@@ -1,343 +1,1261 @@
...
+- Note: with the proprietary `tapo://` protocol there is no way to access the smaller `/stream2`.
```
salvacorts, line 537: As @AlexxIT pointed out, you can use the `subtype` query param

Issue #1397, opened 2024-10-12
Nest source encounters 409 and retry attempts hit 429 rate-limit


trvrnrth:
I have a nest hello gen 1 wired doorbell source which, after some time experiences a 409 response:
```
error="streams: nest: wrong status: 409 Conflict, exec: rtsp://127.0.0.1:8554/front_door?audio: Invalid data found when processing input\n" stream=front_door
```

It appears that there is no back-off on attempts to re-connect so these are then rate-limited:
```
error="streams: nest: wrong status: 429 Too Many Requests, exec: rtsp://127.0.0.1:8554/front_door?audio: Invalid data found when processing input\n" stream=front_door
```

The full error message grabbed from the [Device Access Sample App](https://device-access-sample.web.app/) at this time is:
```
{
  "error": {
    "code": 429,
    "message": "Rate limited for the GenerateWebRtcStream API for the user.",
    "status": "RESOURCE_EXHAUSTED"
  }
}
```

At this point a restart is not sufficient to re-connect. go2rtc must be stopped for some time to allow the rate limit to "clear" before being started again (in my case this is a case of manually stopping, waiting a few minutes then starting the HA Frigate add-on).

I am unclear what causes the initial 409 conflict response, but implementing a back-off on at least the 429 response would be good. I guess it might make sense to just back-off on any error which would lessen the chance of hitting the rate limit in the first place.

Here is a full extract from the logs at the time of last failure:
```
2024-10-12 15:16:29.799077795  16:16:29.798 WRN github.com/AlexxIT/go2rtc/internal/streams/producer.go:171 > error=EOF url=ffmpeg:front_door#audio=aac
2024-10-12 15:16:50.871397445  16:16:50.871 WRN [rtsp] error="streams: nest: wrong status: 409 Conflict, exec: rtsp://127.0.0.1:8554/front_door?audio: Invalid data found when processing input\n" stream=front_door
2024-10-12 15:16:55.987067326  16:16:55.986 WRN [rtsp] error="streams: nest: wrong status: 409 Conflict, exec: rtsp://127.0.0.1:8554/front_door?audio: Invalid data found when processing input\n" stream=front_door
2024-10-12 15:17:01.104080094  16:17:01.103 WRN [rtsp] error="streams: nest: wrong status: 409 Conflict, exec: rtsp://127.0.0.1:8554/front_door?audio: Invalid data found when processing input\n" stream=front_door
[...]

PR #1841, opened 2025-08-19
Security Patch: Sanitize credentials on websocket error messages

hugoaboud:
I've found that `ffmpeg` can sometimes leak credentials through WebSocket error messages, as shown below:
image

### How to reproduce

1. Create a `ffmpeg` stream with a non-existing URL.
```
streams:
  TEST:
    - ffmpeg:rtsp://username:password@10.1.2.3:554/stream1
```

2. Open the WebRTC stream URL:
```
http://localhost:1984/stream.html?src=TEST&mode=webrtc
```

The connection timed out error is an example, but I came across it on a 500 error caused by network instability between go2rtc and the cameras.

### How it could be exploited

On a system which uses _go2rtc_ for providing web access to RTSP cameras, a bad actor could disrupt a camera's network, attempt to open a WebRTC stream through _go2rtc_ and then read the WebSocket messages to get the system credentials for said camera. 
This can also happen on partial network outages, where the _go2rtc_ server is temporarily unable to communicate with the camera, but still able to communicate with clients.

It's not unusual - although terrible practice - to have a common password for multiple cameras, which can escalate the threat.

### Proposed Solution

This PR uses a simple regex to censor URL credentials sent on websocket error responses:

image

> Disclaimer 1: I'm not a Go developer, please let me know of any stylistic or organization issues with the code
> Disclaimer 2: This might affect other places where the raw error is returned, I didn't look deep into it. I believe however that WS is the widest attack surface on this issue.

### DIFF
--- a/internal/api/ws/ws.go
+++ b/internal/api/ws/ws.go
@@ -8,10 +8,10 @@ import (
 	"strings"
 	"sync"
 	"time"
-	"regexp"
 
 	"github.com/AlexxIT/go2rtc/internal/api"
 	"github.com/AlexxIT/go2rtc/internal/app"
+	"github.com/AlexxIT/go2rtc/pkg/core"
 	"github.com/gorilla/websocket"
 	"github.com/rs/zerolog"
 )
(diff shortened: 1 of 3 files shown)


AlexxIT:
Thanks. I have made some changes.

AlexxIT:
#1397

AlexxIT:

PR #1858, opened 2025-09-03
Add tuna, ngrok analog

jidckii:
[Tuna](https://tuna.am/) - is an analogue of ngrok in Russia.

> Ngrok blocked users from Russia with [ERR_NGROK_9040](https://ngrok.com/docs/errors/err_ngrok_9040/) error code.


AlexxIT:
- tuna is NOT pre-installed
- I know about ngrok problem and remove it from docker, so it also NOT pre-installed (https://github.com/AlexxIT/go2rtc/commit/0e49a066baafeb7f7b610787c29908ec62179ee4)
- ngrok has limited the use of TCP on free accounts, so I don't see any point in continuing to support it
- tuna in free mode also has big limitations
- I don't mind supporting tools like tuna and ngrok if we can find a way to take them out of the application code. For example, in calling external scripts or script language like `expr`, that already supported

PR #1909, opened 2025-10-13
Supports RTSP redirection processing and optimizes status code judgme…

eddielth:
feat(rtsp): Supports RTSP redirection processing and optimizes status code judgment logic.

Refactored the status code handling logic in the `Do` method, using a switch statement instead of the original single if statement.
Added support for HTTP 301 and 302 redirect response processing, including Location header parsing, connection reestablishment, and request retries.
Also, the handling of the 401 Unauthorized status is retained, and the final error return logic has been adjusted.

### DIFF
--- a/pkg/rtsp/client.go
+++ b/pkg/rtsp/client.go
@@ -93,7 +93,46 @@ func (c *Conn) Do(req *tcp.Request) (*tcp.Response, error) {
 
 	c.Fire(res)
 
-	if res.StatusCode == http.StatusUnauthorized {
+	switch res.StatusCode {
+	case http.StatusOK:
+		return res, nil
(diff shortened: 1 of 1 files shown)


AlexxIT:
@felipecrs Thanks! I need to decline on this because the binary will be saved in the commit history.

felipecrs:
@AlexxIT if a commit to remove the binary had been made, you could have chosen the _Squash and Merge_ strategy for merging this PR, and then the binary would not be part of the repository's git history.

AlexxIT:
Yes. I know about squash. But anyway PR should be rewrited.

eddielth:
Sorry, I'm not very familiar with submitting pull requests, which led to accidentally including binary files.

AlexxIT:

PR #1910, opened 2025-10-13
Add compiled go2rtc to gitignore

felipecrs:
This is the standard output location when you compile go2rtc.

It should avoid issues like [this](https://github.com/AlexxIT/go2rtc/pull/1909#discussion_r2426085632).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -9,6 +9,8 @@ go2rtc_linux*
 go2rtc_mac*
 go2rtc_win*
 
+/go2rtc
+
 0_test.go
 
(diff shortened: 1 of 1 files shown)


AlexxIT:
And it allows you to understand the programmer’s qualifications :)
An experienced programmer always looks at the files they commit, rather than blindly doing `git add`

felipecrs:
That is true. Lol

AlexxIT:
Thanks!

AlexxIT:

PR #1934, opened 2025-11-12
Log unknown producers

edenhaus:
So we have a warning log entry if an unknown (or not activated) source is used. This helps if a user tries to use a source that is not enabled or does not exist.

### DIFF
--- a/internal/streams/handlers.go
+++ b/internal/streams/handlers.go
@@ -27,6 +27,10 @@ func HasProducer(url string) bool {
 		if _, ok := redirects[scheme]; ok {
 			return true
 		}
+
+		log.Warn().Str("scheme", scheme).Msg("[streams] Unknown producer scheme")
+	} else {
+		log.Warn().Str("url", url).Msg("[streams] Invalid producer URL")
 	}
 
 	return false


felipecrs:
> This helps if a user tries to use a source that is not enabled

Just to clarify, there's no such thing as _source that is not enabled_. (Right?)

- https://github.com/AlexxIT/go2rtc/issues/183

AlexxIT:
I don't think this is necessary because you'll get a 400 error on a PUT request:
https://github.com/AlexxIT/go2rtc/blob/e2c7d067308df08ea9b7b8d4c5c2296845d3ec12/internal/streams/api.go#L55-L58

edenhaus:
> I don't think this is necessary because you'll get a 400 error on a PUT request:

You don't know why exactly it failed. It could be that the source is invalid or something else... I assume 400 is returned if the params for the given source are incorrect. Status code 400 means bad request, but could be many things

AlexxIT:
I can modify the code so that it throws an text error. This makes more sense.

AlexxIT:
The following code has been added:
https://github.com/AlexxIT/go2rtc/commit/ac80f1470ef09093c3b59aae628ffd6f9849883e

PR #1935, opened 2025-11-12
Log unregistered paths as warning

edenhaus:


AlexxIT:
I don't think this is necessary. If we're talking about HA, python integration can detect broken links.

PR #1939, opened 2025-11-16
Add api endpoint to return supported schemas

edenhaus:
Home Assistant needs to know which schemas are supported by go2rtc. Currently this is hardcoded in https://github.com/home-assistant/core/blob/dev/homeassistant/components/go2rtc/__init__.py#L63-L91 but since go2rtc supports loading only specific modules, HA needs a way to get it.


AlexxIT:
Thanks!

AlexxIT:
https://github.com/AlexxIT/go2rtc/releases/tag/v1.9.12

PR #1956, opened 2025-11-25
Add `#timeout` param for ffmpeg source

felipecrs:
This PR adds the `#timeout` parameter to the FFmpeg source, similar to the `#timeout` parameter of the RTSP source.

It allows users to easily modify the timeout value for RTSP inputs per stream without the need of copying all the other parameters using `#input=`.

Also, it makes the global default timeout configurable in a similar manner. It still defaults to 5 seconds.

Example:

```yaml
ffmpeg:
  timeout: 10  # you can override it here for all streams

streams:
  # or you can override per-stream too
  mycam: ffmpeg:rtsp://192.168.1.33/0/av0#timeout=15#video=copy#audio=copy
```

### DIFF
--- a/internal/ffmpeg/ffmpeg.go
+++ b/internal/ffmpeg/ffmpeg.go
@@ -2,7 +2,6 @@ package ffmpeg
 
 import (
 	"net/url"
-	"strconv"
 	"strings"
 
 	"github.com/AlexxIT/go2rtc/internal/api"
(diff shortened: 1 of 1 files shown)


AlexxIT (approved):

AlexxIT:
Thanks!

AlexxIT:

PR #1959, opened 2025-11-28
Allow rtsp params (like #backchannel=0) from onvif sources

FIGIO55:
I added these lines since i needed to disable the backchannel on an onvif source and I did not find a way to do it, so I added this parameters re-append in the resulting stream uri.
I did this because i have issues with the video stream disconnecting when reproducing audio files to the camera if the source is just onvif.

This would allow sources like `onvif://user:password@ip:port/onvif/device_service#backchannel=0` to work with the backchannel disabled.


felipecrs:
Does it work?

FIGIO55:
It does, I tested it with my camera and it seems to work. My final setup looks like this
```
streams:
  camera1:
    - onvif://user:password@ip:8000/onvif/device_service#backchannel=0
    - onvif://user:password@ip:8000/onvif/device_service#backchannel=1#media=audio
```
so that I can use the video stream (first source), reproduce audio files (second source) and not have them "interfere" with each other.
This problem I think lies in the fact that, if using a single source, the strem freezes for half a second when starting the audio reproduction and it causes opencv to lose the stream. With this setup it seems to be working.

felipecrs:
Very cool!

Issue #1736, opened 2025-05-14
snapshots latency is unreliable / uncontrollable


mmerickel:
Configuring a simple RTSP stream such as the following creates a nice webrtc and mjpeg stream that appears in frigate.

However, when requesting a snapshot via `wget http://go2rtc-host:1984/api/frame.jpeg?src=camera1 -O frame.jpg` the latency is high and variable regularly taking between 1-2 seconds to return a response. I believe this is due to the lazy nature of go2rtc and the fact that it is waiting for a new keyframe from the raw RTSP camera stream after my request for a snapshot.

Is there anything that can be done within go2rtc to make the snapshot more snappy / reliable? My hope would be to configure something such that go2rtc is always listening to this feed and always storing a latest snapshot in memory, even by applying the diffs between keyframes such that it is returned as accurately as possible to the time I'm requesting the snapshot.

I can connect directly of course via `ffmpeg -i rtsp://go2rtc-host:8554/camera1 -update 1 foo.jpg` to get an up-to-date snapshot but this requires me basically setting up my own system on top of go2rtc just to get a live snapshot.

I also tried setting up a separate stream like camera1_mjpeg below as seemingly suggested in https://github.com/AlexxIT/go2rtc/issues/482#issuecomment-1601323057 but all it seems to do is add double latency (camera1_mjpeg waits for a keyframe from camera1 which waits for a keyframe from the source rtsp) when requesting a snapshot from it.

```yaml
streams:
  camera1:
  - rtsp://foo:bar@192.168.81.10:554/stream1
  camera1_mjpeg:
  - ffmpeg:camera1#video=mjpeg

api:
  origin: '*'
hass:
  config: /config
log:
  format: text
webrtc:
  candidates:
  - 192.168.20.5:8555
  - stun:8555
rtsp:
  default_query: mp4
ffmpeg:
  bin: /usr/lib/ffmpeg/7.0/bin/ffmpeg
```

mmerickel:
Ok based on your latest comment I setup the following and that is responding significantly more quickly - wget is down from 1-2 seconds to only around 200-400ms which is certainly an improvement.

```yaml
streams:
  camera1:
  - rtsp://foo:bar@192.168.81.10:554/stream1
  camera1_mjpeg:
  - rtsp://foo:bar@192.168.81.10:554/stream1#video=mjpeg
```

This is in comparison to the following other 2 options that don't work and I really don't understand the differences between them in go2rtc parlance.

```yaml
[...]

Issue #1657, opened 2025-03-14
Allow adding delay (`-ss`) to `/api/frame.jpeg` to avoid frames being dark


nadiamoe:
Heyo!

I've been a happy user of go2rtc for quite a long time, and I'd like to suggest a feature that I think could be useful to me and others.

I use go2rtc to expose a USB-connected webcam to my LAN. My config looks like this:
```yaml
streams:
  webcam:
    - "ffmpeg:device?video=/dev/video0&input_format=h264&video_size=1920x1080&rotate=180"
```

My webcam happens to support sourcing h264 video directly from the hardware, which is pretty neat as that means no transcoding is needed.

However, if I try to take a snapshot via `/api/frame.jpeg` while I'm **not** watching the stream, it looks like this:

![Image](https://github.com/user-attachments/assets/bd36b5d9-16d7-4237-9b90-5ef12e651149)

If I **am** watching the stream, it looks like this:

![Image](https://github.com/user-attachments/assets/e8979b18-8dd9-45f8-bb86-9ad4f6499dd0)

What I believe is happening here is that the webcam needs a bit of time to adapt to the lighting conditions. If the webcam is already initialized, the snapshot comes out good. If it needs to be initialized on the spot, it looks very dark.

I have validated my theory by changing my config so it looks like this:
```yaml
streams:
  webcam:
    - "ffmpeg:device?video=/dev/video0&input_format=h264&video_size=1920x1080&rotate=180#raw=-ss 00:00:00.5"
```

This adds half a second delay to all video processing, which is very much not great, but makes `/api/frame.jpeg` output nice snapshots all the time.

I'd love if this delay could only be used in the jpeg endpoint:
https://github.com/AlexxIT/go2rtc/blob/47f32a5f5597e956b51dc5d0f0beadad6918ebf2/internal/ffmpeg/jpeg.go#L37

Either set via config, or from a get param (that could allow turning the 1s delay on/off).

streamthing:
Hi,
I can't reproduce your problem, but could you edit file `internal/ffmpeg/jpeg.go` with this code and test this with `&ss=00:00:00.5`?

```
func parseQuery(query url.Values) *ffmpeg.Args {
	args := defaultArgs()

	var width = -1
	var height = -1
	var r, hw, ss string

	for k, v := range query {
		switch k {
		case "width", "w":
			width = core.Atoi(v[0])
		case "height", "h":
			height = core.Atoi(v[0])
		case "rotate":
[...]

PR #1967, opened 2025-12-03
Add in-memory snapshot cache for /api/frame.jpeg endpoint

adamjacobmuller:
## Summary

This PR adds an intelligent in-memory snapshot caching system for the `/api/frame.jpeg` endpoint to eliminate slow response times caused by large keyframe intervals.

## Problem

The current `/api/frame.jpeg` implementation blocks until the next keyframe arrives. When cameras use large keyframe intervals (5-10+ seconds), snapshot requests can take that long to respond:

- **Typical response time:** 150ms - 10,000ms (depends on when next keyframe arrives)
- **Cached response time:** <10ms (99% improvement)

This creates a poor user experience in:
- Home Assistant dashboards (loading spinners, timeouts)
- Notification thumbnails (delayed or missing images)
- Mobile apps (perceived as "broken" or "slow")

**Additional benefit:** Reduces camera load since multiple clients can share the same cached snapshot instead of each triggering separate connections.

## Solution

Implements background snapshot caching with:
- **Background keyframe consumer** that continuously captures frames
- **Always-ready snapshots** - no waiting for next keyframe
- **Configurable idle timeout** (default: 10 minutes) - stops when not needed
- **Stale detection** - auto-restarts if cache gets too old
- **Opt-in by design** - clients must request `?cached=true` (or configure globally)
- **Graceful fallback** - serves fresh snapshot if cache unavailable

## Configuration

```yaml
mjpeg:
  snapshot_cache: true                        # Enable/disable (default: true)
  snapshot_cache_timeout: 600                 # Idle timeout in seconds (default: 600)
  snapshot_serve_cached_by_default: false     # Default behavior (default: false)
```

## Usage

```bash
# Request cached snapshot - returns immediately (<10ms)
curl http://localhost:1984/api/frame.jpeg?src=camera1&cached=true

# Request fresh snapshot - waits for next keyframe (traditional behavior)
curl http://localhost:1984/api/frame.jpeg?src=camera1&cached=false

# Check cache age via response headers
# X-Snapshot-Age-Ms: 1234
# X-Snapshot-Timestamp: 2025-12-03T12:34:56.789Z
# X-Snapshot-Cached: true/false
```

## Performance Impact

**Before (fresh snapshot):**
- First request: 150-2700ms (average keyframe interval)
- Subsequent requests: 150-2700ms (each waits for keyframe)
- **10-second keyframe interval = 10-second snapshot delay**

**After (cached snapshot):**
- First request: <10ms (if cache exists)
- Subsequent requests: <10ms (served from memory)
- **Eliminates keyframe wait entirely**

**Memory overhead:** ~300KB JPEG per cached stream (only while accessed)

## Benefits

- **Instant snapshots** - no keyframe waiting
- **Better UX** - dashboards/apps feel responsive
- **Reduced camera load** - single persistent connection shared by all clients
- **Production-ready** - extensive logging at TRACE level for debugging

## Implementation Details

- New file: `internal/streams/snapshot_cache.go` (223 lines)
- Modified: `internal/mjpeg/init.go` - add caching logic to frame handler
- Modified: `internal/streams/stream.go` - add cache storage fields
- Includes fix: `pkg/h265/rtp.go` - prevent panic from stale buffer pointers

## Testing

Tested with:
- Cameras with 1s, 5s, and 10s keyframe intervals
- Multiple concurrent clients requesting cached snapshots
- Idle timeout triggering and cache restart
- Stale cache detection and recovery
- H264, H265, and JPEG source codecs

## Related Issues

Directly addresses:
- #1736 - Snapshot latency concerns (snapshots now <10ms)

May help with:
- #1657 - Dark frame issues (cache can skip initial frames)

---

This change is backward compatible - default behavior is unchanged unless clients opt-in with `?cached=true`.

### DIFF
--- a/pkg/h265/rtp.go
+++ b/pkg/h265/rtp.go
@@ -106,7 +106,9 @@ func RTPDepay(codec *core.Codec, handler core.HandlerFunc) core.HandlerFunc {
 
 		clone := *packet
 		clone.Version = h264.RTPPacketVersionAVC
-		clone.Payload = buf
+		// Clone buffer to prevent data corruption - buf will be reused for next frame
+		// and would overwrite clone.Payload's underlying array if they share it
+		clone.Payload = append([]byte(nil), buf...)
 
 		buf = buf[:0]
 		nuStart = 0 // Reset nuStart when buffer is cleared


felipecrs:
Interesting. I wonder how much this overlaps with [GOP cache](https://github.com/AlexxIT/go2rtc/pull/1887).

I mean, if GOP cache was implemented, I suppose it could be reused for the snapshot too.

skrashevich:
I'll just leave it here: #1155

adamjacobmuller:
So you basically did this same thing a year ago or is yours more limited to just MJPEG?

AlexxIT:
Sorry. A lot of code with poor code design.
Adding snapshot logic to the streams package is completely wrong. Moreover, make it dependent on the magic package.
This will definitely not be accepted.

@skrashevich has a short, simple implementation. But there is no control over the timeout from the user side - this is bad.

PR #1977, opened 2025-12-10
Add get request to preload endpoint for listing them

edenhaus:
To add support for preloading in Home Assistant, HA needs a way to get the current preloads. So it can identify if it needs to update them or not. This PR extend the preload endpoint for it


AlexxIT (approved):

AlexxIT:
Thanks! I made some changes. Now preload is an object, not just a string. By analogy with the stream.

PR #1979, opened 2025-12-13
Fix memory leaks in HLS sessions, WriteBuffer, and RTSP UDP handling

matteius:
- Add MaxSessions limit (100) to prevent HLS session accumulation
- Add session cleanup goroutine with 30-second monitoring interval
- Add MaxBufferSize (16MB) limit to HLS session buffers
- Add MaxWriteBufferSize (8MB) limit to WriteBuffer
- Add sync.Pool for UDP read buffers in RTSP conn to reduce allocations
- Add debug logging for session creation/expiration

These fixes address memory growth issues observed during extended operation where go2rtc memory would grow from ~150MB to 7GB+ over 2.5 days with usage from lightNVR.

### DIFF
--- a/internal/hls/hls.go
+++ b/internal/hls/hls.go
@@ -51,12 +51,39 @@ func sessionCleanup() {
 	defer ticker.Stop()
 
 	for range ticker.C {
-		sessionsMu.Lock()
+		now := time.Now()
+		var staleIDs []string
+
+		sessionsMu.RLock()
 		count := len(sessions)
-		sessionsMu.Unlock()
+		for id, session := range sessions {
+			// Check if session has exceeded maximum age
+			if now.Sub(session.createdAt) > MaxSessionAge {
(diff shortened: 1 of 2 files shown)


AlexxIT:
> * Add sync.Pool for UDP read buffers in RTSP conn to reduce allocations

Using this doesn't make any sense at all. Because you're allocating new memory for every new packet anyway.
You can simply move the creation of the read buffer outside the loop.

PS. It's better to organize these things into different PR. Because they're completely different.

AlexxIT:
> * Add MaxWriteBufferSize (8MB) limit to WriteBuffer

It's quite rare that `WriteBuffer` will write to `*bytes.Buffer`. I can't say right away where this happens.

And truncating the data is a bad solution. For many formats, it will completely break streaming. And sending the data further makes no sense.

matteius:
Thanks for your feedback, I will try a different approach.

PR #1980, opened 2025-12-14
Replace ace editor to Monaco with config hints

skrashevich:
obsolete #1121

Requires testing in different browsers. tested in Safari

image


AlexxIT:
- No second level key hints
- No validation for duplicate keys (at the same level)
- No validation for key type (number instead of string)
- Autocompletion at the second level sometimes works correctly, sometimes completely incorrectly
- It's better to download `schema.json` from the internet (will be posted on go2rtc.org in the future)

skrashevich:
> * It's better to download `schema.json` from the internet (will be posted on go2rtc.org in the future)

Why better?

felipecrs:
Maybe to decrease the binary size, but the schema is probably very small, specially if compressed.

skrashevich:
> Maybe to decrease the binary size, but the schema is probably very small, specially if compressed.

~20 Kb uncompressed. Dismissively small

AlexxIT:
Thanks! Awesome changes.

AlexxIT:

PR #1991, opened 2025-12-19
Adding onvif name and hardware in info field of onvif discovery

FIGIO55:
I like the onvif discover functinality in the add section, but I wanted some device information.
So I added code to extract the `onvif://www.onvif.org/name` and `onvif://www.onvif.org/hardware` fields from the onvif envelope and used it to populate the `Info` field already present in the `api.Source` object.
To better present this change, I created the `DiscoveryDevice` type in `pkg/onvif/helpers.go` with the fields being:
```go
type DiscoveryDevice struct {
	URL      string
	Name     string
	Hardware string
}
``` 
For clarity purposes, I also renamed the function `DiscoveryStreamingURLs` to `DiscoveryStreamingDevices` from the same package.

### DIFF
--- a/internal/onvif/onvif.go
+++ b/internal/onvif/onvif.go
@@ -7,7 +7,6 @@ import (
 	"net/url"
 	"os"
 	"strconv"
-	"strings"
 	"time"
 
 	"github.com/AlexxIT/go2rtc/internal/api"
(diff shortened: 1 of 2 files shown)


AlexxIT:
Thanks! I have made some changes. I removed duplicate checking and whitespace trimming to improve code readability.
And I added queryunescape because there was encoded space in my camera name.

FIGIO55:
Ah sorry about that, none of my cameras had that space in the name.
Anyway, thank you, glad to help!

AlexxIT:

PR #1992, opened 2025-12-19
Allow rtsp params (like #backchannel=0) from onvif sources

FIGIO55:
I added these lines since i needed to disable the backchannel on an onvif source and I did not find a way to do it, so I added this parameters re-append in the resulting stream uri.
I did this because i have issues with the video stream disconnecting when reproducing audio files to the camera if the source is just onvif.

This would allow sources like onvif://user:password@ip:port/onvif/device_service#backchannel=0 to work with the backchannel disabled.
An example from my usage is:
```
streams:
  camera1:
    - onvif://user:password@ip:8000/onvif/device_service#backchannel=0
    - onvif://user:password@ip:8000/onvif/device_service#backchannel=1#media=audio
```

This is a re-creation of PR #1959 because I messed up the branches on my fork

### DIFF
--- a/internal/onvif/onvif.go
+++ b/internal/onvif/onvif.go
@@ -34,9 +34,6 @@ func Init() {
 var log zerolog.Logger
 
 func streamOnvif(rawURL string) (core.Producer, error) {
-	// Split the ONVIF URL to extract hash-based arguments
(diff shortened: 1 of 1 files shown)


AlexxIT:
Thanks!

AlexxIT:

PR #2010, opened 2025-12-31
Add EXPOSE to Dockerfile

kaysond:
FYI this doesn't actually forward the ports when running. That still has to be done manually in docker compose or docker run. It's just an indication that these are the ports used. Other services, like traefik, can use this information to automatically reverse proxy requests to the correct port.

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -47,8 +47,7 @@ RUN if [ "${TARGETARCH}" = "amd64" ]; then apk add --no-cache libva-intel-driver
 
 COPY --from=build /build/go2rtc /usr/local/bin/
 
-EXPOSE 1984 8554 8555
-EXPOSE 8555/udp
+EXPOSE 1984 8554 8555 8555/udp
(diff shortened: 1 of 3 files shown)


skrashevich:
I have doubts about the need for this change.

But if you consider it necessary, why did you ignore hardware.Dockerfile and port 8555?

kaysond:
> I have doubts about the need for this change.

It's a documented best practice: https://docs.docker.com/build/building/best-practices/#expose
And as mentioned, other services/containers like traefik can and do use it to know what ports the container listens on

> But if you consider it necessary, why did you ignore hardware.Dockerfile and port 8555?

Added. Just an oversight / misunderstanding on my part.

AlexxIT:
Looks like official nginx image also uses EXPOSE. Although this port may be incorrect under custom application configurations.

https://github.com/nginx/docker-nginx/blob/master/stable/alpine-slim/Dockerfile

AlexxIT:
Thanks!

AlexxIT:

PR #2017, opened 2026-01-03
Fix ghost producer

seydx:
When a WebRTC client disconnects while a producer is still dialing, the producer keeps running forever with no consumers attached

## Fix

Added `IsClosed()` to the Consumer interface. After `Dial()` completes, we now check if the consumer is still alive before adding tracks. If not, we bail out early so the producer can stop properly.

WebRTC implements this by tracking when the WebSocket transport closes during connection setup.

**Note:**
`IsClosed()` on the interface allows any consumer to implement proper closed-state detection. Right now only WebRTC needs it, but other consumers (RTSP, HomeKit, etc.) could implement it later if needed. The default in `core.Connection` returns `false`, so existing consumers work unchanged.

### DIFF
--- a/pkg/hass/client.go
+++ b/pkg/hass/client.go
@@ -113,6 +113,10 @@ func (c *Client) Stop() error {
 	return c.conn.Stop()
 }
 
+func (c *Client) IsClosed() bool {
+	return c.conn.IsClosed()
+}
+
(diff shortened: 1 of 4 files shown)


AlexxIT:
I wouldn't make any hasty changes to this logic. It's better to carefully rewrite the core from start to finish.

PR #2038, opened 2026-01-14
Add optional resolution for incoming browser stream

cryptosebek:
Adds optional resolution constraint for browser camera via URL parameter.

Usage:
- `?resolution=1280x720` - request specific resolution
- `?resolution=1920x1080@30` - request resolution with framerate

This allows users to control camera quality when using browser as an incoming stream source.


AlexxIT:
The idea is good, but I'm not sure about the parameter names. They should be the same as in the rest of go2rtc.
For example, the parameters `width` and `height` are now more common.

PR #2051, opened 2026-01-21
StripUserinfo from log entries and errors

edenhaus:
Improve stripping of user information in errors and log entries.
It was reported on https://github.com/home-assistant/core/issues/139902

Used claude.ai to find all spots

### DIFF
--- a/internal/echo/echo.go
+++ b/internal/echo/echo.go
@@ -8,6 +8,7 @@ import (
 
 	"github.com/AlexxIT/go2rtc/internal/app"
 	"github.com/AlexxIT/go2rtc/internal/streams"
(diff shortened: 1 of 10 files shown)


AlexxIT:
Thanks. I did the same in a better place.

AlexxIT:
It looks like we haven't described go2rtc's new feature anywhere.
You can specify the env section in the config and use it anywhere, like environment variables.
The values ​​of these variables will be stripped from all logs and web output as secrets.
This way, user can clear sensitive data anywhere in the link, for example, in the link's query parameters.
https://github.com/AlexxIT/go2rtc/pull/1744#issuecomment-3376677816

Additionally, the next version will strip out any userinfo from typical links.

PR #2064, opened 2026-01-31
webrtc: Fix ice_servers parsing for a list of URLs

melanconj:
Unmarshal gives a []any instead of a []string like the code expected. Fixed the cast & handling for the list plus tests.


AlexxIT:
Thanks!

Issue #2052, opened 2026-01-21
Option to make configuration read-only


michaelosthege:
Hello and thanks a lot for this great project! We love the simplicity in comparison to other solutions!

We would like to lock-down the configuration for added security and to prevent accidental destructive actions.

Just locking down with username/password would be insufficient because we'd like to give colleagues access to the dashboard and streams while denying them the ability to change configuration. (We follow a config-as-code policy.)

While keeping basic "viewing" functionality we would like to disable "add", "config" and "delete" actions from the main dashboard and the "Play audio" and "Publish stream" actions from the links page.

I have marked it in these screenshots:

<img width="639" height="328" alt="Image" src="https://github.com/user-attachments/assets/11e4d432-c506-4a32-8ba8-d1c1b60c9ee9" />

<img width="557" height="1334" alt="Image" src="https://github.com/user-attachments/assets/992481fb-40d5-4b1f-a781-8d60ea081ca4" />

I did not find corresponding options in the README/Wiki.

Is this already possible?

How difficult would it be to implement?

AlexxIT:
There is no such function. It needs to be improved.

Right now, you can configure nginx or something similar to allow only certain URLs with certain queries and methods.

felipecrs:
Why not _Play audio_?

PR #2066, opened 2026-02-01
Add read-only mode to API and UI

skrashevich:
Implement a read-only mode that disables write actions across the API and UI, enhancing user experience with confirmation prompts and server polling on save. This change ensures that users are aware of the read-only status and prevents unintended modifications.

Implements #2052


AlexxIT:
It might be a good idea, but I don't like the implementation details. Every new line needs to be carefully examined.

PR #2089, opened 2026-02-11
api: add GET /api/ready readiness endpoint

### LINKED ISSUE #2088: GET /api/schemes returns incomplete list during startup (race condition)
## Problem

`GET /api/schemes` returns an incomplete list of supported schemes when called shortly after go2rtc starts. The root cause is a race between the HTTP server binding its socket and the remaining modules registering their schemes.

The startup sequence in `main.go` is:

```
1. api.Init()     ← HTTP server binds and starts listening HERE
                    logs: "INF [api] listen addr=..."
2. ws.Init()
3. streams.Init() ← /api/schemes endpoint registered here
4. http.Init()    ← "http"/"https" schemes registered
5. rtsp.Init()    ← "rtsp"/"rtsps" schemes registered
6. webrtc.Init()  ← "webrtc" scheme registered
7. ffmpeg.Init()  ← "ffmpeg" scheme registered
   ...and so on
```

Any client that queries `/api/schemes` between step 1 and the final module `Init()` will receive an incomplete — or in early steps, empty — scheme list.

## Impact: Home Assistant integration

Home Assistant's go2rtc integration manages the go2rtc process and watches its stdout for the message `"INF [api] listen addr="` to know when the server is ready ([`server.py:21`](https://github.com/home-assistant/core/blob/dev/homeassistant/components/go2rtc/server.py#L21), [`server.py:246-247`](https://github.com/home-assistant/core/blob/dev/homeassistant/components/go2rtc/server.py#L246-L247)).

This message is emitted in the `listen()` function — the moment the TCP socket binds, which happens inside `api.Init()` — the **first** module to be initialized. At this point, none of the stream-source modules have run yet.

Immediately after this message is seen, HA calls `GET /api/schemes` to build a cache of supported schemes. Because this call races with the remaining module initializations, the cache is frequently missing schemes like `ffmpeg`, `webrtc`, `onvif`, `exec`, and `rtsp`, causing valid stream sources to be incorrectly rejected.

## Proposed fix

Add a `GET /api/ready` endpoint that returns `503 Service Unavailable` while modules are still initializing, and `200 OK` once all modules have completed their `Init()` calls. This would require:

1. A readiness flag in the `api` package, set via an exported `SetReady()` function
2. The `readyHandler` registered at `api/ready` in `Init()`
3. A call to `api.SetReady()` in `main.go` after the module init loop completes

This gives clients a reliable readiness probe rather than relying on a log message that fires too early. It also doubles as a useful Docker/Kubernetes health check endpoint.

Happy to submit a PR.

twhittock:
Closes #2088

## Summary

The HTTP server starts listening inside `api.Init()`, which is the first module initialized. All other modules register their schemes afterward. Any client querying `/api/schemes` between `api.Init()` and the completion of the module init loop receives an incomplete list.

This PR adds a `GET /api/ready` endpoint that returns:
- `503 Service Unavailable` (body: `starting`) while modules are still initializing
- `200 OK` once all modules have completed their `Init()` calls

`SetReady()` is called from `main.go` after the module init loop completes.

## Changes

- `internal/api/api.go`: add `ready atomic.Bool`, exported `SetReady()`, `readyHandler`, and register `api/ready`
- `main.go`: call `api.SetReady()` after the module init loop

### DIFF
--- a/internal/api/api.go
+++ b/internal/api/api.go
@@ -11,12 +11,30 @@ import (
 	"strconv"
 	"strings"
 	"sync"
+	"sync/atomic"
 	"syscall"
 	"time"
 
 	"github.com/AlexxIT/go2rtc/internal/app"
 	"github.com/rs/zerolog"
 )
+
+// ready is set to true after all modules have finished initializing.
+var ready atomic.Bool
+
+// SetReady marks the server as fully initialized. Call this once all modules
+// have registered their handlers and schemes.
+func SetReady() {
+	ready.Store(true)
+}
+
+func readyHandler(w http.ResponseWriter, r *http.Request) {
+	if ready.Load() {
+		w.WriteHeader(http.StatusOK)
+	} else {
+		http.Error(w, "starting", http.StatusServiceUnavailable)
+	}
+}
 
 func Init() {
 	var cfg struct {
@@ -58,6 +76,7 @@ func Init() {
 	HandleFunc("api/exit", exitHandler)
 	HandleFunc("api/restart", restartHandler)
 	HandleFunc("api/log", logHandler)
+	HandleFunc("api/ready", readyHandler)
 
 	Handler = http.DefaultServeMux // 4th
 
--- a/main.go
+++ b/main.go
@@ -121,5 +121,8 @@ func main() {
 		}
 	}
 
+	// Signal that all modules have finished initializing and all schemes are registered.
+	api.SetReady()
+
 	shell.RunUntilSignal()
 }


felipecrs:
Useful for https://github.com/AlexxIT/go2rtc/pull/707, and for https://github.com/blakeblackshear/frigate/blob/6accc382756f39d36f547e631b337dbb71728282/docker/main/rootfs/etc/s6-overlay/s6-rc.d/go2rtc-healthcheck/run#L15.

AlexxIT:
There's no point in creating new API endpoints for this task. It's a bad solution.

twhittock:
OK so what does a good solution look like? It's a genuine problem I had with a home assistant integration I was writing, which was extremely difficult to track down. You think parsing strings in a log file and hoping localisation or a random logging change won't break anything is a good solution?

AlexxIT:
If /api/schemes returns irrelevant information, it makes sense to delay this output.
Or delay the response /api. Or even delay any APIs.
But I haven't studied this problem yet. I'll take a look.

twhittock:
@AlexxIT  Thank you, so if I understand correctly, you suggest I reorder the startup sequence? I considered it but don't know enough about the project to be sure it's a safe change - if I open another PR with the API subsystem initialised last, would that work for you?

AlexxIT:
If I understand correctly, the API only returns partial data. I propose that the API not return anything at all until the start is complete. The request may "hang" until the start is complete.

twhittock:
So you suggest I patch the API to block all HTTP responses until the ready flag I introduce here is set, and remove the ready endpoint?

AlexxIT:
I'm not sure about everyone. It might only make sense for /api/schemes.

felipecrs:
@AlexxIT I completely agree that for the specific race condition described, having the API block/hang until initialization is complete is the correct fix for clients.

However, I think there is still a strong case for a dedicated `/api/ready` or `/api/health` endpoint, independent of that fix.

It is a standard pattern for containerized applications to require a lightweight, specific endpoint to verify the service is up. E.g. Kubernetes probes, Docker `HEALTHCHECK`, etc.

This information is used to know if the service/container is alive or needs restarting, or even when it's safe to route traffic to it in case there are additional container replicas.

Having a dedicated API would allow tools like Frigate or even Home Assistant to check the health of go2rtc without making a heavier call to `/api/streams` (or any other arbitrary endpoint).

Would you be willing to accept a, perhaps more robust, implementation of a health check endpoint?

AlexxIT:
`/api` is best option for healthcheck.

PR #2092, opened 2026-02-13
260213 fix schemas race condition

skrashevich:
Proposed fix of #2088

Benefits:

- A universal patch that will allow to fix all similar errors by adding one line of code
- Minimal changes in existing code

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -21,3 +21,4 @@ website/.vitepress/dist
 
 node_modules
 package-lock.json
+CLAUDE.md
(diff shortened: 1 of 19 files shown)


twhittock:
ha! didn't notice it coming in sorry. well whichever one works best - doesn't really matter to me :) i didn't see the testing stuff, so the other pr is probably better from that POV.

twhittock:
> ha! didn't notice it coming in sorry. well whichever one works best - doesn't really matter to me :) i didn't see the testing stuff, so the other pr is probably better from that POV.

*this PR - I got confused which PR i was commenting on 🤦

PR #2094, opened 2026-02-13
api: block /api/schemes and /api until initialization is complete

### LINKED ISSUE #2088: GET /api/schemes returns incomplete list during startup (race condition)

twhittock:
Fixes #2088.

Closes #2089 (supersedes the previous `/api/ready` approach).

## What this does

Replaces the `atomic.Bool` ready flag with a channel that is closed once all modules have finished initializing (`SetReady()` in `main.go`, unchanged position).

`/api/schemes` and `/api` now block on that channel instead of returning partial data. When `SetReady()` fires, all concurrent waiters unblock at once (closing a channel broadcasts to all receivers). Context cancellation is respected so disconnecting clients do not leak goroutines.

After startup completes, receives from a closed channel return immediately — existing callers are unaffected.

## Why this approach

As suggested by @AlexxIT in #2089: rather than a dedicated readiness endpoint, delay the response of `/api/schemes` (and `/api`) until startup is complete.

### DIFF
--- a/internal/api/api.go
+++ b/internal/api/api.go
@@ -1,6 +1,7 @@
 package api
 
 import (
+	"context"
 	"crypto/tls"
 	"encoding/json"
 	"fmt"
@@ -11,7 +12,6 @@ import (
 	"strconv"
 	"strings"
 	"sync"
-	"sync/atomic"
 	"syscall"
 	"time"
 
@@ -19,20 +19,22 @@ import (
 	"github.com/rs/zerolog"
 )
 
-// ready is set to true after all modules have finished initializing.
-var ready atomic.Bool
+// readyCh is closed once all modules have finished initializing.
+var readyCh = make(chan struct{})
 
 // SetReady marks the server as fully initialized. Call this once all modules
 // have registered their handlers and schemes.
 func SetReady() {
-	ready.Store(true)
+	close(readyCh)
 }
 
-func readyHandler(w http.ResponseWriter, r *http.Request) {
-	if ready.Load() {
-		w.WriteHeader(http.StatusOK)
-	} else {
-		http.Error(w, "starting", http.StatusServiceUnavailable)
+// WaitReady blocks until SetReady has been called or the context is cancelled.
+func WaitReady(ctx context.Context) error {
+	select {
+	case <-readyCh:
+		return nil
+	case <-ctx.Done():
+		return ctx.Err()
 	}
 }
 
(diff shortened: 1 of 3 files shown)


felipecrs:
Hm... There's also:

- https://github.com/AlexxIT/go2rtc/pull/2092

skrashevich:
Great. We have a race condition pull requests that fix the race condition

PR #2102, opened 2026-02-20
Fix: Instead of 'src = ''' use removeAttribute('src') (video-rtc.js)

IgorA100:
An empty SRC is invalid for the VIDEO tag and will cause problems. For example, when executing `"this.video.play(),"` we'll get a minor error like 

> "DOMException: The element has no supported sources." 

But this will trigger catch(), resulting in `"this.video.muted = true,"` even though the problem is something completely different, not "muted."

### DIFF
--- a/www/video-rtc.js
+++ b/www/video-rtc.js
@@ -338,7 +338,7 @@ export class VideoRTC extends HTMLElement {
             this.pc = null;
         }
 
-        this.video.src = '';
+        this.video.removeAttribute('src');
         this.video.srcObject = null;
     }
 


felipecrs:
Potentially related to https://github.com/AlexxIT/go2rtc/pull/2090.

IgorA100:
> Potentially related to #2090.

Yes, the problem is the same...
But I didn't propose a solution to the already existing MUTED problem (due to an empty SRC attribute), but rather a way to prevent the problem from arising in the first place.

PR #2106, opened 2026-02-22
feat: add AV1 codec support for MSE, WebRTC and MP4 streaming

### LINKED ISSUE #2260: Feature request: AV1 support for MSE/fMP4 streaming
Hi,
With the recent introduction of AV1 native encoding in IP cameras — most notably Axis Communications shipping AV1 as the default codec on all new ARTPEC-9 based cameras starting in 2025 — AV1 RTSP streams from cameras are becoming increasingly common in the wild.

go2rtc already supports AV1 for WebRTC, which is great. However MSE (the low-latency browser streaming path most commonly used with Frigate) currently only handles H264 and H265. When an AV1 camera stream arrives, go2rtc has no path to serve it via MSE without transcoding to H264/H265 first.

AV1 in fMP4 containers is well-standardized (ISO 14496-15, av01 box), and all major browsers — Chrome, Firefox, Edge — support AV1 MSE natively. The infrastructure is there on both the server and client side.

Would it be feasible to add AV1 to the MSE/fMP4 muxer path, similar to how H265 MSE support was added?
This would allow AV1 camera streams to be served directly to browsers without any intermediate transcoding step — which is aligned with go2rtc's philosophy of codec passthrough over transcoding.

Thanks for the great project.

RaHehl:
AV1 cameras are becoming common, most visibly the ARTPEC-9 based Axis models that ship with AV1 as the default codec. go2rtc could negotiate such a stream but had no way to serve it without transcoding first. Closes #2260.

Adds `pkg/av1` (OBU parsing, keyframe detection, sequence header, av1C record, RTP depay/pay) and wires AV1 into the MP4 muxer, MSE, HLS, WebRTC and the browser client. A second commit from @vertoforce adds the FLV producer path, which is how ffmpeg fed AV1 reaches go2rtc.

Two things worth a reviewer's attention:

- AV1 carries its codec parameters in the sequence header, not in the SDP, so the MP4 consumer waits for one before building the init segment. The wait is limited to AV1 tracks and ends on `Stop`, so H264 and H265 are unchanged.
- Registering AV1 changes every SDP answer, and an offerer sends with the first codec of the answer. Without a guard a browser publishing through `webrtc.html` or WHIP would silently switch from H264 to AV1, so transceivers the remote sends on keep AV1 last.

Verified against ffmpeg rather than against itself: the av1C records this produces are byte for byte identical to ffmpeg's for 640x480, 1280x720, 1920x1080, 4K 10-bit and a colour tagged stream, and keyframe detection agrees with ffmpeg's own flags over 120 real temporal units. End to end, AV1 over RTSP and over FLV both decode cleanly through `stream.mp4`, `frame.mp4`, HLS and WebRTC, and H264 output is byte identical to master.

`go test ./pkg/...` fails in exactly the same 14 packages as master, none of them touched here.

### DIFF
--- a/README.md
+++ b/README.md
@@ -358,13 +358,13 @@ Some examples:
 
 ### Codecs madness
 
-`AVC/H.264` video can be played almost anywhere. But `HEVC/H.265` has many limitations in supporting different devices and browsers.
+`AVC/H.264` video can be played almost anywhere. But `HEVC/H.265` and `AV1` have many limitations in supporting different devices and browsers.
 
 | Device                                                             | WebRTC                                  | MSE                                     | HTTP*                                        | HLS                         |
 |--------------------------------------------------------------------|-----------------------------------------|-----------------------------------------|----------------------------------------------|-----------------------------|
 | *latency*                                                          | best                                    | medium                                  | bad                                          | bad                         |
-| Desktop Chrome 136+ <br/> Desktop Edge <br/> Android Chrome 136+   | H264, H265* <br/> PCMU, PCMA <br/> OPUS | H264, H265* <br/> AAC, FLAC* <br/> OPUS | H264, H265* <br/> AAC, FLAC* <br/> OPUS, MP3 | no                          |
-| Desktop Firefox                                                    | H264 <br/> PCMU, PCMA <br/> OPUS        | H264 <br/> AAC, FLAC* <br/> OPUS        | H264 <br/> AAC, FLAC* <br/> OPUS             | no                          |
+| Desktop Chrome 136+ <br/> Desktop Edge <br/> Android Chrome 136+   | H264, H265*, AV1* <br/> PCMU, PCMA <br/> OPUS | H264, H265*, AV1* <br/> AAC, FLAC* <br/> OPUS | H264, H265*, AV1* <br/> AAC, FLAC* <br/> OPUS, MP3 | no |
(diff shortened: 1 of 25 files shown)


vertoforce:
I have been running this branch for AV1 cameras, thanks for it. My setup feeds AV1
in over FLV, which the branch does not cover yet, so I added that on top:
https://github.com/vertoforce/go2rtc/commit/9f81a15330e5ce17976c92c43b86fb418586b703.
While running it I hit three bugs and wrote a test for each: the video size comes
out wrong on some streams, a viewer can hang forever waiting for a keyframe, and the
chroma subsampling in the config box is off for some profiles. Happy to open a PR
against your branch with all of it. Details below.

<details>
<summary>Details (repro, code path, numbers; written up with AI help)</summary>

ffmpeg tags AV1 in FLV as `av01`, which go2rtc skips, so an AV1 source never gets past
probe. Enhanced-RTMP carries the timing offset for HEVC only, so `av01` video starts 5
bytes in, not 8.

Bug one, wrong video size. The loop over operating points, the stream's quality layers,
skips two delay fields (`pkg/av1/config.go:89`). The bit reader then desyncs before the
frame size. A 1920x1080 header parses as 175x64 with initial_display_delay_present_flag
set, and 1x1 with decoder_model_info. Neither is zero, so the muxer's 1920x1080 fallback
never fires.

Bug two: a viewer waits for a keyframe with no way out (`pkg/mp4/consumer.go:218`). If
none arrives it never returns, even after `Stop`, and leaks a goroutine. It hits H264
and H265 too. The test passes on master c245815.

Bug three, chroma subsampling in the AV1 config box `av1C`, which AV1 spec 5.5.2 derives
from the profile.

Fixes and tests, the three commits after FLV:
https://github.com/vertoforce/go2rtc/tree/feat/av1-flv-on-pr2106

</details>

PR #2122, opened 2026-02-27
Feature: configurable minimum TLS version

btgoodwin:
Some compliance regulations require limiting TLS versions to above 1.2; this change provides users a way to configure that setting alongside the other TLS configuration options.

### DIFF
--- a/internal/api/api.go
+++ b/internal/api/api.go
@@ -130,6 +130,12 @@ func tlsListen(network, address, certFile, keyFile string, minVersion string) {
 	var minTLSVersion uint16
 	var cert tls.Certificate
 	var err error
+
+	minTLSVersion, err = parseTLSVersion(minVersion)
+	if err != nil {
+		log.Error().Err(err).Caller().Send()
+		return
+	}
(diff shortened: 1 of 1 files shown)


internal/api/api.go
```diff
@@ -108,7 +110,24 @@ func listen(network, address string) {
...
+	v := strings.ToUpper(strings.TrimSpace(versionStr))
+	if val, ok := tlsVersionMapp[v]; ok {
+		return val, nil
+	}
+	return 0, fmt.Errorf("invalid TLS version: %s", versionStr)
+}
+
+// Map string values to the crypto/tls constants
+var tlsVersionMapp = map[string]uint16{
+	"TLS1.0": tls.VersionTLS10,
```
skrashevich:
```suggestion
	if val, ok := tlsVersionMap[v]; ok {
		return val, nil
	}
	return 0, fmt.Errorf("invalid TLS version: %s", versionStr)
}

// Map string values to the crypto/tls constants
var tlsVersionMap = map[string]uint16{
	"TLS1.0": tls.VersionTLS10,
```
typo?

JeremyH76:
Is this feature only for connection to go2rtc, or also from go2rtc to a camera ?

Issue #2108, opened 2026-02-23
Nest hickup every * minutes


PrutsMeneer:
Hi!  I’m running into a very consistent issue that’s been driving me crazy for the past two weeks: parts of my Nest recordings are missing because the video feed briefly freezes every few minutes.

What happens

Using Go2RTC with Nest, the Nest video feed freezes for a couple of seconds and then recovers. This repeats very consistently, for example, exactly every 5 minutes and 24 seconds.

Each time it happens, I see the following log line:
`debug | [streams] retry=0 to url=nest:?client_id=**********&client_secret=****&device_id=*****&project_id=****&protocols=WEB_RTC&refresh_token=***`

Environment
	•	go2rtc version=1.9.14 platform=linux/amd64 revision=b5948cf

```
api:
  listen: :1984

streams:
  voor_tuin_nest:
    - nest:?client_id=****&client_secret=****&device_id=***&protocols=WEB_RTC&refresh_token=***

  achter_tuin_nest:
    - nest:?client_id=****&client_secret=****&device_id=***&protocols=WEB_RTC&refresh_token=***

rtsp:
  listen: :8554

log:
  level: info  # default level
  api: debug
  exec: debug
  rtsp: debug
  streams: trace
  webrtc: debug
```

PrutsMeneer:
Hi @AlexxIT,

I think i solved the problem. I need to admit: with some help of ChatGPT (Codex),
But I've been running this for about a week now and the problem seems to be solved; no longer i did have any reconnect! 

**What has been done:** 
- make Nest token refresh deterministic by storing credentials in API and calling OAuth refresh directly
- make ExtendStream retry behavior more robust for transient 401/409/429 responses
- keep extend timer loop running after transient extend errors to avoid unnecessary stream reconnects

**Why**
When stream extension fails once, the stream can expire and streams reconnect starts ([streams] retry=...). These changes reduce avoidable drops by improving retry and refresh behavior.

I created a PR
[PR](https://github.com/AlexxIT/go2rtc/pull/2128)

benholtz:
Ran into exactly this (4 Nest cameras, same credentials, hiccup every ~5 minutes on all but the most recently started stream) and can confirm the root cause and fix in #2351: the nest API instance is shared across streams on the same credentials, so only the last-started stream gets its session extended, and the old extend timer only ever fired once.

[...]

PR #2128, opened 2026-03-03
nest: improve stream extension resilience and token refresh

PrutsMeneer:
## Summary
- make Nest token refresh deterministic by storing credentials in API and calling OAuth refresh directly
- make ExtendStream retry behavior more robust for transient 401/409/429 responses
- keep extend timer loop running after transient extend errors to avoid unnecessary stream reconnects

## Why
When stream extension fails once, the stream can expire and streams reconnect starts ([streams] retry=...). These changes reduce avoidable drops by improving retry and refresh behavior.

## Testing
- go test ./pkg/nest

### DIFF
--- a/pkg/nest/api.go
+++ b/pkg/nest/api.go
@@ -12,8 +12,11 @@ import (
 )
 
 type API struct {
-	Token     string
-	ExpiresAt time.Time
+	Token        string
+	ExpiresAt    time.Time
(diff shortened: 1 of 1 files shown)


PrutsMeneer:
bug at: https://github.com/AlexxIT/go2rtc/issues/2108

PR #2136, opened 2026-03-06
feat(streams): add HasConsumer helper for idempotency

57Darling02:
Stream lacks a way to check if a consumer is already attached.Adding Stream.HasConsumer(cons) allows upper modules to verify the attachment status during reconnection or recovery, preventing duplicate AddConsumer calls.


AlexxIT:
It's better to add such things right away with an example of use. For now, it's useless code.

PR #2151, opened 2026-03-13
Fix ONVIF NVR compatibility for RTSP streaming

Moonbesk:
## Summary

This PR fixes three issues that cause ONVIF clients (e.g. Hikvision ISC/iSecure Center) to fail when connecting to go2rtc's ONVIF server for RTSP streaming. The failure rate was approximately 80-90%.

### Root Causes

1. **ONVIF capabilities declare RTP_TCP as false** — ONVIF clients check `GetCapabilities` and `GetServiceCapabilities` before attempting RTSP SETUP. When `RTP_TCP=false`, clients like Hikvision ISC skip the SETUP entirely, resulting in silent stream failure.

2. **RTSP server rejects UDP transport requests** — When an ONVIF client sends `SETUP` with `RTP/AVP` (UDP), the server returns `461 Unsupported transport`. Since go2rtc only supports TCP interleaved, the server should accept the request but respond with TCP interleaved transport parameters.

3. **SDP track accumulation on re-DESCRIBE** — Some ONVIF clients reuse the same RTSP connection to send multiple DESCRIBE requests (e.g. when switching profiles). The `Senders` and `Receivers` slices were never reset, causing duplicate tracks to accumulate and break playback.

### Changes

- **pkg/onvif/server.go**: Set `RTP_TCP` to `true` in both `GetCapabilities` and `GetServiceCapabilities` responses
- **pkg/rtsp/server.go**: Accept `RTP/AVP` (UDP) SETUP requests and respond with TCP interleaved transport
- **internal/rtsp/rtsp.go**: Clean up previous consumer and reset Senders/Receivers on re-DESCRIBE

### Testing

Tested with Hikvision ISC (iSecure Center) NVR platform connecting to go2rtc v1.9.14 ONVIF server. After the fix, streams connect reliably (previously ~80-90% failure rate).

### DIFF
--- a/internal/rtsp/rtsp.go
+++ b/internal/rtsp/rtsp.go
@@ -167,6 +167,14 @@ func tcpHandler(conn *rtsp.Conn) {
 				return
 			}
 
+			// Clean up previous consumer on re-DESCRIBE (e.g. ONVIF clients)
+			if closer != nil {
+				closer()
+				closer = nil
+			}
+			conn.Senders = nil
+			conn.Receivers = nil
+
 			name = conn.URL.Path[1:]
 
 			stream := streams.Get(name)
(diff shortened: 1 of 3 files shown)


sukhneerguron:
Hey @AlexxIT, just wanted to bump this. I ran into the exact same issue getting a Hikvision DVR to add go2rtc as an ONVIF source, and testing this PR fixed it for me. Would be great to see this land when you get a chance to look at it.

PR #2159, opened 2026-03-16
Fix PCMA and PCMU handling in incoming streams

ludufre:
This pull request adds support for two new audio stream types, PCMA and PCMU (Tapo variants), to the MPEG-TS producer. The main changes ensure that these audio types are recognized and properly mapped to their respective codecs and media definitions.

Support for new audio stream types:

* Added `StreamTypePCMATapo` and `StreamTypePCMUTapo` to the list of recognized stream types in the probing logic, so the producer now waits for these types when encountered.
* Implemented handling for `StreamTypePCMATapo` and `StreamTypePCMUTapo` by mapping them to the appropriate `core.Codec` (`CodecPCMA` and `CodecPCMU` respectively) with a clock rate of 8000, and appending the corresponding `core.Media` entries to the producer's media list.

### DIFF
--- a/pkg/mpegts/producer.go
+++ b/pkg/mpegts/producer.go
@@ -175,6 +175,8 @@ func StreamType(codec *core.Codec) uint8 {
 		return StreamTypeAAC
 	case core.CodecPCMA:
 		return StreamTypePCMATapo
+	case core.CodecPCMU:
+		return StreamTypePCMUTapo
 	case core.CodecOpus:
 		return StreamTypePrivateOPUS
 	}
--- a/pkg/mpegts/producer.go
+++ b/pkg/mpegts/producer.go
@@ -91,7 +91,7 @@ func (c *Producer) probe() error {
 		case StreamTypeMetadata:
 			for _, streamType := range pkt.Payload {
 				switch streamType {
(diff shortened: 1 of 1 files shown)


pkg/mpegts/producer.go
skrashevich, line 166: Did you forget to add PCMU here?

ludufre: You're right! I fixed it now.

skrashevich (commented):
lgtm for dev branch

AlexxIT:
Why do we need this? You won't find MPEG-TS files in these formats in the real world. It's not a standard.
And Tapo's PCMU isn't 8000 at all.

ludufre:
This is needed for ICsee/XMEye cameras streamed via an exec source I just made. These cameras use the DVRIP via proprietary tunnel protocol which delivers MPEG-TS + G.711a (PCMA) audio. 

The demuxer's GetPacket() already handles StreamTypePCMATapo, but the probe() doesn't detect it, so the audio is silently dropped. I removed the PCMU case since my use case only needs PCMA (which is always 8000 Hz per G.711 spec).

AlexxIT (approved):

AlexxIT:
OK. The changes are minor and safe. Although, PCMA within MPEG-TS isn't a standard at all. Even FFmpeg doesn't support it. It just sets a stream type of 6 for any unknown codec.

PR #2160, opened 2026-03-17
WebCodecs Support

seydx:
This PR adds a new WebCodecs-based streaming mode to go2rtc, delivering raw encoded frames over WebSocket for browser-side decoding via the WebCodecs API.

## Features

- Raw encoded video/audio frames via WebSocket
- H.264 and H.265 video support
- AAC, Opus, PCMA, PCMU audio support (no server-side encoding needed)
- Binary frame protocol with 9-byte header
- AVCC to Annex-B conversion for browser compatibility
- Native-style player controls (play/pause, volume, fullscreen)
- GPU-accelerated rendering: WebGPU → WebGL2 → Canvas 2D (automatic fallback)
- Configurable renderer cascade via `renderer` query parameter

## Usage

```
# Default (full cascade)
stream.html?src=camera&mode=webcodecs

# Force specific renderer
stream.html?src=camera&mode=webcodecs&renderer=webgl

# Skip WebGPU
stream.html?src=camera&mode=webcodecs&renderer=webgl,2d
```


www/video-rtc.js
```diff
@@ -501,6 +526,218 @@ export class VideoRTC extends HTMLElement {
...
+                    error: () => {
+                        this._audioDecoder = null;
+                    },
+                });
+                this._audioDecoder.configure({
+                    codec: info.audio.codec,
+                    sampleRate: info.audio.sampleRate,
+                    numberOfChannels: info.audio.channels,
+                });
+            }
```
skrashevich: potential audio drift?
_next_time increased without compensation

www/links.html
```diff
@@ -55,6 +55,7 @@ <h2>Any codec in source</h2>
             <h2>H264/H265 source</h2>
             <li><a href="stream.html?src=${src}&mode=webrtc">stream.html</a> WebRTC stream / browsers: all / codecs: H264, PCMU, PCMA, OPUS / +H265 in Safari</li>
             <li><a href="stream.html?src=${src}&mode=mse">stream.html</a> MSE stream / browsers: Chrome, Firefox, Safari Mac/iPad / codecs: H264, H265*, AAC, PCMA*, PCMU*, PCM* / +OPUS in Chrome and Firefox</li>
+            <li><a href="stream.html?src=${src}&mode=webcodecs">stream.html</a> WebCodecs stream / browsers: Chrome, Edge, Firefox, Safari Mac/iPad / codecs: H264, H265, AAC, OPUS, PCMA, PCMU</li>
```
skrashevich, line 58: > Firefox

sure?

pkg/webcodecs/consumer.go
```diff
@@ -0,0 +1,257 @@
...
+			}
+
+			payload := annexb.DecodeAVCC(packet.Payload, true)
+			flags := byte(0x80) | trackID // video flag
+			if keyframe {
+				flags |= 0x40 // keyframe flag
+			}
+
+			c.mu.Lock()
+			msg := buildFrame(flags, packet.Timestamp, payload)
```
skrashevich: I’m absolutely not a specialist in video codecs, but isn’t it supposed to be microseconds here? And the chain all the way to the client?

www/links.html
AlexxIT: https://caniuse.com/?search=VideoDecoder

www/video-rtc.js
seydx: Fixed

`_nextTime` now resets to `currentTime` when it falls behind, prevents drift after stalls or suspend/resume.

pkg/webcodecs/consumer.go
seydx: Good catch — timestamps are now converted to microseconds before sending. It worked before because the decoder only uses timestamps for ordering, not absolute timing. But the spec says microseconds, so we should follow it.

www/links.html
seydx: WebCodecs is supported in all major browser, but hevc not:

### Microsoft Edge on Windows
By default, Chromium browsers uses FFMpeg internally for WebCodecs API. However, Microsoft Edge, when running on Windows, uses Media Foundation decoders instead.

Decoding H.265 requires the [HEVC Video Extensions](https://www.microsoft.com/store/productId/9NMZLZ57R3T7) ($0.99) or HEVC Video Extensions from Device Manufacturer (free but not available anymore) app from Microsoft Store.

### Firefox
Firefox 133 supports playing H.265 videos, but does not support decoding H.265 streams using WebCodecs API yet.

AlexxIT:
1. I like using new, modern technologies - it's cool.
2. I don't like the large amount of JavaScript code in the main player. Ideally, move everything into a separate file and make it easily connectable.
3. I'm not sure about the header for the binary format. It needs to be carefully designed. Otherwise, it will be difficult to change later. You're using `uint32` for the time. That might not be enough. The header includes the packet size (maybe that's a good thing), but it isn't being used.

seydx:
> 1. I like using new, modern technologies - it's cool.
> 2. I don't like the large amount of JavaScript code in the main player. Ideally, move everything into a separate file and make it easily connectable.
> 3. I'm not sure about the header for the binary format. It needs to be carefully designed. Otherwise, it will be difficult to change later. You're using `uint32` for the time. That might not be enough. The header includes the packet size (maybe that's a good thing), but it isn't being used.

- Code is now split into three modules: video-webcodecs.js (player/decoder/audio/controls), video-renderer.js (WebGPU/WebGL2/Canvas2D cascade), and the existing video-rtc.js just wires them together
- Header redesigned: dropped the unused payload length field, timestamp is now uint64 (microseconds). 9 bytes total: flags(1) + timestamp_us(8) + payload.

skrashevich:
merged into dev branch
@seydx please note the fixes in da908ca6c1cd96c43ca65ed87e96f2cc1f82b60a

PR #2193, opened 2026-03-31
fix(h264,homekit): add bounds guards against malformed device data

MechanicalCoderX:
`h264 GetFmtpLine`: check `len(avc) >= 8` and `size <= len(avc)` before accessing `avc[5:8]` for the SPS profile bytes. A truncated SPS NALU from a misbehaving device caused an index-out-of-range panic.

`homekit videoToMedia`: clamp `profileID` and `level` to the bounds of the `videoProfiles`/`videoLevels` tables before indexing them. A non-compliant HomeKit device advertising an out-of-range profile or level value caused an index-out-of-range panic.

### DIFF
--- a/pkg/h264/h264.go
+++ b/pkg/h264/h264.go
@@ -130,6 +130,9 @@ func GetFmtpLine(avc []byte) string {
 
 		switch NALUType(avc) {
 		case NALUTypeSPS:
+			if len(avc) < 8 || size > len(avc) {
+				return s
+			}
 			s += ";profile-level-id=" + hex.EncodeToString(avc[5:8])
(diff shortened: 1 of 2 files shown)


ajplotkin:
Adopted these bounds guards in my build. I capture SPS/PPS from the incoming Nest H264 RTP to synthesize `sprop-parameter-sets` for the RTSP SDP, so `GetFmtpLine` runs on real device data on my hot path — a truncated/malformed SPS from a camera would otherwise be an index-out-of-range panic that takes down the whole process. Small, obviously-correct, defends a real crash. +1.

PR #2194, opened 2026-03-31
fix(nest): fix stream expiry, token refresh, timeouts, and resource leaks

MechanicalCoderX:
- `ExtendStream`: replace one-shot `*time.Timer` with an `extendStop` channel loop so streams are extended reliably every ~55 seconds instead of expiring after the first extension fires and is not rescheduled

- `refreshToken`: store `ClientID`/`ClientSecret`/`RefreshToken` on the `API` struct so `refreshToken()` can re-authenticate directly; the previous cache-key lookup failed after the first token rotation because the cached entry had already been replaced

- `NewAPI`/`GetDevices`/`ExchangeSDP`: fix HTTP client timeout of `5000 * time.Second` to `10 * time.Second`

- `rtcConn`: call `pc.Close()` on all three error paths (CreateCompleteOffer, ExchangeSDP, SetAnswer) to prevent WebRTC peer connection leaks

- Fix typos `cliendID` → `clientID`, `cliendSecret` → `clientSecret`, `compataiility` → `compatibility`

### DIFF
--- a/pkg/nest/api.go
+++ b/pkg/nest/api.go
@@ -15,6 +15,12 @@ type API struct {
 	Token     string
 	ExpiresAt time.Time
 
+	// Credentials stored so refreshToken() can call OAuth directly without
+	// searching the cache (which breaks after the first token rotation).
+	ClientID     string
+	ClientSecret string
+	RefreshToken string
+
 	StreamProjectID string
 	StreamDeviceID  string
 	StreamExpiresAt time.Time
@@ -26,7 +32,8 @@ type API struct {
 	StreamToken          string
 	StreamExtensionToken string
 
-	extendTimer *time.Timer
+	extendMu   sync.Mutex
+	extendStop chan struct{}
 }
 
 type Auth struct {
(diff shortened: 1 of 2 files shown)


ajplotkin:
Adopted the HTTP-client timeout part of this in my deployment (Nest → go2rtc → Homebridge/HomeKit) and it's a genuine fix — the stock `Timeout: time.Second * 5000` is ~83 minutes, so a single hung request to Google blocks that goroutine for over an hour instead of failing fast and letting the producer reconnect. Dropped it to 10s across the `api.go` request paths.

For the session-expiry/token-refresh piece there's overlap with #2351 (which I ended up using for the extension loop because it's `api.go`-only and doesn't touch `client.go`'s `rtcConn`, where I carry a local udp4-ICE patch). Worth the maintainers reconciling the two — between them they cover the one-shot-timer bug, the shared-`*API` state bug, the 83-min timeout, and the `pc.Close` leak. One gap neither fully handles: transient extend/refresh errors are treated as terminal, so a momentary Google 5xx silently kills extension and the stream freezes ~10 min later without a reconnect. I retry those a few times before giving up. Thanks for digging into this area — it needed it.

PR #2216, opened 2026-04-16
Fix/dtls frame resync

### LINKED ISSUE #2215: Wyze TUTK producer: [OOO] reset loop discards entire frame and breaks HLS playback
Affected version: v1.9.14 (revision b5948cf); code still present on master.

Component: pkg/tutk/frame.go, FrameHandler.handleVideo.

Reproduction:

- Configure a Wyze V3 (WYZE_CAKP2JFUS) or V4 (HL_CAM4) source:
```yaml
 streams:
   - cam: wyze://USER:PASS@CAM_MAC?uid=...&enr=...&model=WYZE_CAKP2JFUS
 ```
- Open http://localhost:1984/api/stream.m3u8?src=cam in VLC.
- Observe: one frame renders at 0:00, playback stalls at ~0:01 (end of first fMP4 segment).
- Concurrently play rtsp://localhost:8554/cam in VLC — plays fine.

**Expected:** HLS streams continuously like RTSP does.
**Actual:** HLS freezes after ~1s. stdout shows repeated:

```text
[OOO] ch=0x05 #3643 frameType=0x00 pktTotal=107 expected pkt 0, got 102 - reset
[OOO] ch=0x05 #3643 frameType=0x00 pktTotal=107 expected pkt 0, got 103 - reset
[OOO] ch=0x05 #3643 frameType=0x00 pktTotal=107 expected pkt 0, got 104 - reset
[OOO] ch=0x05 #3643 frameType=0x00 pktTotal=107 expected pkt 0, got 105 - reset
[OOO] ch=0x05 #3643 frameType=0x01 pktTotal=107 expected pkt 0, got 106 - reset
```

Note that `expected` is 0 on every line for the same frame. After the first reset the handler should either advance or stop processing this `FrameNo`, not re-enter with `waitSeq=0` each time.

Root cause: In [ `handleVideo` ](https://github.com/AlexxIT/go2rtc/blob/master/pkg/tutk/frame.go#L368), when [ `hdr.FrameNo != cs.frameNo` ](https://github.com/AlexxIT/go2rtc/blob/master/pkg/tutk/frame.go#L376) the state is unconditionally re-seeded `waitSeq = 0`. The next check [ `fi hdr.PktIdx != cs.waitSeq` ](https://github.com/AlexxIT/go2rtc/blob/master/pkg/tutk/frame.go#L388) then fails for any mid-frame packet, calls `cs.reset()` (zeroing `frameNo`), and returns. The next continuation packet for the same `FrameNo` retriggers the same path. The assembler clobbers itself on every stray packet until a new keyframe starts. RTSP tolerates the resulting frame drops but HLS's fMP4 muxer cannot produce segment 2.

Suggested patch:  Don't reinitialize when `hdr.PktIdx == 0`, and on _OOO_ inside a `hasStarted` frame set `cs.waitSeq = cs.pktTotal` as a sentinel rather than calling `cs.reset()` so later continuation packets for the same `FrameNo` are ignored without disturbing state:

```go
 if hdr.FrameNo != cs.frameNo {
+    if hdr.PktIdx != 0 {
+        return // stray continuation for a frame we never started / gave up on
+    }
     if cs.hasStarted && cs.waitSeq < cs.pktTotal {
         fmt.Printf("[DROP] ch=0x%02x #%d INCOMPLETE: got %d/%d pkts\n",
             channel, cs.frameNo, cs.waitSeq, cs.pktTotal)
     }
     cs.reset()
     cs.frameNo  = hdr.FrameNo
     cs.pktTotal = hdr.PktTotal
 }
 if hdr.PktIdx != cs.waitSeq {
-    fmt.Printf("[OOO] ... - reset\n", ...)
-    cs.reset()
+    fmt.Printf("[OOO] ... - drop frame\n", ...)
+    cs.waitData  = cs.waitData[:0]
+    cs.frameInfo = nil
+    cs.hasStarted = false
+    cs.waitSeq   = cs.pktTotal // sentinel: ignore remaining pkts of this frame
     return
 }
```
Note: [OOO], [DROP], [SIZE], [RX], [OK] are emitted via fmt.Printf rather than the project's structured logger, perhaps we should be routing through the logger at appropriate levels.

With this correction a single packet loss discards exactly one frame; all subsequent packets of that frame are silently ignored; assembly resumes cleanly at the next frame start. Same path is exercised for all TUTK sources (Xiaomi/Kasa/Tapo), not just Wyze, HLS users notice it most visibly.

IDisposable:
Drops rest of frame when the packets are out of sync instead of just reset and never recover
Fixes #2215


IDisposable:
Wrong target branch, sorry.

PR #2223, opened 2026-04-22
fix(api): add support for CORS preflight requests

### LINKED ISSUE #2218: CORS preflight OPTIONS fails with 401 when authentication is enabled
## Description
When both origin: '*' and username/password are configured in the api section, all cross-origin requests from browsers fail with a CORS error. I didn't succeed to connect to a stream in WebRTC.

## Environment
I have the following config:
```
api:
  origin: '*'
  listen: ':1984'
  username: 'admin'
  password: 'pass'
```

## Steps to reproduce
From a browser (different origin, like localhost:4200), make any fetch() POST request to /api/webrtc.
Example:
fetch("http://localhost:1984/api/webrtc?src=mycamera", {
  method: "POST",
  headers: {
    "Content-Type": "application/sdp",
    "Authorization": "Basic " + btoa("admin:pass")
  },
});
The browser sends a CORS preflight OPTIONS request first.

## Observed behavior
OPTIONS http://localhost:1984/api/webrtc?src=mycamera  →  401 Unauthorized
POST    http://localhost:1984/api/webrtc?src=mycamera  →  CORS error (blocked by browser)

## Related issues
#850 — same problem, closed without fix

skrashevich:
- handle OPTIONS method in middlewareAuth to allow preflight requests
- handle OPTIONS method in middlewareCORS to respond with no content
fix #2218

### DIFF
--- a/internal/api/api.go
+++ b/internal/api/api.go
@@ -211,6 +211,11 @@ func isLoopback(remoteAddr string) bool {
 
 func middlewareAuth(username, password string, localAuth bool, next http.Handler) http.Handler {
 	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		if r.Method == "OPTIONS" {
+			next.ServeHTTP(w, r)
+			return
+		}
+
 		if localAuth || !isLoopback(r.RemoteAddr) {
 			user, pass, ok := r.BasicAuth()
 			if !ok || user != username || pass != password {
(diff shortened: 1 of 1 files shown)


fclage (commented):
This change would solve the current issue I'm facing with authentication in go2rtc and a custom application I'm currently working on (netip-cctv).  I would suggest it is merged w/ master.

An `OPTIONS` HTTP verb/request should **NOT** require authentication IMOHO.

> HTTP OPTIONS requests should not be authenticated.In fact, enforcing authentication on OPTIONS requests will break web applications. The WHATWG Fetch Specification explicitly states that web browsers must exclude credentials (like HTTP Authorization headers or cookies) from CORS preflight requests. If your server requires authentication for OPTIONS, the browser's preflight check will fail with a 40[1](https://github.com/pomerium/pomerium/issues/78) Unauthorized error, causing the browser to block the subsequent actual request (POST, PUT, DELETE) entirely

Resources:
- https://fetch.spec.whatwg.org/#cors-preflight-fetch
- https://developer.mozilla.org/en-US/docs/Glossary/Preflight_request

PR #2263, opened 2026-05-17
feat: Native Reolink Baichuan protocol stream & talkback integration

### LINKED ISSUE #755: Integrate Reolink's (reverse engineered) proprietary protocol?
Reolink's RTSP (and RTMP) stream sucks and is pretty unstable. But their own protocol is more reliable, but it's proprietary. 

[George Hilliard has reverse engineered Reolink's proprietary protocol](https://www.thirtythreeforty.net/posts/2020/05/hacking-reolink-cameras-for-fun-and-profit/) and even created a [software client (Neolink) to take the Reolink protocol and spit it out into an RTSP stream](https://github.com/thirtythreeforty/neolink). (A more recently updated fork of the software is available here: [https://github.com/QuantumEntangledAndy/neolink](https://github.com/QuantumEntangledAndy/neolink)) 

It would be great if this protocol was integrated directly into go2rtc!

fzurita:
#### ** AI Generated Overview**
This pull request introduces **native support for Reolink's proprietary Baichuan protocol** inside `go2rtc`!

---

##### **1. Direct Baichuan Protocol Integration (`pkg/reolink`)**
* Created the `reolink` package to dial, log in, and manage native preview/talkback channels directly.
* Implemented the custom `reolink://` schema to easily configure and stream from any Reolink camera.

##### **2. Dynamic Stream Probing & Codec Advertisement**
* Reolink cameras produce either H.264 or H.265. To prevent static misconfigurations, we introduced a **Probe phase**.
* Upon connection, the probe dynamically spins up the stream, waits for the first I-frame to detect the codec (H.264 or H.265), extracts SPS/PPS/VPS parameters dynamically for instant player startup, and then initializes the tracks.

##### **3. Monotonic Timestamp Unwrapping & B-Frame Presentation Integrity**
* **Microsecond Unwrapping**: Reolink's 32-bit hardware microsecond timestamps wrap around every 71.58 minutes. We implemented a continuous 64-bit microsecond `timestampUnwrapper` to align audio/video timelines and avoid recording drops.
* **B-Frame Presentation Preservation**: Implemented `rtpTimestampGuard` to keep timestamps monotonic while **preserving B-frame presentation order** by allowing small backward display-order jumps (<= 1s). This completely eliminates video recording skipping and stuttering in downstream NVRs (like Frigate).

##### **4. Seamless Two-Way Audio (Talkback)**
* Added native talkback setup utilizing PCM-to-ADPCM transcoding.
* Implemented a graceful 5-second idle closing timer to clean up talkback sessions when not in active use.

##### **5. Connection Resilience & Structured Logging**
* The stream connection only triggers a reset after **3 consecutive packet read failures**, preventing continuous reconnect loops on minor network blips.

---

#### ** Verification**
* Fully compiled and validated on `windows/amd64` (`go2rtc.exe`) and cross-compiled for `linux/amd64` (`go2rtc_linux`).
* Extensively tested with 12 simultaneous streams coming from Reolink Duo 2 Wifi, Reolink Doorbell, and 2 X Reolink Trackmix Wifi 6 (2 channels). Each camera has a main and a sub stream.

---

#### ** Credits**
This is heavily based from [reolinkproxy](https://github.com/Shareed2k/reolinkproxy). Most of the source code comes from there and it was integrated into go2rtc.

Fixes https://github.com/AlexxIT/go2rtc/issues/755


fzurita:
I also had to implement pacing adding a small sleep between packets for h264 and h265. Without that WebRTC is constantly skipping.

Legendhidde:
I’ve been using this fork with my Reolink cameras for a while. It is rock solid. But I encounter one issue.

I have Reolink Trackmix cameras with dual lens, one zoom and one normal. I connect to both with Reolink://username:password@IPADDRESS/main?channel=0 and for the zoom I make another camera stream for channel=1.

But only with the Trackmix cameras this results in failed to load via the Reolink app. Maybe there are too many active connections but I was able to get 2 full resolution RTSP streams from the zoom and wide angle lens before and the Scrypted Reolink plugin doesn’t block my Reolink app from working. Even with both lenses configured and consumed.

fzurita:
I do have those cameras as well and I did experience that but I attributed to too many connections. But given that you say that scrypted doesn't have this problem then it sounds like I maybe doing something wrong. I'll take a look and see if I can figure out why.

Legendhidde:
Thank you! So far I’ve worked around it by just using the low res stream in the Reolink app, that works and recordings on the Reolink and tracking are not affected luckily. You’ve single handedly fixed every bug of my smart home system I built on Frigate and Reolink cameras.

fzurita:
I maybe miss reading the source code for scrypted, but it doesn't seem to use the Reolink protocol for video streams?

Legendhidde:
There’s supposed to be an official Reolink plugin for Scrypted, but I can’t really find what protocol that uses exactly

fzurita:
Ok, I wonder if that would have the same problem.

fzurita:
There is a plugin for scrypted: https://github.com/apocaliss92/scrypted-reolink-native

But I wonder if that would run into the same issue.

fzurita:
> Thank you! So far I’ve worked around it by just using the low res stream in the Reolink app, that works and recordings on the Reolink and tracking are not affected luckily. You’ve single handedly fixed every bug of my smart home system I built on Frigate and Reolink cameras.

Ok, from the research I did, there is a limit of 2 main stream connections through the reolink protocol and it's limited by the camera FW. If you have go2rtc viewing channel 0 and channel 1, that's two streams right there which leaves none for the reolink app.

quenthal:
This has been rock solid for me as well. ~~But I'm having one single trouble with with two way audio. If I enable microphone on my laptop on Frigate UI, I can't hear anything from doorbell side. If I disable microphone, only then I can hear audio coming from doorbell. What should I do to achieve duplex?~~

Edit: configuration error on my side! Works most excellent!

track0x1:
@AlexxIT @skrashevich Any chance we can get this released?

skrashevich:
@AlexxIT it's your magic.

wallacebrf:
Using frigate 

Just found this fork of go2rtc and i was able to change my RLC-1212A main stream to the full resolution of 4512*2512 and have the stream stable for the FIRST TIME. this is awesome! i used to get the camera stream to fail and then i see no video, now seeing live is seamless and smooth

i will let it record video too as i previously also had very stutter filled video even on a lower resolution of 2560x1440

The sooner this is added to the main release the better as it solves the headache that is using Reolink cameras

fzurita:
I seem to still be struggling with some audio/video synchronization issues when running over several days, but only on some cameras. I'm hoping this latest commit addresses this.

fzurita:
Also, another thing. If a camera disconnects and reconnects, sometimes frigate attaches to the opus stream which is audio only which causes invalid segments. This seems to be a bug in frigate since if you reset the stream from frigate after the camera reconnects, it will attach itself the correct stream.

NickM-27:
In the latest version Frigate automatically checks that video is included in the recordings and will restart the stream if there is none

I don't see how this could be a Frigate bug, it has no way of deciding which stream to attach to other than the rtsp url which never changes

fzurita:
Possibly, I've been using frigate 0.17. I'll keep investigating that. What leads me to believe that it's a frigate bug is that killing the frigate ffmpeg producing the invalid recording segment seems to put it back into a good state once frigate restarts it.

NickM-27:
Right, because go2rtc goes back to sending the correct data that frigate is requesting via rtsp

fzurita:
@NickM-27 Do you think it would be a good idea for frigate to restart ffmpeg if it's producing invalid segments continuously?

I can't think of a good way to force back to the main stream for recording when a disconnect/reconnect happens from go2rtc.

NickM-27:
That's already how it works. Previously it only checked if ffprobe could properly probe but in 0.17 it also checks for video and restarts if there is no video included in the recording segments

https://github.com/blakeblackshear/frigate/blob/f0d69f7856ee548ddd079dd2054a2d71826868b8/frigate/record/maintainer.py#L335

https://github.com/blakeblackshear/frigate/blob/f0d69f7856ee548ddd079dd2054a2d71826868b8/frigate/video.py#L400

fzurita:
Just to explain better, here is my recommended config for talkback:
```yaml
streams:
  doorbell_1_talkback:
    - reolink://admin:password@192.168.1.123/main
    - reolink://admin:password@192.168.1.123/extern
    - ffmpeg:doorbell_1#audio=opus
```

If someone re-uses this same config for recording, and the camera disconnects and reconnects, I don't think the opus stream ever ends, so frigate ends up connecting to that. Although.... Now that I think about it, it's probably easier to just make a secondary config only for recording and I think that would never happen.
```yaml
streams:
  doorbell_1_recording:
    - reolink://admin:password@192.168.1.123/main
  doorbell_1_talkback:
    - ffmpeg:doorbell_1_recording
    - reolink://admin:password@192.168.1.123/extern
    - ffmpeg:doorbell_1#audio=opus
  doorbell_2_talkback:
    - reolink://admin:password@192.168.1.123/extern
    - ffmpeg:doorbell_1#audio=opus
  doorbell_2_detect:
    - reolink://admin:password@192.168.1.123/extern
```

wallacebrf:
Fzurita, when using frigate I hear audio when watching a stream live so I assumed audio was recorded to file. However I tried for the first time to watch a recorded file to see if I was having any audio sync issues you mentioned and I noticed I had no audio at all. 

I did NOT include the 
ffmpeg:my_camera_name#audio=opus

As I thought that was only needed for two way talk which I am not using. Do I need the ffmpeg line for audio to properly record to file?

fzurita:
You shouldn't, that's only if you want to transcode the audio. I haven't had to transcode the audio for recordings to have audio, which means it should be using the native camera audio format. If for some reason your browser doesn't support the native audio format from the recording, you could add a line to perform transcoding.

wallacebrf:
I was afraid of that as the audio recorded to file fine before I made the change from rstp to the reolink option you created 

And it is not a browser proby. The live view I can hear the audio fine. It is the recorded files that have no audio and I have even used VLC to play the files. 

Edit
I checked files from the reolink camera when it was using RSTP before I used your code and it did not record audio to file then either. I checked a different camera that also has audio (amcrest) and those files have audio, so it is something with my configuration and not with your code 

I will have to see why the camera is not saving audio to file for the reolink

Edit 2

I am dumb. On the reolink config under the camera config I forgot to include the ffmpeg output_args record preset-record-generic-audio-copy

starkillerOG:
Would be great if this would get implemented in the offficial go2rtc, then I can easily add support for life streams of battery cameras in the built in Reolink Home Asssitant integration.
Would safe me the work of having to implement the stream in reolink_aio and then passing it allong to go2rtc.

rddelacosta:
Are there any updates on when this will be merged? Thanks

WiredLife:
it works great but after some time i got this in the frigate logs from my reolink doorbell poe:
No frames received from klingel_ww in 20 seconds. Exiting ffmpeg...
and then tons of this:
[vost#0:0/copy @ 0x5636baf4fa80] Non-monotonic DTS; previous: 419648789, current: 611417; changing to 419648790. This may result in incorrect timestamps in the output file.

did i something wrong?

edit: firmware is v3.0.0.6460_2605211898 and i'm using the main stream for detections

fzurita:
Rebased with master and fixed potential memory leak. Also, fix issues where streams start not during key frames.

WiredLife:
with the new version the video starts with pixelated (sometimes green) image on live view (when still changes to substream) and in the recordings, looks same with pr-2413
and two way audio on mainstream did not work anymore (no microphone icon)
did clean build with
> git clone https://github.com/AlexxIT/go2rtc.git
git fetch origin pull/2263/head:pr-2263
git checkout pr-2263

the old version, before the new changes runs smoothly
i use /extern for substream

fzurita:
Currently, it's intended to not allow talkback on the main stream. I did have issues with one of my cameras supporting talkback on the main stream. Since it's not a guaranteed working config, I disabled it. I'll see if anything changed that could cause that, maybe not sending frames until a keyframe is available is causing issues in your setup.

WiredLife:
great work
can you please add two way audio on mainstream so we dont need a second line to substream to activate that?

fzurita:
Yeah, I'll add a parameter to disable it for streams that can't support it.

fzurita:
Ok, feel free to try that. Talkback can now be disabled on the main stream with ```backchannel=false```

WiredLife:
works like a charm, thank you very much <3

ReX1983:
Out of curiosity, any idea if this will be ultimately merged or there is something preventing it (e.g. project or architecture decision)?

xconverge:
I am guessing it is solely that @AlexxIT has other things going on in life and this repo hasn't seen much activity in a few months.

This PR/branch works great, so I do not believe it has anything to do with the architecture or implementation of this specific feature.

ReX1983:
I have seen another fork which implements Baichuan and provides a comparison of their implementation with the one from @fzurita. I can't find the link right now, but I was curious if the two will converge or if not, which one will eventually be merged (if any).

Here is it: https://github.com/AlexxIT/go2rtc/pull/2413

WiredLife:
i've tested 2413 and this here has way less artifacts/pixalated images on fast changing objects/switching from still to sub on live view and so on

WiredLife:
can you tell me the best practice to rotate the picture 90 degrees?
currently i'm using this

> ffmpeg:http://192.168.12.11/flv?port=1935&app=bcs&stream=channel0_main.bcs&user=admin&password=admin#video=h264#rotate=90#audio=opus#hardware

wallacebrf:
> i've tested 2413 and this here has way less artifacts/pixalated images on fast changing objects/switching from still to sub on live view and so on

do you have a preferred version based on your testing? either #2263 or #2413?

ChristianCavallo96:
Has anyone tested this with cameras connected through a Reolink Home Hub (Not Pro)?

I'm curious whether it works when pointing go2rtc at the hub's IP with the camera's channel, something like:

```
streams:
  camera:
    - reolink://admin:password@HUB_IP/main?channel=0
```

In particular:

Does streaming work for both wired/Wi-Fi cameras and battery cameras (including the battery doorbell)?
Does talkback work through the hub?
Any issues with wake-up delay or dropped streams on battery cameras?

Thanks for the great work on this!

Issue #1982, opened 2025-12-15
Known Xiaomi cameras


AlexxIT:
Currently, [Xiaomi source](https://github.com/AlexxIT/go2rtc/blob/master/internal/xiaomi/README.md) only supports the CS2 protocol. The TUTK protocol is not supported.

Please tell us which models worked for you, what vendor and codecs they use.

| Name                               | Models                                  | Vendor | Codecs     | Comments   |
| ---------------------------------- | --------------------------------------- | ------ | ---------- | ---------- |
| Xiaomi Outdoor Camera CW100 Dual   | chuangmi.camera.075ae1                  | c2     | hevc, opus | dual       |
| Mi 360 Home Security Camera 2K Pro | chuangmi.camera.021a04                  | с2     |            |            |
| Mi 360 Home Security Camera 2K     | chuangmi.camera.029a02                  | с2     |            |            |
| IMILAB Security Camera EC3 Pro     | chuangmi.camera.042a02                  | c2     |            |            |
| Xiaomi Smart Camera C200           | chuangmi.camera.046c04                  | c2     | hevc, opus | have       |
| IMILAB EC5 Floodlight Camera       | chuangmi.camera.055a02                  | с2     |            |            |
| IMILAB EC5 Floodlight Camera       | chuangmi.camera.055c02                  | с2     |            |            |
|                                    | chuangmi.camera.72ac1                   | c2     | hevc, opus | dual, have |
|                                    | chuangmi.camera.v2                      | tutk   |            |            |
|                                    | isa.camera.df3                          | tutk   |            | have       |
| Mi Home Security Camera 1080p      | isa.camera.hlc6                         | c2     | hevc, pcma | have       |
| Mi Camera 2K                       | isa.camera.hlc7 / MJSXJ03HL / BHR5255GL | с2     | hevc, pcma | have       |
| Mi Home Security Camera            | mijia.camera.v1                         | tutk   |            |            |

titovskiy:
по IMILAB Security Camera EC3 Pro
chuangmi.camera.042a02

"medias": [
        "video, recvonly, H265",
        "audio, recvonly, PCMA/8000",
        "audio, sendonly, PCMA/8000"
      ],

      "protocol": "cs2+udp",

artt652:
Xiaomi室外摄像机CW400
isa.camera.hlc8
MJSXJ04HL

"medias": [
        "video, recvonly, H265",
        "audio, recvonly, OPUS/48000/2",
[...]

Issue #255, opened 2023-02-13
Dahua vto 2wayaudio and audio settings


morpheus8888:
![image](https://user-images.githubusercontent.com/49568239/218529338-abad941a-4bef-4dae-9ac9-78f1b40c30a9.png)

I am using the 2wayaudio function on my VTO dahua 2202F with satisfaction. The problem is that to make it work I had to set the correct bitrate with the happy onvif program, otherwise it doesn't work. In the vto web page I don't have any audio configuration to set. Everything works fine for a few days, after which the audio settings go back to the default settings and nothing works anymore and I have to set it again. Is there a definitive procedure to prevent this? The vto is updated to the latest firmware.

felipecrs:
Yes, there is:

- This you can do: https://github.com/AlexxIT/go2rtc/issues/49#issuecomment-1288196783
- Maybe go2rtc will also support this in the future: https://github.com/AlexxIT/go2rtc/issues/105

So, I'd suggest you close this issue.

AlexxIT:
#49

Issue #254, opened 2023-02-12
ampersand escaping question (Frigate related)


randyg503:
Pasting the appropriate portion of the below URL into my browser results in auth failure.  If I replace '\u0026' with an ampersand successful auth is achieved.  Should the code be escaping this? 

{
  "producers": [
    {
      "url": "ffmpeg:http://192.168.1.xxx:2323/?cmd=getCamshot\u0026password=xxxxx"
    }
  ],
  "consumers": null
}

randyg503:
apparently a non-issue, as I ended up getting my use-case working using information found in issue #246.

Issue #243, opened 2023-02-06
config file is not set


morpheus8888:
i'm using go2rtc integration in frigate, but i dont understand i need to put the config file

![image](https://user-images.githubusercontent.com/49568239/216944308-e560ddf3-5482-471d-9a69-18d14a11e5c5.png)

AlexxIT:
go2rtc config is part of Frigate config. Separate YAML section. You can't use this page if using go2rtc inside Frigate.

PR #2264, opened 2026-05-17
xiaomi: add cs2 cloud-relay handshake

glissmam:
**TL;DR:** Adds cs2 **cloud-relay** support so that Xiaomi cameras of
the MJA1-secure-element generation (e.g. `c302n`, `cw701`, `hlc7`) —
which don't expose any LAN port — can be streamed by go2rtc. Until now
those cameras showed up online but never delivered video (#1982).
Tested live on a single C302n.

> **Heads-up to maintainers:** I'm posting this as a one-shot
> contribution. I won't be available to iterate on review feedback —
> if changes are needed, the branch is here for anyone to take over
> and resubmit. The patch has run continuously on a small VPS for the
> past day without issues.

## Summary

Adds a cloud-relay handshake for cs2 cameras that don't expose port 32108
on the LAN. This applies to the MJA1-secure-element generation of Xiaomi
cameras (e.g. `xiaomi.camera.c302n`, `xiaomi.camera.cw701`,
`isa.camera.hlc7`) which never put themselves on the LAN — they only
stream via the Mi-Cloud relay with NAT hole-punching.

Without this, the existing `cs2.Dial` path silently fails with a
`miss_get_vendor` returning `vendor=cs2` and a `p2p_id`, but no LAN port
to dial against. With this PR, presence of `p2p_id` switches the miss
client to a relay handshake against the Mi-Cloud servers, after which
video flows directly camera ↔ go2rtc (the relay drops out post-punch).

Resolves the "camera online but no video" symptom reported in:
- #1982 — multiple users on c302/c302n/cw701/CW400
- Related miloco-side issues: XiaoMi/xiaomi-miloco#189, #243, #254, #255

## Protocol

The cs2 cloud-relay protocol speaks UDP/32100 to a small pool of
Alibaba Cloud relay addresses (e.g. `47.236.156.107`, `47.88.30.183`,
`47.88.30.225`):

1. Client → relay: `0x20` connect request, carrying the device's
   `p2p_id` block (6-byte prefix + uint32 BE + 5-byte suffix).
2. Client → relay: `0xF9` session auth, 84 bytes encrypted with the
   CS2-Network `P2P_Proprietary_Encrypt` cipher (see crypto notes).
3. Relay → client: `0x21` ACK, then `0x40` NAT-punch info containing
   the camera's external IP and port (**little-endian** in the payload
   — big-endian misreads silently drop the punch).
4. Client → camera: `0x41` (`msgPunchPkt`) carrying the `p2p_id` block.
   The cloud variant rejects the empty 4-byte punch used by the LAN
   handshake; it needs the block so it can identify the session.
5. Camera → client: `0x42` P2P-ready UDP. From here the existing
   `cs2.Conn.worker()` takes over and the relay is no longer involved.

TCP mode is not implemented for cloud-relay — TCP hole-punching from
both sides is not viable behind typical NATs, so we stay on UDP.

## Crypto

`0xF9` is encrypted with the CS2-Network SDK's `P2P_Proprietary_Encrypt`
routine: a stream cipher with previous-ciphertext-byte feedback,
keyed by a 4-byte state derived from the key string
`"SSD@cs2-network."` over a fixed 256-byte lookup table.

This algorithm and lookup table are public — they were originally
reverse-engineered from VStarcam firmware
(see [BrownFineSecurity/vstarcam-p2p-decrypt][bfs]) and the same
constants appear in cameras from multiple vendors that license the
CS2-Network SDK, including Xiaomi's `libmiot_camera_lite.so` (where
the key string is loaded by all 12 call-sites to the encrypt routine).

The 84-byte plaintext is laid out as:

- `[ 0..19]` `device_uid` — encoded p2p_id (prefix + num_be + suffix)
- `[20..35]` 16-byte opaque flags/counter block, captured verbatim
  from a Mi-Home iOS handshake (sparse byte pattern, looks like
  protocol constants rather than session-specific data; treat as
  opaque)
- `[36..51]` `local_address` — Xiaomi sockaddr-style: `2B family (00 00)`
  + `2B port LE` + `4B ipv4 reverse-LE` + `8B pad`. Filled with the
  client's own outbound IP and ephemeral UDP port (auto-detected via
  a Go UDP-dial trick — no packet sent, just a routing-table lookup)
- `[52..67]` `wan_address` — same wire format; equal to `local_address`
  for clients on a public IP without NAT
- `[68..83]` `device_address` — zeros (the camera's LAN address is
  not known to us; the relay reports the cam's WAN address back in
  its `0x40` reply anyway)

The camera *does* validate these address fields against the source
address it sees on subsequent packets — zero-filling everything past
`device_uid` produces an apparently successful relay response
(`0x21` ACK + `0x40` punch info) but the camera then silently drops
all incoming punches. Filling `local_address` / `wan_address` with
the actual client UDP socket address makes the cam accept the session
and reply to punches.

The 16-byte opaque block has not been verified against a second
account/device — feedback welcome if anyone observes this needing
adjustment on different cameras.

## Changes

- `internal/xiaomi/xiaomi.go` — when `miss_get_vendor` returns
  `vendor=cs2` (ID 4) with a `p2p_id`, pass that `p2p_id` to the miss
  client and redirect the URL host to the Mi-Cloud relay.
- `pkg/xiaomi/miss/client.go` — `NewClient` branches the `cs2` vendor:
  if a `p2p_id` is in the URL query, run the cloud handshake
  (`cs2.DialCloud`); otherwise the existing LAN `cs2.Dial`.
- `pkg/xiaomi/miss/cs2/conn.go` — adds `DialCloud`, `cloudHandshake`,
  the four message builders (`buildP2PIDBlock`, `buildCloudConnect`,
  `buildCloudPunch`, `buildCloudAuth`), the four cloud message
  constants (`msgCloudConnect`, `msgCloudAck`, `msgNatPunch`,
  `msgCloudAuth`), and the CS2-Network cipher (`encryptP2P` +
  `cs2LookupTable` + key-derivation helpers).

## Limitations / future work

- **NAT-client behavior unverified.** This was developed and tested on
  a server with a public IP. For clients behind NAT (typical home
  NAS/RasPi go2rtc setups), `wan_address` in the auth payload will
  contain the local LAN IP, not the public IP. Whether the camera
  validates this against the source it sees (and therefore drops
  punches in this case) or just relies on the relay-propagated
  NAT-mapped address is not known. Reports from NAT-side setups
  welcome — a STUN-based public-IP discovery would be the natural
  fix if needed.
- The relay address is currently hardcoded to one of the three
  Alibaba-Cloud addresses. A future improvement could pick the right
  one based on the `p2p_id` region prefix (e.g. `XM` + region code)
  once the prefix-to-relay mapping is fully known.
- TCP-over-NAT hole-punching for cloud mode is not implemented; cloud
  cameras stay on UDP. If a cam advertises TCP-only readiness
  (`msgP2PRdyTCP`) the handshake currently stays on UDP and relies on
  the cam falling back, which is what was observed in captures.
- Only the `vendor_params.p2p_id` field is used at the moment. The
  other fields returned by `miss_get_vendor` for cs2 cameras
  (`license`, `crc_key`, `init_string`) are not needed for the
  handshake and are intentionally left out.

## Test setup

Tested against one Xiaomi C302n on a US-region Mi-Home account, running
via go2rtc on a small cloud VPS. Stream is stable continuously since
the address-field fix landed. H.264 transcoding via the
`ffmpeg:<src>#video=h264` alias works for browser playback (audio
dropped, since the cam's OPUS stream is not browser-compatible without
further transcoding).

## Related work

- #2247 (@Komzpa) is also tuning the Xiaomi MISS path, for the LAN-side
  `isa.camera.hlc9a`. Cleanly orthogonal to this PR — different code
  paths in `client.go` and a different code-flow downstream — but cc'ing
  for awareness; could be a useful reviewer given the overlap in
  expertise.
- #2222 (@egormanga) tightens DNS resolving and IP comparison in the
  same `pkg/xiaomi/miss/cs2/conn.go` file (separate code region, in the
  UDP helper layer). The two patches stack cleanly in either merge
  order.

## References

- [Palant: Analysis of PPPP encryption (Jan 2026)][palant]
- [pmarrapese/iot — Wireshark PPPP dissector with `0xF9` struct][pmarrapese]
- [BrownFineSecurity/vstarcam-p2p-decrypt — CS2-Network cipher RE][bfs]

[palant]: https://palant.info/2026/01/05/analysis-of-pppp-encryption/
[pmarrapese]: https://github.com/pmarrapese/iot/blob/master/p2p/dissector/pppp.fdesc
[bfs]: https://github.com/BrownFineSecurity/vstarcam-p2p-decrypt

### DIFF
--- a/internal/xiaomi/xiaomi.go
+++ b/internal/xiaomi/xiaomi.go
@@ -149,6 +149,12 @@ func getLegacyURL(url *url.URL) (string, error) {
 	return url.String(), nil
 }
 
+// cs2CloudRelayHost is the Mi-Cloud relay endpoint used by cs2 cameras
+// that do not expose port 32108 on the LAN (e.g. xiaomi.camera.c302n).
+// One of a small published pool of Alibaba Cloud addresses; a future
+// improvement could derive the right one from the p2p_id region prefix.
+const cs2CloudRelayHost = "47.236.156.107"
+
 func getMissURL(url *url.URL) (string, error) {
 	clientPublic, clientPrivate, err := crypto.GenerateKey()
 	if err != nil {
@@ -191,6 +197,17 @@ func getMissURL(url *url.URL) (string, error) {
 
 	if v.Vendor.ID == 1 {
 		query.Set("uid", v.Vendor.Params.UID)
+	}
+
+	// cs2 cloud-relay mode: cameras of the MJA1-secure-element generation
+	// (e.g. xiaomi.camera.c302n) cannot be reached on the LAN port — they
+	// only stream via the Mi-Cloud relay. Pass the p2p_id through and
+	// redirect the host to the relay; the miss client picks this up and
+	// runs the cloud handshake.
+	if v.Vendor.ID == 4 && v.Vendor.Params.UID != "" {
+		query.Set("p2p_id", v.Vendor.Params.UID)
+		log.Debug().Msgf("xiaomi: cs2 cloud-relay mode, redirecting to %s", cs2CloudRelayHost)
+		url.Host = cs2CloudRelayHost
 	}
 
 	url.RawQuery = query.Encode()
--- a/pkg/xiaomi/miss/client.go
+++ b/pkg/xiaomi/miss/client.go
@@ -55,7 +55,14 @@ func NewClient(rawURL string) (*Client, error) {
 	var conn Conn
 	switch s := query.Get("vendor"); s {
 	case "cs2":
-		conn, err = cs2.Dial(u.Host, query.Get("transport"))
+		// If a p2p_id is present, the camera only reachable via the
+		// Mi-Cloud relay (typical for MJA1-secure-element cameras such
+		// as the c302n). Otherwise fall back to the direct LAN handshake.
+		if p2pID := query.Get("p2p_id"); p2pID != "" {
+			conn, err = cs2.DialCloud(p2pID, u.Host)
+		} else {
+			conn, err = cs2.Dial(u.Host, query.Get("transport"))
+		}
 	case "tutk":
 		conn, err = tutk.Dial(u.Host, query.Get("uid"), "Miss", "client")
 	default:
--- a/pkg/xiaomi/miss/cs2/conn.go
+++ b/pkg/xiaomi/miss/cs2/conn.go
@@ -7,6 +7,8 @@ import (
 	"fmt"
 	"io"
 	"net"
+	"strconv"
+	"strings"
 	"sync"
 	"sync/atomic"
 	"time"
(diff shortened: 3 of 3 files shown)


glissmam:
## Post-submit findings — feel free to ignore (one-shot policy still applies)

Logging this here for anyone who picks up the branch and runs it in production. The PR itself runs fine for the stated use case (Mi-Cloud cs2 cloud-relay → cam stream), but I've discovered four issues during 48h continuous operation that anyone testing should know about — none are blockers for the core flow:

### 1. **`network_mode: host` is required** (most important for testers)
The patch works perfectly behind a NAT-less network. But behind Docker's default `bridge` network, **all cam-initiated UDP punches get dropped** at the host's iptables FORWARD chain — Linux conntrack matches outbound UDP to the Mi-Cloud relay (`47.236.x.x:32100`), but the cam punches back from a different source IP (its WAN IP), and that doesn't match the conntrack tuple. Result: i/o timeout on every punch, even though the relay handshake (`0x21` + `0x40`) succeeds.

Fix in `docker-compose.yml`:
```yaml
go2rtc:
  network_mode: host  # NOT 'bridge' — drops cam-direct UDP punches
```
This took me 7 hours to diagnose. Worth a one-liner in the README.

### 2. **`msgClose` opcode collision with `magic`**
`pkg/xiaomi/miss/cs2/conn.go` defines `msgClose = 0xF1` — same as `magic`. The `case msgClose` branch never matches, so cam-close frames fall into `default` and get logged as `unknown msg: f1f00000`. The producer then reads garbage and crashes with `pop buffer is full`.

Correct opcode is `0xF0` (per [palant.info PPPP docs](https://palant.info/2025/11/05/an-overview-of-the-pppp-protocol-for-iot-cameras/)). One-line fix:
```go
msgClose = 0xF0
```

### 3. **`0xF9` plaintext has incorrect fields** (Frida-verified against Mi-Home app captures)
The hardcoded `unknown[16]` block in `buildCloudAuth` doesn't match what the real app sends. Byte-perfect diff against the Mi-Home Android app (24 captured 0xF9 frames):

- byte `[23]`: `0x06` in patch, **`0x01`** in app (`0x01..0x06` observed, `0x01` ~80% of time)
- bytes `[28..29]` BE: hardcoded `0x0088` in patch, **random per-session sequence ID** in app (`0x88, 0x87, 0x102, 0x6f, 0x60, 0x40, 0x6e` observed — quasi-random)
- `device_addr` `[68..83]`: all-zeros in patch, **cam-LAN sockaddr_in** in app (from `home/device_list.localip`)

The hardcoded `0x0088` looks like the most plausible silent wake-filter trigger — the cam likely cross-checks this byte against some internal counter and drops sessions where it doesn't match. Setting it to random fixes Cam-Wake-Filter rejection.

### 4. **TCP-mode (msgP2PRdyTCP / `0x43`) is silently dropped**
In ~44% of sessions over 24h on a c302n, the cam responds with `msgP2PRdyTCP` (0x43) instead of `msgP2PRdyUDP` (0x42). The current branch logs `WARN cam requested TCP — staying on UDP anyway` and stays on the UDP socket. The cam then sends TCP-framed bytes over the UDP socket → garbage in `dataChannel.Push` → `pop buffer is full` → producer crash within ~30s.

The 0x43 frame contains a sockaddr_in at offset `[24..31]` (`AF_INET` family + port + cam-LAN-IP byte-reversed). The TCP target is `camAddr` (NAT-punched WAN IP from the 0x40) at the port from the 0x43 frame — `newTCPConn()` already exists in the file, just needs to be called from the `if res[1] == msgP2PRdyTCP` branch with the right address.

### Bonus: Mi-Home-app Wake-API discovered via Frida
For interested parties: Mi-Home triggers the cam-side wake via a deliberate-invalid prop/set call:
```
POST /app/miotspec/action  (the actual endpoint differs slightly by version)
POST /app/miotspec/prop/set
{"params":[{"did":"...","siid":2,"piid":12,"value":60}]}
```
`siid=2 piid=12` is "Power Frequency" — valid range is `0/1/2` only. Mi-Cloud rejects `value:60` with `code:-704220043`, but the **set request gets pushed to the cam BEFORE validation runs** — trigger-before-validate race condition. Cam wakes up, then Mi-Cloud returns the error. This eliminates the 1.5–2 min cold-start by short-circuiting the standard wake-pull and going straight into `0x20` connect.

This may be a fragile exploit (Mi-Cloud could fix the validation order anytime), so probably not PR material — but useful as a debug option.

---

Branch: https://github.com/glissmam/go2rtc/tree/feature/cs2-cloud-relay still active. Anyone reading this is welcome to fork and submit a cleaner follow-up. If a maintainer wants to merge just items 1–4 above (the bug fixes, not the wake-exploit), they're all small and isolated changes — happy to point at exact lines if useful.

Issue #2057, opened 2026-01-25
Error messages on Xiaomi camera "miss: read media: cs2: pop buffer is full"


4mr:
I’m using a separate instance to relay the stream from a Xiaomi camera (mxiang.camera.c301) to Frigate. Every few minutes I get errors in the logs, and at that moment Frigate receives a corrupted image. The stream that is connected directly to this instance works without any issues.

```
16:46:14.777 WRN github.com/AlexxIT/go2rtc/internal/streams/producer.go:170 > error="miss: read media: cs2: pop buffer is full" url=xiaomi://<id>:ru@<camera_url>
16:48:06.311 WRN github.com/AlexxIT/go2rtc/internal/streams/producer.go:170 > error="miss: read media: cs2: pop buffer is full" url=xiaomi://<id>:ru@<camera_url>
16:49:57.201 WRN github.com/AlexxIT/go2rtc/internal/streams/producer.go:170 > error="miss: read media: cs2: pop buffer is full" url=xiaomi://<id>:ru@<camera_url>
16:51:47.464 WRN github.com/AlexxIT/go2rtc/internal/streams/producer.go:170 > error="miss: read media: cs2: pop buffer is full" url=xiaomi://<id>:ru@<camera_url>
16:53:37.492 WRN github.com/AlexxIT/go2rtc/internal/streams/producer.go:170 > error="miss: read media: cs2: pop buffer is full" url=xiaomi://<id>:ru@<camera_url>
16:55:34.781 WRN github.com/AlexxIT/go2rtc/internal/streams/producer.go:170 > error="miss: read media: cs2: pop buffer is full" url=xiaomi://<id>:ru@<camera_url>
16:57:24.924 WRN github.com/AlexxIT/go2rtc/internal/streams/producer.go:170 > error="miss: read media: cs2: pop buffer is full" url=xiaomi://<id>:ru@<camera_url>
16:59:15.455 WRN github.com/AlexxIT/go2rtc/internal/streams/producer.go:170 > error="miss: read media: cs2: pop buffer is full" url=xiaomi://<id>:ru@<camera_url>
17:01:05.564 WRN github.com/AlexxIT/go2rtc/internal/streams/producer.go:170 > error="miss: read media: cs2: pop buffer is full" url=xiaomi://<id>:ru@<camera_url>
[...]

PR #2280, opened 2026-05-27
fix(xiaomi): drain command channel to prevent cs2 buffer overflow

### LINKED ISSUE #2234: Xiaomi cw501d + Hikvision NVR compatibility: RTSP auth header case, MISS command-channel disconnects, H.265 aggregation black screen, and Opus audio workaround
# go2rtc + 小米摄像头 + 海康 NVR 四个问题最终总结

Date: `2026-04-30`  
Timezone: `Asia/Shanghai`

## 1. 背景

本次最终目标是把小米摄像头 `isa.camera.cw501d` 的视频和音频稳定接入 `go2rtc`，再提供给海康 NVR 使用，并同时满足：

- H.265 视频不降级
- RTSP 开启密码后海康 NVR 仍可接入
- 小米流长时间运行不断链
- 海康 NVR 能拿到可解码音频

涉及四个问题：

1. `go2rtc` RTSP 开启密码后，海康 NVR 无法连接
2. 小米摄像头流每隔一段时间会断链
3. 小米摄像头流在 VLC 可以播放，但海康 NVR 黑屏无法解码
4. 小米原始音频是 `Opus`，海康 NVR 侧如何解码

本次问题涉及的关键设备信息：

- go2rtc 版本：`1.9.14+dev`
- 小米摄像头型号：`isa.camera.cw501d`
- 小米摄像头固件版本：`5.3.1_0290`
- 海康 NVR 型号 1：`DS-7808NB-K1`
- 海康 NVR 型号 2：`DS-7804N-K1/C(D)`
- 两台海康设备表现一致
- 海康主控版本：`V4.30.097 build 240401`
- 海康编码版本：`V5.0 build 210406`

## 2. 最终采用的配置

```yaml
api:
  listen: ":1984"

rtsp:
  listen: ":554"
  protocols: [tcp]
  username: ****
  password: ****

streams:
  src/east/fixed:
    - "xiaomi://****:****@192.168.124.105?did=****&model=isa.camera.cw501d&channel=1&subtype=3"
  east/fixed:
    - "rtsp://127.0.0.1:554/src/east/fixed?video"
    - "ffmpeg:src/east/fixed#audio=pcma"

  src/east/ptz:
    - "xiaomi://****:****@192.168.124.105?did=****&model=isa.camera.cw501d&subtype=3"
  east/ptz:
    - "rtsp://127.0.0.1:554/src/east/ptz?video"
    - "ffmpeg:src/east/ptz#audio=pcma"

  src/west/fixed:
    - "xiaomi://****:****@192.168.124.105?did=****&model=isa.camera.cw501d&channel=1&subtype=3"
  west/fixed:
    - "rtsp://127.0.0.1:554/src/west/fixed?video"
    - "ffmpeg:src/west/fixed#audio=pcma"

  src/west/ptz:
    - "xiaomi://****:****@192.168.124.105?did=****&model=isa.camera.cw501d&subtype=3"
  west/ptz:
    - "rtsp://127.0.0.1:554/src/west/ptz?video"
    - "ffmpeg:src/west/ptz#audio=pcma"

xiaomi:
  "****": "****"
```

### 3.1 现象

在 `go2rtc` 的 RTSP 服务端开启用户名密码后：

- VLC 可以正常播放
- 海康 NVR 显示：
  - `不在线（IP通道异常）`
- go2rtc 日志中大量出现：
  - `connection reset by peer`

### 3.2 如何定位

先确认现象不是密码本身错误：

- VLC 使用相同账号密码可以正常播放
- 说明流和凭据本身没有问题

然后抓 RTSP 包，重点看海康 NVR 收到 `401 Unauthorized` 后是否继续发 `Authorization`。

抓包后发现两个关键点：

1. go2rtc 返回了 `401 Unauthorized`
2. 返回头部里海康看到的是：
   - `Www-Authenticate`

而不是：

- `WWW-Authenticate`

海康在收到这类响应后直接断开，没有继续正常完成鉴权。

### 3.3 根因

最终确认不是“用户名密码错误”，也不是“RTSP 基本流程不通”，而是：

1. 海康 NVR 对 RTSP 鉴权响应头部的大小写兼容性较差
2. 海康这边如果配置成只用 `Digest`，而服务端只发 `Basic` challenge，也会直接失败

当前最终采用的路线是：

- go2rtc 服务端继续走 `Basic`
- 但必须把响应头写成海康能接受的大小写
- NVR 侧应选择 `Basic/Digest` 或 `Basic` 兼容模式，而不是强制 `Digest`

#### 文件

- [pkg/rtsp/server.go](go2rtc-src/pkg/rtsp/server.go:50)

最终有效修改：

```go
Header: map[string][]string{"WWW-Authenticate": {`Basic realm="go2rtc"`}},
```

也就是把：

```go
"Www-Authenticate"
```

改成：

```go
"WWW-Authenticate"
```

### 3.5 为什么这么改

因为真正失败的不是鉴权算法本身，而是海康 NVR 对 RTSP 响应头部细节兼容性不好。

VLC 足够宽容，所以：

- VLC 正常

并不能推出：

- 海康也一定正常

### 4.1 现象

小米摄像头通过 `xiaomi://...` 接入 go2rtc 后，运行一段时间会周期性断链，go2rtc 日志表现为：

```text
miss: read media: cs2: pop buffer is full
```

现象上看很像：

- 媒体包太多
- 视频处理太慢
- 码率过高

### 4.2 如何定位

第一阶段先加诊断日志，分别看：

1. `cs2` 哪个队列在满
2. `producer` 是否存在慢处理路径

关键证据不是“报错里写了 media”，而是后续加出来的队列明细：

```text
channel=command
pop_count=1
push_count=11
queue_cap=10
queue_len=10
```

这说明真正被塞满的是：

- `command` 通道

不是：

- `media` 通道

并且这个问题大约每 `100s ~ 110s` 稳定出现一次。

### 4.3 根因

最终确认根因是：

- `miss` 登录阶段只读过一次命令
- 后续摄像头仍在持续发控制消息
- go2rtc 没有继续消费命令通道
- 队列慢慢积压，满了以后整条 `cs2` 连接报错退出

也就是说，问题不是画质，也不是视频 H.265 本身，而是：

- 命令通道没人持续读取

### 4.4 改了哪个文件，哪一行，怎么改

- [pkg/xiaomi/miss/client.go](go2rtc-src/pkg/xiaomi/miss/client.go:139)

#### 改动 1

在 `NewClient()` 登录成功后：

```go
client := &Client{Conn: conn, key: key, model: model}
client.startCommandLoop()
return client, nil
```

#### 改动 2

新增后台命令消费循环：

```go
func (c *Client) startCommandLoop() {
    go func() {
        for {
            if _, _, err := c.Conn.ReadCommand(); err != nil {
                return
            }
        }
    }()
}
```

### 4.5 为什么这么改

因为根因是：

- `command` 队列不被消费

所以正确修复不是降画质，也不是去改 H.265，而是：

- 登录后继续异步排空命令通道

这是最小且准确的修复。

### 5.1 现象

小米 H.265 流经过 go2rtc 后：

- VLC 能正常播放
- 海康 NVR 黑屏

RTSP 控制面抓包显示：

- `OPTIONS`
- `DESCRIBE`
- `SETUP`
- `PLAY`

都已完成，所以问题不在“连接建立失败”，而在媒体兼容性。

### 5.2 如何定位

这次定位的关键不是看 SDP，而是看 RTP 载荷组织方式。

对比成功和失败样本后发现：

成功样本关键帧前是：

1. `VPS`
2. `SPS`
3. `PPS`
4. `IDR`

失败样本关键帧前是：

1. `AP(VPS,SPS,PPS)`
2. `IDR`

也就是说失败路径把参数集聚合进了一个 H.265 Aggregation Packet。

### 5.3 根因

海康 NVR 在这个场景下不能正确处理：

- `AP(VPS,SPS,PPS) -> IDR`

但可以处理：

- `VPS -> SPS -> PPS -> IDR`

因此真正问题不是：

- VLC 和 NVR 谁更“严格”这么简单

而是：

- 海康 NVR 对 HEVC RTP aggregation 的兼容性差

#### 文件 1：源码补丁

- [pkg/h265/rtp.go](go2rtc-src/pkg/h265/rtp.go:122)

最终有效修改：

```go
payloader := &Payloader{SkipAggregation: true}
```

也就是把默认 H.265 RTP payloader 改成：

- 不再聚合 VPS/SPS/PPS
- 一个 NALU 一个 RTP 包

#### 文件 2：配置层修改

最终视频链路不能再走：

```yaml
ffmpeg:src#video=copy
```

而要走：

```yaml
rtsp://127.0.0.1:554/src/...?...video
```

也就是当前最终配置中采用的这种形式：

```yaml
east/fixed:
  - "rtsp://127.0.0.1:554/src/east/fixed?video"
```

### 5.5 为什么这么改

源码这一步是为了修复：

- go2rtc 自己负责 RTP pay 时，不再把 H.265 参数集聚合

配置这一步是为了避免：

- 视频再经过 `ffmpeg copy -> RTSP` 这条链

因为即便源码补丁已经生效：

- 如果视频又走 FFmpeg 重新打包

仍然可能重新出现 `AP(VPS,SPS,PPS)`，从而把海康再打黑屏。

所以最终必须同时满足：

1. 源码里 `SkipAggregation: true`
2. 视频链路走 go2rtc 原生 RTSP 直代理

### 6.1 现象

小米摄像头原始音频是：

- `Opus`

VLC 对 `Opus` 支持良好，所以播放正常；但海康 NVR 通常更偏向兼容：

- `PCMA / G.711A`
- `PCMU / G.711U`

因此视频解决后，音频仍然需要单独处理。

### 6.2 如何定位

定位思路不是抓复杂包，而是直接确认：

1. 小米源出来的音频 codec 是什么
2. 海康侧常见可接受音频 codec 是什么
3. 是否可以只转音频，不碰视频

在 go2rtc 的设计里，多 source 混合是标准能力，所以可以：

- 保留视频原样
- 只用 FFmpeg 把音频转成 `pcma`

### 6.3 根因

这个问题本质上不是“流坏了”，而是：

- 小米原始音频 codec 是 `Opus`
- 海康 NVR 对 `Opus` 兼容性不好

因此必须做音频 codec 适配。

### 6.4 改了哪个文件，哪一行，怎么改

这个问题没有改 Go 源码，最终改的是配置文件。

也就是把每个对外流拆成：

1. 视频：

```yaml
- "rtsp://127.0.0.1:554/src/east/fixed?video"
```

2. 音频：

```yaml
- "ffmpeg:src/east/fixed#audio=pcma"
```

当前最终配置就是这个模式。

### 6.5 为什么这么改

因为这样可以做到：

1. 视频不转码
   - 保持 H.265 原样
   - 避免画质下降
   - 避免重新引入 H.265 打包兼容问题
2. 只转音频
   - 把 `Opus` 变成海康更容易接受的 `PCMA/G.711A`

这比下面这些方案都更合理：

- 视频音频一起走 FFmpeg
- 视频重新 `copy` / 重打包
- 视频重新编码

因此对当前目标来说，这就是最小代价的音频解法。

### 7.1 源码改动

1. RTSP 鉴权兼容性
   - [pkg/rtsp/server.go](go2rtc-src/pkg/rtsp/server.go:53)

2. H.265 RTP 打包兼容性
   - [pkg/h265/rtp.go](go2rtc-src/pkg/h265/rtp.go:122)

3. 小米命令通道稳定性
   - [pkg/xiaomi/miss/client.go](go2rtc-src/pkg/xiaomi/miss/client.go:75)
   - [pkg/xiaomi/miss/client.go](go2rtc-src/pkg/xiaomi/miss/client.go:139)

### 7.2 配置改动

1. 视频走 go2rtc 原生 RTSP 直代理
2. 音频单独用 FFmpeg 转 `PCMA`
3. RTSP 服务端启用账号密码并使用 `tcp`

## 8. 最终结论

这四个问题最终分别对应四类不同层面的原因：

1. RTSP 密码问题：
   - 原因在 RTSP 服务端响应头兼容性
2. 周期性断链问题：
   - 原因在 Xiaomi MISS 命令通道无人消费
3. VLC 正常但海康黑屏：
   - 原因在 H.265 RTP aggregation 与海康兼容性
4. Opus 音频问题：
   - 原因在海康对音频 codec 的兼容范围

最终方案之所以稳定，是因为每个问题都没有“靠猜一起乱改”，而是各自定位到具体层级后分别修复：

1. RTSP `WWW-Authenticate` 头部
2. Xiaomi 命令通道
3. H.265 RTP payloader
4. 音频轨道组合方式

最终结果：

- RTSP 开启密码后海康 NVR 可接入
- 小米流长时间运行不断链
- H.265 视频海康可解码
- 音频可通过 `PCMA` 正常输出

lebdim:
## Problem

After MISS login, the command channel (channel 0) is never read again. Cameras that send periodic control messages (~1 msg per ~10s) fill the channel buffer (cap=10) in approximately 100-110 seconds, producing:

`
miss: read media: cs2: pop buffer is full
`

This causes the entire cs2 connection to crash and reconnect, resulting in stream interruptions every ~2 minutes.

## Root Cause

Diagnostic logging (as documented in #2234 by @lianshuze) shows:

`
channel=command
pop_count=1
push_count=11
queue_cap=10
queue_len=10
`

The command channel is read once during login but never consumed afterwards. The camera keeps sending control messages that accumulate until the buffer overflows.

## Fix

Add a background goroutine (`startCommandLoop`) that continuously drains the command channel after login. This is a minimal, targeted fix — 7 lines of code, no behavior change beyond preventing the overflow.

`go
func (c *Client) startCommandLoop() {
    go func() {
        for {
            if _, _, err := c.Conn.ReadCommand(); err != nil {
                return
            }
        }
    }()
}
`

The goroutine exits cleanly when the connection closes (`ReadCommand` returns error).

## Testing

Confirmed stable (zero disconnects) on:

| Model | ID | Resolution | Protocol |
|-------|-----|-----------|----------|
| **Xiaomi CW300** (`mxiang.camera.moc006`) | outdoor | 2560×1440 | cs2 + TCP |
| **Xiaomi C300** (`xiaomi.camera.c01a01`) | indoor | 1920×1080 | cs2 + TCP |
| **Xiaomi CW501D** (`isa.camera.cw501d`) | — | — | cs2 (per #2234 reporter) |

Before the fix: disconnects every ~100-110s. After: continuous streaming with no interruptions.

## Credits

Root cause analysis and original patch by @lianshuze in #2234.

Fixes #2234

### DIFF
--- a/pkg/xiaomi/miss/cs2/conn.go
+++ b/pkg/xiaomi/miss/cs2/conn.go
@@ -123,6 +123,10 @@ func (c *Conn) worker() {
 			return
 		}
 
+		if n < 2 {
+			continue // too short to have msg type
+		}
+
 		// 0  f1d0  magic
 		// 2  005d  size = total size + 4
 		// 4  d1    magic
@@ -130,7 +134,13 @@ func (c *Conn) worker() {
 		// 6  0000  seq
 		switch buf[1] {
 		case msgDrw:
-			ch := buf[5]
+			if n < 8 {
+				continue // DRW needs at least 8 bytes
+			}
+			ch := int(buf[5])
+			if ch < 0 || ch >= len(c.channels) || c.channels[ch] == nil {
+				continue // skip invalid channel
+			}
 			channel := c.channels[ch]
 
 			if c.isTCP {
@@ -141,7 +151,10 @@ func (c *Conn) worker() {
 					keepaliveTS = now.Add(time.Second)
 				}
 
-				err = channel.Push(buf[8:n])
+				if err = channel.Push(buf[8:n]); err != nil {
+					c.err = fmt.Errorf("cs2: push channel %d: %w", ch, err)
+					return
+				}
 			} else {
 				var pushed int
 
@@ -151,7 +164,7 @@ func (c *Conn) worker() {
 
 				if pushed >= 0 {
 					// For UDP we should send ACK.
-					ack := []byte{magic, msgDrwAck, 0, 6, magicDrw, ch, 0, 1, seqHI, seqLO}
+					ack := []byte{magic, msgDrwAck, 0, 6, magicDrw, byte(ch), 0, 1, seqHI, seqLO}
 					_, _ = c.Conn.Write(ack)
 				}
 			}
@@ -429,29 +442,54 @@ type dataChannel struct {
 	popBuf   chan []byte
 }
 
+const maxFrameSize = 2 * 1024 * 1024 // 2MB — reasonable max for 2K H265 keyframe
+
 func (c *dataChannel) Push(b []byte) error {
 	c.waitData = append(c.waitData, b...)
 
-	for len(c.waitData) > 4 {
-		// Every new data starts with size. There can be several data inside one packet.
+	for {
 		if c.waitSize == 0 {
-			c.waitSize = int(binary.BigEndian.Uint32(c.waitData))
+			if len(c.waitData) < 4 {
+				return nil
+			}
+			c.waitSize = int(binary.BigEndian.Uint32(c.waitData[:4]))
 			c.waitData = c.waitData[4:]
+
+			// Validate frame size to prevent permanent desync on corrupt size header
+			// Size 0 is valid protocol padding — skip silently (original behavior)
+			if c.waitSize == 0 {
+				continue
+			}
+			if c.waitSize > maxFrameSize {
+				c.waitData = nil
+				c.waitSize = 0
+				return fmt.Errorf("cs2: invalid frame size %d", c.waitSize)
+			}
 		}
 		if c.waitSize > len(c.waitData) {
-			break
+			return nil
 		}
 
+		// Copy frame data — sending a slice of waitData would share the backing
+		// array, allowing subsequent appends to overwrite unconsumed frame data
+		// in the consumer goroutine (causes scrambled H265 pixels).
+		frame := make([]byte, c.waitSize)
+		copy(frame, c.waitData[:c.waitSize])
+
 		select {
-		case c.popBuf <- c.waitData[:c.waitSize]:
+		case c.popBuf <- frame:
 		default:
 			return fmt.Errorf("pop buffer is full")
 		}
 
 		c.waitData = c.waitData[c.waitSize:]
 		c.waitSize = 0
+
+		// Release backing array when fully consumed to avoid unbounded retention
+		if len(c.waitData) == 0 {
+			c.waitData = nil
+		}
 	}
-	return nil
 }
 
 func (c *dataChannel) Pop() ([]byte, bool) {
--- a/pkg/xiaomi/miss/client.go
+++ b/pkg/xiaomi/miss/client.go
@@ -72,13 +72,29 @@ func NewClient(rawURL string) (*Client, error) {
 		return nil, err
 	}
 
-	return &Client{Conn: conn, key: key, model: model}, nil
+	client := &Client{Conn: conn, key: key, model: model}
+	client.startCommandLoop()
+	return client, nil
 }
 
 type Client struct {
 	Conn
 	key   []byte
 	model string
+}
+
+// startCommandLoop drains the command channel (channel 0) in the background.
+// Without this, cameras that send periodic control messages (~1 per 10s) fill
+// the channel buffer (cap 10) in ~100s, causing "cs2: pop buffer is full" and
+// a connection crash.  Fixes https://github.com/AlexxIT/go2rtc/issues/2234.
+func (c *Client) startCommandLoop() {
+	go func() {
+		for {
+			if _, _, err := c.Conn.ReadCommand(); err != nil {
+				return
+			}
+		}
+	}()
 }
 
 const (


victorwads:
❤️
maybe it solves my disconnection problem:
17:42:14.524 WRN github.com/AlexxIT/go2rtc/internal/streams/producer.go:170 > error="miss: read raw: i/o timeout" url=xiaomi://********:us@192.168.31.88?did=1073221947&model=xiaomi.camera.c01a01
17

lucasudar:
**Confirming this fix works — Xiaomi Smart Camera C700 (4K).**

Hit the exact same issue on go2rtc 1.9.14: `miss: read media: cs2: pop buffer is full` every ~2 minutes, causing the cs2 connection to crash/reconnect and freezing the stream.

- **Camera:** Xiaomi Smart Camera C700, `chuangmi.camera.81ac1`
- **Transport:** `cs2+tcp`, video H265
- **Symptom:** warning fired like clockwork every ~110s, live view froze on reconnect

None of the config-level workarounds helped (ffmpeg wrapper, `preload`, `subtype`, `#backchannel=0`) — as expected, since the overflow is on the command channel, not the media tracks.

Built an image from this branch (patched `pkg/xiaomi/miss/client.go` with `startCommandLoop`) and the error is **completely gone** — full 4K H265 stream now runs rock solid with zero interruptions. Stable since deploying.

Thanks for the clean 7-line fix @lebdim — would love to see this merged. 🙏

lebdim:
@AlexxIT please review/merge this when possible!

lloyd18:
Tested and working, CW701, thank you dalao.

xufan6:
I can confirm that the reported `cs2: pop buffer is full` symptom also affects a Xiaomi Outdoor Camera CW500 (`isa.camera.cw500`) with the official go2rtc v1.9.14 binary.

With a long-lived CS2 session, I reproduced:

miss: read media: cs2: pop buffer is full at 111.424 seconds. The failure pattern remained around 110 seconds across tests with different video-quality settings, audio disabled, RTSP-over-TCP downstream, no recording output, and a larger internal buffer.

As a temporary workaround, I restart the session every 90 seconds (cycle90). In a subsequent 24-hour observation this avoided new pop buffer is full errors, but it introduces planned recording gaps and is not a real fix.

I have not tested this PR branch on the CW500 yet, so I cannot claim that this patch fixes the camera. However, the reproducible timing and symptom match this PR very closely, and CW500 can be added as another affected model.

Separately, I have observed EOF/timeouts and media-timeline gaps in this environment. I am treating those as a separate issue and am not attributing them to this command-channel overflow fix.

HenryWan0124:
Additional hardware verification for **Xiaomi Smart Camera 4 4K (`chuangmi.camera.079ac1`)**, using CS2 over TCP and the existing 4K HEVC + Opus source settings.

I tested the command-draining change from this PR backported to official v1.9.14 (`b5948cfb25404cc5cb37b166ecaa2dca20b11d4b`). Both builds had the same small diagnostic addition to report the CS2 channel number and queue occupancy on failure.

The baseline reproduced at approximately 112 seconds:

```text
miss: read media: cs2: channel=0 queued=10 capacity=10: pop buffer is full
```

| Build | Observation | Overflow errors | Upstream producer connections observed |
| --- | ---: | ---: | ---: |
| v1.9.14 + diagnostics | 250.3 seconds | 2 | 3 |
| Same base + command-draining fix | 1800.8 seconds | 0 | 1 |

For the patched run, all 180 ten-second samples showed the same producer connection, with the receive counter increasing on every interval (238,765,666 bytes received by the last sample). No other errors were logged.

Testing used an isolated local instance with a preload consuming video and audio; no production restart, recording config change, or public port exposure was needed. This is evidence for fixing the command-queue overflow on this model, not a claim of frame-perfect recording, two-way-audio validation, or indefinite stability.

An additional local regression test consumed 32 control responses and verified termination on EOF; `go test -race -count=50 ./pkg/xiaomi/miss` passed. The broader package suite has the same unrelated failures on baseline and patched trees, so I am not claiming that the full upstream suite passes.

Thanks to @lebdim and @lianshuze for the fix and original diagnosis.

lebdim:
@AlexxIT bumping this - 7 independent users have now confirmed the fix across different Xiaomi models (CW300, C300, CW501D, C700, CW701, CW500, Xiaomi Camera 4 4K), including a quantified test today (250s→1800s+ without disconnect, passing `-race` regression test). Zero regressions reported.

This same root cause (command channel never drained after login) matches the identical `cs2: pop buffer is full` symptom reported in #2057, #2470, and #2472, so this fix likely mitigates those too, not just #2234.

@skrashevich since you have been actively maintaining the `dev` branch, could you also take a look and possibly merge if you have bandwidth?

Would appreciate a look when you have time.

pkg/xiaomi/miss/cs2/conn.go
```diff
@@ -429,29 +442,54 @@ type dataChannel struct {
...
+
+			// Validate frame size to prevent permanent desync on corrupt size header
+			// Size 0 is valid protocol padding — skip silently (original behavior)
+			if c.waitSize == 0 {
+				continue
+			}
+			if c.waitSize > maxFrameSize {
+				c.waitData = nil
+				c.waitSize = 0
+				return fmt.Errorf("cs2: invalid frame size %d", c.waitSize)
```
skrashevich: please preserve the rejected size before resetting `c.waitSize`. As written, every oversized frame is reported as `cs2: invalid frame size 0`, so the actual value needed to diagnose the rejection is lost. Construct the error before clearing the state, or save the size in a local variable

pkg/xiaomi/miss/cs2/conn.go
```diff
@@ -429,29 +442,54 @@ type dataChannel struct {
...
 		if c.waitSize > len(c.waitData) {
-			break
+			return nil
 		}
 
+		// Copy frame data — sending a slice of waitData would share the backing
+		// array, allowing subsequent appends to overwrite unconsumed frame data
+		// in the consumer goroutine (causes scrambled H265 pixels).
+		frame := make([]byte, c.waitSize)
+		copy(frame, c.waitData[:c.waitSize])
```
skrashevich: could you add a reproducer for the overwrite described here? In the previous implementation, `waitData` advances past the emitted frame, so subsequent appends write after the remaining data rather than over the emitted slice. This adds an allocation and a full copy for every frame. Copying can reduce retention of a shared backing array, but the claimed corruption fix needs a regression test; otherwise I would leave this change out of the command-draining fix

pkg/xiaomi/miss/cs2/conn.go
```diff
@@ -429,29 +442,54 @@ type dataChannel struct {
 	popBuf   chan []byte
 }
 
+const maxFrameSize = 2 * 1024 * 1024 // 2MB — reasonable max for 2K H265 keyframe
```
skrashevich, line 445: what protocol or device evidence establishes 2 MiB as the maximum valid frame size? This limit applies to every CS2 frame, and exceeding it terminates the connection. A bound justified only for a 2K H265 keyframe could reject valid frames from other resolutions or encodings. I have not reproduced such a rejection on hardware, but this introduces a compatibility limit unrelated to draining commands. Please justify the bound and add boundary tests, or handle it in a separate change

pkg/xiaomi/miss/cs2/conn.go
lebdim: Confirmed, fixed in 6b62cc7: capture the value before `c.waitSize` is reset, so the error now reports the real rejected size instead of always logging 0.

pkg/xiaomi/miss/cs2/conn.go
lebdim: Agreed, removed in 6b62cc7. Wrote a standalone repro of the old slice-based `Push` (50 frames, chunked into 40-byte writes to stress reslicing/capacity growth, consumer reads only after all pushes complete): zero corruption, matches your read that `waitData` already advances past the emitted frame before the next append. No tester report in this PR actually mentions pixel corruption either, so the extra alloc/copy is gone and this PR is back to just the command-drain fix, the malformed-packet bounds checks, and frame-size validation.

pkg/xiaomi/miss/cs2/conn.go
lebdim: Kept for now as a defensive cap against a corrupted size header causing permanent desync, not a protocol-derived limit; 2 MiB is generous headroom over typical H265 keyframe sizes even at 4K/high bitrate. Open to raising it, making it configurable, or splitting into a separate change, whichever you'd prefer, just let me know.

lebdim: Fair challenge, and it turned out the bound was not just poorly justified, it was also not doing its job. Reworked in dc60373.

**Why a bound is needed at all, independent of protocol or codec.** The 4-byte frame length is not constrained by anything else on the wire: the outer DRW header carries a 16-bit size, so any larger frame is reassembled across packets by `Push`, and that length is the only thing deciding when to stop appending. Without a bound, one corrupt length makes `Push` buffer up to 4 GiB before it can conclude the frame will never complete.

**The previous check did not protect 32-bit builds.** `c.waitSize = int(binary.BigEndian.Uint32(...))` is negative on 386, arm and mipsle (all shipped by `build.yml`) for any length at or above 2 GiB, so `c.waitSize > maxFrameSize` was false and execution reached `c.waitData[:c.waitSize]`. Reproduced against the previous implementation with `GOARCH=386`:

```
panic: runtime error: slice bounds out of range [:-1]
	(*dataChannel).Push(...) conn.go:475
```

The comparison now happens in the uint32 domain before the conversion, so the bound holds on every architecture. Your earlier point about reporting the real rejected size is preserved: the size lives in a local, so the error still prints the actual value.

**Value.** 2 MiB to 16 MiB, and the keyframe rationale is gone, since you are right that it was a compatibility limit dressed up as a protocol one. 16 MiB sits far above anything these cameras emit in practice while staying 256x below the unguarded worst case, which matters for the small 32-bit targets in the support matrix such as the mipsel Gateway 3 and Wyze builds.

**Tests.** Added `pkg/xiaomi/miss/cs2/conn_test.go`, the first tests in this package: exact bound accepted, bound+1 rejected with the real size in the error, `0xFFFFFFFF` and `0x80000000` rejected instead of panicking, zero-size padding still skipped, split writes still reassembled. They pass under both `GOARCH=amd64` and `GOARCH=386`.

If you would still rather keep this PR to the drain fix alone, I am happy to split the size validation into its own change or make the limit configurable, just say which you prefer.

@skrashevich could you take another look when you get a chance?

lebdim:
Hardware evidence for the 16 MiB bound, measured on real cameras rather than argued from the protocol alone.

I ran an instrumented build of this branch against two Xiaomi cameras (CW300 2K HEVC and C300) for 35 minutes, logging every frame size, every rejection and every oversize pop:

- Largest inner frame observed: **213914 bytes** (209 KiB), a 2K HEVC keyframe. That is 79x below the bound.
- Command channel frames: 85 bytes.
- Frames rejected by the guard: **0**.
- Oversize pops: **0**.

Since a bound that terminates the connection would show up as extra reconnects, I also A/B tested it against the unmodified build on the same camera, in identical 15:15 windows:

| build | producer reconnects |
|---|---|
| this branch (16 MiB guard) | 5 |
| previous build | 4 |

The difference is noise. Every reconnect in both arms was an `i/o timeout` from a camera on a weak RF link, and the guard never fired in either.

On the bound itself: the outer DRW header is 16-bit, so anything above 64K is reassembled by `Push`, which makes the 4-byte inner length the only thing deciding when reassembly stops. Without a bound, a corrupt length buffers up to 4 GiB.

The check also had a real bug, now fixed and covered by the new tests: it compared *after* the `int()` conversion, so on a 32-bit build a length at or above 2 GiB went negative, passed the check and panicked on the reslice. `build.yml` ships 386, arm and mipsle, so this was reachable. Running the new tests under `GOARCH=386` against the old code reproduces `panic: runtime error: slice bounds out of range [:-1]`; against the fixed code they pass.

@skrashevich this is the sizing evidence I owed you. Happy to move the constant if you would prefer a different headroom factor.

billxc:
Confirmed on **`chuangmi.camera.069a01`**, using go2rtc **v1.9.14** over CS2/TCP.

I reproduced the issue **without Frigate**, using a direct RTSP/TCP client: the original version failed after about **110 seconds**, with the audio timestamp resetting.

After applying **only the command-draining change**, the same check ran for **5 minutes without disconnects or timestamp regression**. A subsequent **15-minute recording** also completed; the first, middle and last segments decoded successfully with both video and audio.

Thanks to @lianshuze and @lebdim for the diagnosis and fix. Hopefully this additional hardware confirmation helps with the review.

PR #2296, opened 2026-06-11
h265: de-aggregate RFC 7798 Aggregation Packets (type 48) in RTPDepay

Novalien0424:
## Problem

`RTPDepay` in `pkg/h265/rtp.go` only special-cases Fragmentation Units (`NALUTypeFU` = 49). RFC 7798 §4.4.2 **Aggregation Packets (type 48)** fall through to the single-NAL branch and are written into the AVCC stream as one length-prefixed entry whose first payload byte (`0x60`) parses as an unspecified non-VCL NAL type.

Any H.265 RTP source that aggregates VPS/SPS/PPS(+SEI) ahead of a RAP produces an undecodable bytestream downstream:

- `magic.Keyframe` never sees the keyframe
- AVCC→AnnexB hands ffmpeg consumers corrupt data → `Invalid data found when processing input` (exit 1/183)
- `frame.jpeg` returns HTTP 500, `stream.mp4` emits 0 bytes, MSE consumers get a malformed init/bytestream

This notably affects go2rtc's own `exec:ffmpeg` producers: libavformat's RTSP muxer emits type-48 APs for HEVC when parameter sets repeat (e.g. `repeat-headers=1`), so any HEVC camera routed through an ffmpeg source (audio transcode, filters, etc.) hits this — while the same camera consumed natively works, which makes it confusing to diagnose.

## Reproduction (no camera needed)

```yaml
streams:
  hevc_ap_test:
    - exec:ffmpeg -re -f lavfi -i testsrc2=size=640x360:rate=10 -c:v libx265 -preset ultrafast -tune zerolatency -x265-params repeat-headers=1:keyint=20 -f rtsp {output}
```

`GET /api/frame.jpeg?src=hevc_ap_test` → **500 before this patch, 200 + JPEG after.** Verified both ways.

## Fix

Adds `NALUTypeAP = 48` and a de-aggregation branch in `RTPDepay` that walks the AP payload (2-byte NALU-size + NALU, per RFC 7798 §4.4.2) and emits each constituent NAL as its own length-prefixed AVCC entry. Truncated/malformed AP payloads stop parsing at the malformed unit (covered by a test) rather than emitting garbage.

Unit tests included for the AP split and malformed-AP handling; existing tests untouched and passing.

Running in production on three sites (Raspberry Pi 5 / Apple Silicon / x86_64 Docker) since 2026-06-11 against Dahua and VIVOTEK HEVC cameras.

### DIFF
--- a/pkg/h265/helper.go
+++ b/pkg/h265/helper.go
@@ -17,6 +17,7 @@ const (
 	NALUTypePPS       = 34
 	NALUTypePrefixSEI = 39
 	NALUTypeSuffixSEI = 40
+	NALUTypeAP        = 48
 	NALUTypeFU        = 49
 )
 
--- a/pkg/h265/rtp.go
+++ b/pkg/h265/rtp.go
@@ -92,6 +92,27 @@ func RTPDepay(codec *core.Codec, handler core.HandlerFunc) core.HandlerFunc {
 				buf = append(buf, (data[0]&0x81)|(nuType<<1), data[1])
 				buf = append(buf, data[3:]...)
 			}
+		} else if nuType == NALUTypeAP {
+			// RFC 7798 §4.4.2 Aggregation Packet: [PayloadHdr][2-byte size + NALU]*
+			// (no DONL — sprop-max-don-diff=0). Emitted by libavformat's HEVC RTP
+			// packetizer (e.g. exec/ffmpeg sources), which bundles VPS+SPS+PPS into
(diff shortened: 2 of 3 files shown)


tsightler:
Wow, just spent way too many hours tracing down this same exact issue, and building a fix that is nearly identical.  Causes issues with HEVC streams in Home Assistant because they always setup ffmpeg input for RTSP streams.  Would love to see this one merged at some point.

PR #2300, opened 2026-06-14
Add API to stop publishing a stream

### LINKED ISSUE #2037: Outgoing RTMPS publish consumer does not stop via API (?src= empty) – remains stuck in /api/streams
### Describe the bug
When publishing a stream to an external RTMPS server using the API, the documented method to stop the publish (sending an empty src parameter) fails to tear down the consumer.

Status: The consumer remains visible in /api/streams (bytes sent continue to increase).

Process: The underlying FFmpeg process for the publish stays running in the container.

**To Reproduce**
1. Minimal Config (go2rtc.yaml):
```
api:
  listen: ":1985"
streams:
  my_camera: ffmpeg:rtsp://admin:pass@IP_ADDRESS/...#video=h264#audio=aac
```
2. Start Publish: `curl -X POST "http://localhost:1985/api/streams?src=my_camera&dst=rtmps://URL/live/test"`
3. Verify: `curl http://localhost:1985/api/streams` shows active consumer.
4. Attempt Stop: `curl -X POST "http://localhost:1985/api/streams?src=&dst=rtmps://URL/live/test"`
5. Result: After 30s, the consumer is still present and FFmpeg is not terminated.

Expected behavior
The RTMPS consumer should be removed and the FFmpeg process terminated immediately upon receiving the empty src request.

Actual behavior
The consumer persists indefinitely. Only a container restart or a manual kill -9 on the FFmpeg PID stops the stream. Using DELETE /api/streams/my_camera works but is a destructive workaround as it removes the entire stream definition.

**Additional context**

Installation: Docker (alexxit/go2rtc:latest)

Issue persists specifically with RTMPS publish.

Source is an FFmpeg-wrapped RTSP stream.

Related issues: #1733 and #1254 (similar hangs, but not specifically for RTMPS publish).

zephleggett:
Closes #2037.

We publish out to media servers (Ant Media over RTMP/RTMPS) from production, and wanted to be nicer to them: stop a push cleanly so they aren't left holding a dead connection, and have a way to stop publishing without nuking the whole stream.

Today there's no way to stop an outgoing publish on its own. The empty-`src` call from #2037 doesn't tear down the publish consumer, so the push and its FFmpeg process keep running until the container restarts. `DELETE /api/streams?src=NAME` works but it's destructive: it removes the stream definition too.

This bit us publishing to Ant Media over an unstable network. When the outbound connection dropped, the original retry (a flat 5s sleep then recurse, no backoff, no way to cancel) reconnected every 5s forever, the dead connection lingered, and FFmpeg stayed up. A flaky link turned into a constant reconnect loop, and the only way out was restarting the container.

Two stop paths:

```text
DELETE /api/streams?src=NAME&dst=URL   stop one publish, keep the stream
DELETE /api/streams?src=NAME           stop all publishes, then delete the stream
```

The single-arg form keeps its old behaviour, it just stops active publishes first so it no longer leaks a running push.

Stopping a publish closes its live consumer and cancels the retry loop. Closing the consumer is what makes it immediate: the publish goroutine spends most of its life inside `run()` copying frames to the remote, so closing the consumer unblocks that copy and ends the push right away instead of waiting for the remote to drop. Cancelling the loop covers the gaps between retries.

Making the retry cancellable meant turning the `// TODO: more smart retry` placeholder (`sleep 5s` then recurse) into a real loop, which brings two small behaviour changes:

- retry uses the same exponential backoff schedule as `producer.go` `reconnect`, instead of a flat 5s
- publishing the same `dst` again replaces the previous publish instead of stacking a second one

Both stop calls are documented in `internal/streams/README.md`.

### Graceful RTMP teardown

The second commit tears the RTMP connection down cleanly on stop. Before, closing the consumer closed the socket with a bare TCP close, so the server only learned the publish ended when it noticed the dropped socket and freed the stream name after an idle timeout. A fast republish of the same key could then be rejected, and publish-done hooks fired late. The client now sends the `FCUnpublish` + `deleteStream` pair a normal publisher sends before disconnecting, then closes. It's best-effort (the socket is often already dead, which is why we're closing), bounded by a write deadline, and only runs for a client that reached the published state. Players and the server side are unchanged.

### On the API shape

I went with `DELETE` rather than mirroring the playback stop (`POST dst=camera&src=` empties the source). The mirror for publish would be `POST src=NAME&dst=`, but that can only stop every publish on a stream at once. A stream can publish to several destinations (the README example sends one stream to both Telegram and YouTube), and `DELETE src=NAME&dst=URL` can stop just one of them while the rest keep running. Happy to switch it to `POST` if you'd rather have the consistency, just wanted to flag the trade-off rather than decide it for you.

### Testing

- `internal/streams/publish_test.go` covers `StopPublish`, `StopAllPublish`, and the `removeHandle` republish guard: the live consumer gets closed, the retry context is cancelled, the map is cleaned up, and a stale handle doesn't clobber a newer one.
- `pkg/rtmp/unpublish_test.go` checks that closing a publishing connection sends `FCUnpublish` + `deleteStream`, and a non-publishing one sends nothing. All pass under `-race`.
- `go build`, `go vet`, and `gofmt` are clean.
- Manually: start a publish, confirm the consumer shows in `/api/streams`, then `DELETE ...&dst=...` and confirm the consumer and FFmpeg process both go away while the stream stays. Plain `DELETE ...?src=NAME` still removes the stream.

(`TestRecursion` and `TestTempate` in this package already fail on a clean `master` checkout when the package is tested in isolation, because the source handlers are registered by the main binary's blank imports and aren't linked into a bare package test. Both are unrelated to this change.)

### DIFF
--- a/internal/streams/README.md
+++ b/internal/streams/README.md
@@ -40,6 +40,14 @@ You can use the API:
 POST http://localhost:1984/api/streams?src=camera1&dst=rtmps://...
 ```
 
+Stop one publish without deleting the stream by sending a `DELETE` with the same `src` and `dst`:
+
+```text
+DELETE http://localhost:1984/api/streams?src=camera1&dst=rtmps://...
+```
+
+A `DELETE` with only `src` stops every publish for that stream and removes the stream, as before.
+
 Or config file:
 
 ```yaml
--- a/internal/streams/api.go
+++ b/internal/streams/api.go
@@ -98,6 +98,20 @@ func apiStreams(w http.ResponseWriter, r *http.Request) {
 		}
 
 	case "DELETE":
+		// with dst - stop one publish, keep the stream
+		if dst := query.Get("dst"); dst != "" {
+			if stream := Get(src); stream != nil {
+				stream.StopPublish(dst)
+			} else {
+				http.Error(w, "", http.StatusNotFound)
+			}
+			return
+		}
+
+		// without dst - stop all publishes, then delete the stream
+		if s, ok := streams[src]; ok {
+			s.StopAllPublish()
+		}
 		delete(streams, src)
 
 		if err := app.PatchConfig([]string{"streams", src}, nil); err != nil {
--- a/internal/streams/publish.go
+++ b/internal/streams/publish.go
@@ -1,27 +1,146 @@
 package streams
 
-import "time"
+import (
+	"context"
+	"time"
+
+	"github.com/AlexxIT/go2rtc/pkg/core"
+)
+
+// publishHandle tracks one active publish. Closing cons unblocks an in-flight
+// run() so a stop takes effect at once; cancel and cons are guarded by Stream.mu.
+type publishHandle struct {
+	cancel context.CancelFunc
+	cons   core.Consumer
+}
 
 func (s *Stream) Publish(url string) error {
+	s.StopPublish(url) // replace any existing publish for this url
+
+	ctx, cancel := context.WithCancel(context.Background())
+	h := &publishHandle{cancel: cancel}
+
+	s.mu.Lock()
+	if s.publishes == nil {
+		s.publishes = make(map[string]*publishHandle)
+	}
+	s.publishes[url] = h
+	s.mu.Unlock()
+
 	cons, run, err := GetConsumer(url)
 	if err != nil {
+		cancel()
+		s.removeHandle(url, h)
 		return err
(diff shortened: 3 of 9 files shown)


zephleggett:
@skrashevich this is useful to use to be nice to our media servers so they don't hold connections and have some control to stop publishing.

Issue #1225, opened 2024-06-23
Using go2rtc for HomeKit in MacOS Sonoma


AlZvl:
When using go2rtc for HomeKit in MacOS Sonoma, we get an error: ERR internal/homekit/homekit.go:128 > error="no interfaces for listen" I think this is due to the fact that port 5353 is occupied by the system (mDNSRespo).

AlexxIT:
I don't know how to reproduce this issue.

Luavis:
@AlexxIT I got same error in my laptop.  I guess it because my home network use Class-B network which is same range for docker. pkg/xnet/net.go filter out this addresses from interfaces list

PR #2305, opened 2026-06-15
feat(mdns,homekit): IPv6 transport for HomeKit accessories

### LINKED ISSUE #2266: Huge problems in setting up HomeKit: homekit.go:122 > error="no interfaces for listen"
I get this error when setting up my streams for HomeKit: **error="no interfaces for listen"**
```
23:07:54.630 INF go2rtc platform=windows/amd64 revision=b5948cf version=1.9.14
23:07:54.665 INF config path="d:\\Programme\\go2rtc\\go2rtc.yaml"
23:07:54.667 INF [api] listen addr=:1984
23:07:54.667 INF [rtsp] listen addr=:8554
23:07:54.677 INF [webrtc] listen tcp addr=[::]:8555
23:07:54.681 INF [webrtc] listen udp addr=172.16.2.21:8555
23:07:54.693 ERR github.com/AlexxIT/go2rtc/internal/homekit/homekit.go:122 > error="no interfaces for listen"
```
At the moment i struggeled for more than 6 hours in setting up my streams for homekit, everything else working fine.
I tested go2rtc as LXC (priviledged, unpriviledged) in proxmox (installed by community script) and after that several times in windows 11 25H2.
That error is persistant in 100% of cases and I have no idea what am I doing wrong.
My config is:
```
streams:
  kamera_haustuer:
    - rtsp://kamera-haustuer.lan/ch0_0.h264#video=copy&audio=none&buffer=0
rtsp:
  listen: ":8554"
api:
  listen: ":1984"
webrtc:
  listen: ":8555"
homekit:
  kamera_haustuer:
```

TA2k:
## Summary

Adds an IPv6 transport for HomeKit mDNS so go2rtc keeps working on hosts where another mDNS responder already holds the IPv4 `:5353` wildcard bind. On those hosts the HomeKit module currently exits with `error="no interfaces for listen"` (see #2266) because IPv4 was the only path wired.

The IPv6 path runs **alongside** IPv4 best-effort: when both bind, both serve; when one family fails, the other still works; when neither comes up the returned error names both causes via `errors.Join`. IPv4 success path is unchanged.

## Why

Multiple mDNS responders coexist on a single host all the time (avahi for SMB/AirPrint/AirPlay, then go2rtc for HomeKit on top, etc.). Linux serializes the IPv4 multicast `0.0.0.0:5353` bind under `SO_REUSEADDR` if both peers don't also set `SO_REUSEPORT`, so go2rtc's IPv4 bind reliably loses to the system daemon. The IPv6 multicast group `ff02::fb` typically *is* shared cleanly under `SO_REUSEADDR`, so a parallel IPv6 path is a robust fallback.

Beyond just binding the v6 socket, two more fixes were necessary to actually carry HomeKit traffic over an IPv6-only mDNS path:

1. **`AppendEntry` only emitted A xor AAAA based on the answering local IP family.** When a query arrived over IPv6, the response carried only AAAA; iOS' Home app, which prefers IPv4 for the HAP TCP connect, then had no usable address and silently gave up. The fix emits both A and AAAA when the same interface has both — same behaviour as avahi/Bonjour.

2. **SRTP `SetupEndpoints` response built its `IPAddr` via `addr.IP.To4().String()` unconditionally**, which yields `<nil>` on an IPv6 socket. iOS rejects that response and never starts the live stream. The fix detects the family, sets `Address.IPVersion` per HAP R15 §10.4.2, strips the `%zone` suffix in the textual form, and patches the local zone into the remote SRTP endpoint when the controller's announced RTP address is IPv6 link-local (otherwise `WriteTo` to `fe80::` fails with `EINVAL`).

## Files changed

```
 pkg/homekit/consumer.go     |  39 ++++++--
 pkg/mdns/client.go          | 217 ++++++++++++++++++++++++++++++++++++--
 pkg/mdns/server.go          | 219 ++++++++++++++++++++++++++++++++++++++++-
 pkg/mdns/syscall.go         |  13 +++
 pkg/mdns/syscall_bsd.go     |  10 ++
 pkg/mdns/syscall_windows.go |  12 +++
 pkg/srtp/session.go         |  14 ++-
 7 files changed, 524 insertions(+), 59 deletions(-)
```

## Backwards compatibility

- IPv4 success path is unchanged
- `AppendEntry(msg, entry, service, ip)` signature is unchanged — it delegates to the new internal dual-stack helper. Callers outside this repo are not affected
- New exported API: `mdns.MulticastAddrV6`, `mdns.AppendEntryDual`, `srtp.Session.SetZone`, plus `Browser.SendsV6` / `NetsV6` / `RecvV6` (parallel to existing exported `Sends` / `Nets` / `Recv`)

## Test

```
$ go build ./...
$ go test ./pkg/mdns/
ok      github.com/AlexxIT/go2rtc/pkg/mdns
```

linux/amd64 build deployed in a Frigate container running alongside another mDNS responder that holds the wildcard `:5353` bind. Before: `error="no interfaces for listen"` at HomeKit init. After: cameras pair, snapshots load, live stream plays in the iOS Home app over the LAN.

## Closes / refs

- Fixes #2266
- Refs #1225 (the `xnet.IPNets` 172.16/12 filter blocks similar setups; this PR sidesteps that via the v6 fallback rather than changing the v4 filter)
- Complements #2304 (per-IP bind-error diagnostic)

### DIFF
--- a/pkg/homekit/consumer.go
+++ b/pkg/homekit/consumer.go
@@ -5,6 +5,7 @@ import (
 	"io"
 	"math/rand"
 	"net"
+	"strings"
 	"time"
 
 	"github.com/AlexxIT/go2rtc/pkg/core"
@@ -87,11 +88,29 @@ func (c *Consumer) GetAnswer() *camera.SetupEndpointsResponse {
 	c.videoSession.Local = c.srtpEndpoint()
 	c.audioSession.Local = c.srtpEndpoint()
 
+	// HAP Address.IPVersion: 0 = IPv4, 1 = IPv6 (HAP R15 §10.4.2). The
+	// previous code called addr.IP.To4().String() unconditionally, which
+	// returns "<nil>" on an IPv6 socket and breaks the response.
+	addr := c.conn.LocalAddr().(*net.TCPAddr)
+	var ipVersion byte
+	var ipStr string
+	if ip4 := addr.IP.To4(); ip4 != nil {
+		ipVersion = 0
+		ipStr = ip4.String()
+	} else {
+		ipVersion = 1
+		ipStr = addr.IP.String()
+		if i := strings.IndexByte(ipStr, '%'); i > 0 {
+			ipStr = ipStr[:i] // strip zone — iOS expects the bare address
+		}
+	}
+
 	return &camera.SetupEndpointsResponse{
 		SessionID: c.sessionID,
 		Status:    camera.StreamingStatusAvailable,
 		Address: camera.Address{
-			IPAddr:       c.videoSession.Local.Addr,
+			IPVersion:    ipVersion,
+			IPAddr:       ipStr,
 			VideoRTPPort: c.videoSession.Local.Port,
 			AudioRTPPort: c.audioSession.Local.Port,
 		},
@@ -126,6 +145,15 @@ func (c *Consumer) SetConfig(conf *camera.SelectedStreamConfiguration) bool {
 
 	c.srtp.AddSession(c.videoSession)
 	c.srtp.AddSession(c.audioSession)
+
+	// IPv6 link-local: iOS sends only the bare address in the SetupEndpoints
+	// request, so the kernel cannot pick an outgoing interface for our SRTP
+	// writes. Recover the zone from the local HAP-control TCP socket — that
+	// socket is necessarily on the same interface as the controller itself.
+	if a, ok := c.conn.LocalAddr().(*net.TCPAddr); ok && a != nil {
+		c.videoSession.SetZone(a.Zone)
+		c.audioSession.SetZone(a.Zone)
+	}
 
 	return true
 }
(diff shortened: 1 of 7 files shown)


TA2k:
@skrashevich What is your opinion  about that?

skrashevich:
1. In most cases, daemons listen to [::]:5353, which binds them on both ipv4 and ipv6. Give a real repeatable example when the port is occupied on ipv4, but not occupied on ipv6

2. PR needs to be rebased from dev to main anyway

TA2k:
1. this is a very common use case if your host system is a NAS with mdns support for Timemachine this will occupy the ipv4 port if you now setup go2rtc in a docker you cannot use the mdns port and this means no HomeKit support
2. done

skrashevich:
> this is a very common use case if your host system is a NAS with mdns support for Timemachine 

Synology, for example?

TA2k:
Synology, Truenas, green nas

TA2k:
Did you think about this?

Issue #1750, opened 2025-05-25
2-way audio through SIP client


felipecrs:
This is an alternative to https://github.com/AlexxIT/go2rtc/issues/409.

Most SIP doorbells like Dahua and Hikvision have their own SIP server built-in which other devices can register into.

This means go2rtc does not actually need to act as a SIP server. It could act as a SIP client only.

Example on how this could be configured:

```yaml
stream:
  dahua_doorbell:
    doorbell:
      # H264, PCMA
      - rtsp://admin:pass@192.168.1.40/cam/realmonitor?channel=1&subtype=0
      # 2-way audio over SIP
      - sip://9001:sippassword@192.168.1.40:5060/?call_to=8001
```

```
sip://<username to register in sip server>:<password to register in sip server>@<sip server IP>:<sip server port>/?call_to=<sip number to call>
```

So when 2-way audio is requested, go2rtc would register itself in the SIP server and initiate a call to the number.

altserg:
From my experience SIP often uses low quality PCMU/8000 codec while the native camera stream can be PCM/16000 (For Dahua doorbell)
I also have not managed to connect SIP audio to Dahus SIP server using pjsip lib. Command channel and SDP negotiation work, but the audio channel is not established for some reason. If someone will manage to connect SIP audio Dahua doorbell, I would like to know the recipe
https://github.com/pjsip/pjproject/issues/2823

felipecrs:
Through Asterisk (pjsip) it works just fine. You can refer to my project (https://github.com/felipecrs/dahua-vto-on-home-assistant). Even though I don't use SIP for 2-way audio, it was working last time I tried with the [sip-hass-card](https://github.com/TECH7Fox/sip-hass-card).

altserg:
Dahua embedded SIP server does not work for me, but this might be my local mistake. I do not want to install Asterick as this is yet another level of complexity and dependency. But I did tried it in the past, and indeed Dahua does work via 3rd party SIP server

felipecrs:
Yeah. I was also not able to connect to the embedded SIP server.

ngres:
I'm currently working on integrating both a SIP User Agent Server and Client using sipgo: https://github.com/ngres/go2rtc

[...]

Issue #409, opened 2023-05-04
Basic SIP server for doorbells


felipecrs:
This project never stops surprising me. It is like the always missing glue for cameras and smart homes. And I have another suggestion.

An embedded and very basic SIP server that doorbells like Dahua VTO or Hikvision can connect to.

go2rtc has 2-way audio support for doorbells, which is awesome. But that's not enough for a fully functional doorbell setup. In other words, a SIP server is still required even though go2rtc can do 2-way audio.

That's because the doorbell still needs to signal and receive signals to react on certain events like:

- Doorbell is calling
- Call has been accepted
- Call has been ended

In my understanding, if go2rtc could expose a basic SIP server in which the doorbell could connect to, no other project like the [Asterisk add-on](https://github.com/TECH7Fox/asterisk-hass-addons) would be required (Asterisk is VERY hard to manage, and overkill for the purpose).

go2rtc as a SIP server would:

- Allow to receive calls from the doorbell (i.e. it would let the doorbell call it and ring)
  - When the VTO doesn't have a SIP server configured to accept a call from it, it basically doesn't work.
  - Users can then create Home Assistant cards and automations to notify when the doorbell is ringing
- Send the call accepted signal to the doorbell as soon as some 2-way audio connection is established
- Send the call ended signal to the doorbell as soon as that 2-way audio connection is dropped

Luckily, there seems to be some good SIP servers written in golang out there:

- https://github.com/emiago/sipgo
- https://github.com/ghettovoice/gosip
- https://github.com/jart/gosip

So, what do you think?

Refs #49

AlexxIT:
I knows nothing about SIP technologies and didn't have any hardware that using it.

felipecrs:
That's sad, but thank you.

I know very little of Golang, but nevertheless I'll try to build something.

felipecrs:
It does not look very difficult according to the examples of https://github.com/emiago/sipgo. However, my doorbell is in a very inaccessible place which makes it very hard for me to work with it.

But I will try to build something when I get a chance.

AlexxIT:
Can your doorbell works with this integration?
https://www.home-assistant.io/integrations/voip/

felipecrs:
I don't think so. That does not even look like SIP also.

I think that integration only works with the Grandstream conversor for old SIP phones.

---

Off topic, SIP itself could be used as a 2-way audio source provider. In fact, even as a video source.

[...]

PR #2317, opened 2026-06-25
go2sip: Simple Self-Answering SIP Consumer

tylerransdell:
# SIP Self-Answer Consumer (edited down to audio-only, less logic)

You configure a port and a stream, then dial `sip:go2rtc:<port>`. go2rtc answers automatically, negotiates audio, and gives you real-time RTP access to that stream.

The number of use cases is broad. Users can interact with cameras directly. Builders get directish access to go2rtc's dozens of camera integrations.

## Config

```yaml
sip:
  rtp_port_range: "31000-31100" #optional, default shown
  host_ip: "192.168.1.26" #optional SDP c= IN IP4 "" for Docker NAT, otherwise local
  consumers:
    - port: 5060
      stream: front_door
    - port: 5061
      stream: back_door
```

## Codec negotiation

go2rtc negotiates the best codec that both the phone and the camera support.
No transcoding is performed — the user handles that via `exec`/`ffmpeg` on the
stream if needed.

| Caller + Camera support | Answer |
|---|---|
| Opus | sendrecv Opus |
| G722 | sendrecv G722 |
| PCMA | sendrecv PCMA |
| PCMU | sendrecv PCMU |
**Package answers with sendrecv to give caller a keepalive path in case they do not send RTCP.

## Backchannel

If main codec is same as back, a single sendrecv SDP is made. This is most compatible.
If codecs differ, a separate sendonly and recvonly are made. This is valid but less universal.

## Video

**Not implemented.** This is an audio-only implementation.

### RTCP keepalive
BYE and CANCEL sre primary medhods to terminate sessions but are not always reliable.
The RTP endpoint listens on two sockets: the RTP port and RTCP port (port+1).
Both RTP and RTCP packets from the caller reset the activity timer. This keeps
one-way sessions (e.g. no common codec where only RTCP flows back) alive.

- All RTP packets → OnActivity fires, session stays alive
- All RTCP packets (SR/RR/other compound packets) → OnActivity fires, session stays alive
- RTCP Receiver Reports are always generated in response to incoming SRs
- 1-minute silent timeout reaps dead sessions regardless of direction

## Limitations

- No transcoding is performed. The same codec/s are offered as the camera
  provides. The user is responsible for providing matching in and out codecs
  and for not providing AAC.

- I have included just the tools to get through Docker NAT and into a LAN. There is no auth so that should be the limited scope of use cases. A proper PBX or other integration is required to get beyond LAN.

### Packages added

- **`pkg/rtp/`** - RTP endpoint. Implements `core.Consumer` (camera to caller) and `core.Producer` (caller to camera backchannel). No transcoding. Handles RTCP keepalive via Receiver Reports. Dual-socket design (RTP + RTCP on port+1).

- **`internal/sip/`** - SIP protocol handler. Listens for INVITE/BYE/CANCEL, negotiates SDP, manages sessions, and wires RTP endpoints to go2rtc streams. Audio-only.

### Integration with go2rtc core

The SIP consumer uses only public go2rtc APIs (`stream.AddConsumer`, `stream.RemoveConsumer`). No modifications to the core streams package beyond a minimal `Producers()` accessor for codec discovery. All SIP-specific logic - codec selection, SDP parsing, session management - lives in `internal/sip/`.

### Lifecycle

```
Phone                    go2rtc                         Camera
  |--- INVITE (SDP) ---->|                              |
  |                      |--- lookup stream ----------->|
  |                      |--- create RTP endpoint ----->|
  |<-- 200 OK (SDP) -----|                              |
  |========= RTP audio (camera phone) =================>|
  |<======== RTP audio (phone camera) ==================|
  |--- BYE ------------->|                              |
  |                      |--- cleanup session ----------|
```

## Status

Still a draft. General architecture and methods are now finalized. Significant testing is needed with various hardware and other codecs.


felipecrs:
This looks awesome.

tylerransdell:
Much simpler now.... No smart matching logic or transcoding. It's just if matches, make a standard one-liner. If not, offer independently.

felipecrs:
Refs https://github.com/AlexxIT/go2rtc/issues/1750, https://github.com/AlexxIT/go2rtc/issues/409

PR #2324, opened 2026-07-02
Add quality selection for webrtc format=wyze source

jdrusso:
My Wyze Bulb Camera defaults to 360p over WebRTC until you request something else. This PR add an optional #profile= param that sends the switch command on connect, like

    webrtc:http://192.168.1.123:5000/signaling/cam?kvs#format=wyze#profile=2k

We have to use OnDial instead of OnOpen because the camera-side Amazon KVS SDK never sends a DATA_CHANNEL_ACK 🤷 

Tested on a Wyze Bulb Cam HL_BC

Note this same protocol also has a `resolutions`/`query` message that lists a camera's supported profiles

Thanks Fable!

### DIFF
--- a/internal/webrtc/client.go
+++ b/internal/webrtc/client.go
@@ -55,7 +55,7 @@ func streamsHandler(rawURL string) (core.Producer, error) {
 				return milestoneClient(rawURL, query)
 			} else if format == "wyze" {
 				// https://github.com/mrlt8/docker-wyze-bridge
-				return wyzeClient(rawURL)
+				return wyzeClient(rawURL, query)
 			} else if format == "creality" {
 				return crealityClient(rawURL)
 			} else {
--- a/internal/webrtc/kinesis.go
+++ b/internal/webrtc/kinesis.go
@@ -68,6 +68,38 @@ func kinesisClient(
 		return nil, err
 	}
 
+	// Wyze cameras select stream quality via a viewer-created data channel;
+	// without the switch command the camera serves its lowest profile.
+	if profile := query.Get("profile"); profile != "" {
+		dc, dcErr := pc.CreateDataChannel("My Data", nil)
+		if dcErr != nil {
+			log.Warn().Err(dcErr).Caller().Send()
+		} else {
+			dc.OnMessage(func(m pion.DataChannelMessage) {
+				log.Debug().Msgf("[webrtc] wyze dc recv: %s", m.Data)
+			})
+			// OnDial instead of OnOpen: OnOpen fires only after the remote's
+			// DATA_CHANNEL_ACK, which the Amazon KVS WebRTC C SDK on these
+			// cameras never sends (its DCEP handler consumes OPEN without
+			// replying). OnDial fires once OPEN is sent, and RFC 8832 allows
+			// sending data without waiting for the ACK.
+			dc.OnDial(func() {
+				var dcID uint16
+				if dc.ID() != nil {
+					dcID = *dc.ID()
+				}
+				log.Debug().Msgf("[webrtc] wyze dc dialed, requesting profile=%s", profile)
+				msg := fmt.Sprintf(
+					`{"function":"switch","version":1,"payload":{"profile":%q},"interface":"resolution","resource_id":%d}`,
+					profile, dcID,
+				)
+				if err := dc.SendText(msg); err != nil {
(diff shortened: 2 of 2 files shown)


seitzbg:
Can confirm this works for me too.

seitzbg:
Confirmed working on 6 Wyze Bulb Cams (`HL_BC`) — all came up at 2304x1296 with this. Thanks for putting it together.

Two notes from running it through docker-wyze-bridge (the IDisposable fork):

- There's no default profile, so anything that builds the source URL without a `#profile=` fragment still gets 360p. The bridge generates the wyze WebRTC source itself and doesn't add one, so out of the box it stays 360p. I patched go2rtc to default the profile to `2k` when the fragment is absent and that fixed all six. Might be worth defaulting it for `format=wyze`, or at least calling out that the consumer has to add the fragment.
- Since the source is on-demand, a cam can come up at 360p and only switch to 2k once something consumes it — a one-shot snapshot races the switch. Once a persistent consumer is on it, it holds 2k. Not a bug, just a heads-up for anyone testing.

Happy to test anything else on `HL_BC` hardware.

seitzbg:
Ended up going the native TUTK route on these instead of WebRTC, and it gets full 2K without touching the cloud path at all. The trick for HL_BC is the K10056 resolution byte: it wants wire byte 5 (the FrameSizeFloodlight value), not the usual 2K byte 4. With byte 4 it just stays at 360p even though the camera ACKs the request. Filed that as #2497.

The native path had two transport bugs too: it misread the 0x09 extended frame ending (a 0x28 frame-info length was getting treated as packet index 40), and it dropped fragments of the big 2K keyframes. #2498 has the frame-ending fix; the keyframe loss needed selective retransmission on top of that.

Running all six of my Bulb Cams on the native build now at 2304x1296, and since it's local there's none of the WebRTC packet loss/corruption. Figured the resolution fix in #2497 is worth folding in either way.

brockpalen:
Tested this on a different model and signaling backend — reporting the results since they may be useful for scoping.

**TL;DR:** `#profile=` can't work on the Wyze Cam Pan Duo (`GW_DUO`). The data channel is negotiated in SDP but the SCTP association never completes, so `OnDial` never fires. That's one layer below the `DATA_CHANNEL_ACK` problem this PR works around.

## Setup

| | |
|---|---|
| Build | `jdrusso/go2rtc@a7f9c64`, go1.27.1, `CGO_ENABLED=0` |
| Camera | Wyze Cam Pan Duo (`GW_DUO`), dual 2304×1296 sensors |
| Signaling | Wyze **mars-webcsrv** (not AWS KVS) |
| Shim | `IDisposable/docker-wyze-bridge` 4.5.0 |
| Source URL | `webrtc:http://127.0.0.1:5080/internal/wyze/webrtc/<cam>#format=wyze#profile=2k` |

Run standalone on loopback so it didn't contend with the production bridge.

## Result

Resolution is **1296×728 with and without `#profile=2k`** — the camera's default composite of both lenses.

The PR's code path *is* reached: our SDP offer contains the `m=application … webrtc-datachannel` section, and the camera's answer accepts it with `a=sctp-port:5000`. But `[webrtc] wyze dc dialed` never logs.

ICE and DTLS connect, **media flows normally** (frames decode fine) — but SCTP stays in `connecting` for the entire session lifetime and the data channel never opens.

So on this camera the sequence stops earlier than on the `HL_BC`. There, SCTP came up and DCEP OPEN was sent (hence `OnDial` firing); the camera just never ACKed. Here the camera appears to accept the `m=application` section in SDP and then never answer SCTP INIT. If that's right, no data-channel-based approach can reach this model without a firmware change — it isn't something the `OnDial`-vs-`OnOpen` workaround can bridge.

Worth noting the camera advertises `profile-level-id=640032` (H.264 High, Level 5.0), so the hardware is clearly capable of more than 1296×728 — the cap looks server-side.

## Unrelated gotcha for anyone else testing

**The Duo appears to accept only one WebRTC viewer at a time.** With another client already streaming, my `SDP_OFFER` got **no answer at all** (30s+ silence, `pc` stuck in `new`). Freeing the camera made the answer arrive in **1.1s**. 

Easy to misread as "the profile parameter broke signaling" when it's really session contention. Worth freeing the camera before testing.

Happy to run further experiments on this hardware if it would help — I have the build environment set up.

PR #2327, opened 2026-07-03
Reap the keyframe consumer when the client disconnects

zephleggett:
We poll snapshots from a lot of cameras, and when one goes offline the keyframe consumers just keep growing until the container restarts.

`GET /api/frame.jpeg` and `/api/frame.mp4` add a keyframe consumer and block in `cons.WriteTo(once)` until the first keyframe arrives, then call `RemoveConsumer` once it returns. The sink is an in-memory `OnceBuffer`, decoupled from the socket, so unlike the streaming handlers (which write to the live `ResponseWriter`) a client disconnect doesn't error the write and unblock it.

So when a source never delivers a keyframe (an offline camera that's still registered, a very long GOP, a broken re-publish) and the client times out and drops, `WriteTo` blocks forever, `RemoveConsumer` is never reached, and the consumer leaks. One per request. A snapshot poller hitting `frame.jpeg` every 30s against a stream that couldn't produce a keyframe built up thousands of stuck consumers, each pinning a receiver on the producer so memory kept climbing, while `ss` showed zero live connections to the API port.

The fix ties the consumer's lifetime to the request context, the same idiom `handlerMP4` already uses: a goroutine stops the consumer when the client disconnects or the handler returns, which unblocks `WriteTo` and lets it be removed. It also skips the transcode and response when the client is already gone.

There's one related fix in `pkg/core`. `WriteBuffer.Close` read `w.Writer` outside the mutex while `Reset` writes it, a race the detector flags as soon as `Close` runs concurrently with `WriteTo`, which this change now does on every disconnect. Snapshot the writer under the lock and add a `closed` guard so a `Close` that lands before `WriteTo` registers its wait can't re-block. `pkg/core/writebuffer_test.go` covers both the reap and the race.

### DIFF
--- a/internal/mjpeg/mjpeg.go
+++ b/internal/mjpeg/mjpeg.go
@@ -83,11 +83,23 @@ func handlerKeyframe(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
+	// WriteTo below blocks until a keyframe arrives, and the in-memory OnceBuffer
+	// sink never errors on client disconnect to unblock it. Reap the consumer when
+	// the request context ends so a stuck WriteTo can't leak it.
+	ctx := r.Context()
+	go func() {
+		<-ctx.Done()
+		_ = cons.Stop()
+		stream.RemoveConsumer(cons)
+	}()
+
 	once := &core.OnceBuffer{} // init and first frame
 	_, _ = cons.WriteTo(once)
 	b = once.Buffer()
 
-	stream.RemoveConsumer(cons)
+	if ctx.Err() != nil {
+		return // client gone before a keyframe arrived
+	}
 
 	switch cons.CodecName() {
 	case core.CodecH264, core.CodecH265:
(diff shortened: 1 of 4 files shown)


zephleggett:
@skrashevich when polling an offline camera for snapshots the consumers continuously grow.

skrashevich:
You have an additional goroutine for every http request. it doesn't look cool.

Maybe it's better to do it through select?

ajplotkin:
Pulled this into my go2rtc build (Nest cameras served to HomeKit via Homebridge, with a snapshot-warmer that polls `frame.jpeg` across several cameras). The "poll a lot of cameras, one goes offline, keyframe consumers pile up on the blocking `WriteTo`" case is exactly my setup, so reaping the consumer on client disconnect is the right hardening — thanks for writing it. Applied cleanly against 1.9.14 via `gh pr diff | git apply` (the branch history made a straight cherry-pick awkward, but the diff itself applies fine). +1 to merging.

PR #2329, opened 2026-07-03
Propagate RTMP publish errors and write extended timestamps

zephleggett:
Two correctness fixes in the RTMP publish path, both of which bit us publishing to Ant Media from production.

**1. Propagate publish errors.** `writePublish` and `writePlay` returned nil when `readResponse` failed, so a dropped or rejected publish (server close, auth or app error, a network blip) came back as success. The consumer then wrote media into a dead socket and only failed later, turning a transient error into a reconnect loop. Now they return the error, and `DialPublish`/`DialPlay` close the connection on failure so we don't leak the socket on every retry.

**2. Extended timestamps.** The chunk writer truncated timestamps to 24 bits (the old `TODO: timeMS more than 24 bit`). Publishing to Ant Media for long stretches, once the relative timestamp crosses `0xFFFFFF` (~4h40m) it wraps and the stream drops or stutters. Now it writes the `0xFFFFFF` sentinel plus a 4-byte extended timestamp per RTMP spec 5.3.1.3, repeated on every type 3 continuation chunk. The read path already handled this.

`pkg/rtmp/conn_test.go` covers both.

### DIFF
--- a/pkg/rtmp/client.go
+++ b/pkg/rtmp/client.go
@@ -25,10 +25,12 @@ func DialPlay(rawURL string) (*flv.Producer, error) {
 
 	client, err := NewClient(conn, u)
 	if err != nil {
+		_ = conn.Close()
 		return nil, err
 	}
 
 	if err = client.play(); err != nil {
+		_ = client.Close()
 		return nil, err
 	}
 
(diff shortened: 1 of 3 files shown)


zephleggett:
@skrashevich when publishing to ANT media server for a long persons of time the timestamp truncation can drop or stutter the stream. The code had a previous TODO to fix this

PR #2330, opened 2026-07-03
Latch FLV timestamp baseline on the first packet

zephleggett:
We work with a large range of IP cameras, and some of them open the stream with the first packet already at RTP timestamp 0. This one is about stream startup: getting a usable timestamp to the client as soon as that first packet lands.

The FLV muxer used `ts0 == 0` as its "baseline not captured yet" flag. When a camera's first packet really is 0, that flag never latched, so it re-baselined on the second packet instead. The publish then opened with two frames at timestamp 0 and a skewed clock, which a strict server like Ant Media rejects and never brings the broadcast live.

The fix captures `ts0` on the first packet with an explicit flag rather than leaning on `0` as a sentinel.

### DIFF
--- a/pkg/flv/flv_test.go
+++ b/pkg/flv/flv_test.go
@@ -3,8 +3,40 @@ package flv
 import (
 	"testing"
 
+	"github.com/AlexxIT/go2rtc/pkg/core"
+	"github.com/pion/rtp"
 	"github.com/stretchr/testify/require"
 )
+
+func tagTimeMS(tag []byte) uint32 {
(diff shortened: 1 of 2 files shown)


zephleggett:
@skrashevich we work with a large range of IP cams. This fix is to improve stream startup times ensuring a timestamp can be read by client as soon as possible.

skrashevich:
Here we need to wait for the review @AlexxIT

I'm not competent enough in this magic :)

Issue #2294, opened 2026-06-09
Xiaomi Smart Camera C300 streams: read udp [::]:xxxx: i/o timeout on Xiaomi camera (Windows / go2rtc 1.9.14)


NEVARLeVrai:
Describe the bug

When using Xiaomi CS2 cameras with go2rtc on Windows, the stream randomly fails after a few seconds with the following error:

streams: read udp [::]:55255: i/o timeout

The stream is initially established successfully, but then the connection drops and go2rtc attempts to reconnect.

Environment

OS: Windows 10/11 (amd64)
go2rtc version: 1.9.14 (b5948cf)
Camera: Xiaomi C01A01 (xiaomi.camera.c01a01)
Protocol: xiaomi:// (vendor=cs2)
Network: LAN (192.168.1.x)

Logs

INF go2rtc version=1.9.14
DBG xiaomi: dial ... vendor=cs2
DBG [webrtc] new consumer src=hall-camera
DBG [mp4] add consumer error="streams: read udp [::]:55255: i/o timeout"
DBG xiaomi: dial ... (reconnect attempt)

Expected behavior

Stream should remain stable without UDP read timeouts.

Actual behavior

After ~5 seconds:

UDP read timeout occurs on [::]:xxxxx
MP4/WebRTC consumer fails
go2rtc retries Xiaomi dial

Notes / observations

Connection to camera is initially successful
Timeout happens on local UDP socket ([::]), not on camera IP
Reconnect loop is triggered automatically
Possibly related to UDP/WebRTC handling on Windows IPv6 dual-stack sockets

Questions

Is this a known issue with Xiaomi CS2 backend?
Should vendor=cs2 use forced TCP instead of UDP?
Could this be related to IPv6 [::] UDP socket binding on Windows?

Additional info

Stream config:

streams:
  hall-camera:
    - xiaomi://...@192.168.1.75?model=xiaomi.camera.c01a01

Issue #2026, opened 2026-01-08
Unstable xiaomi camera connection (model: chuangmi.camera.051a01)


Johnnybyzhang:
I've been having unstable connection to my xiaomi camera. The logs keeps showing:

> 13:06:46.562 | warn | undefined error=miss: read raw: i/o timeout url=xiaomi://REDACTED:cn@192.168.5.68?did=1072525428&model=chuangmi.camera.051a01 caller=github.com/AlexxIT/go2rtc/internal/streams/producer.go:170

I was wondering if anything could be done to help with debugging this behaviour.

Side note: I am attempting to intergrate this camera into frigate through go2rtc, everything else has been flawless, except that this camera is giving me headaches.

This seem like a problem that only occurs over time, all my initial tests seems fine.

sumsethan:
same issue here , I  use rtsp-to-onvif to pack rtsp stream for UniFi protect, streaming seems like stuck or jump backwards 

The logs keeps showing follows too

> warn	undefined error=miss: read raw: i/o timeout url=xiaomi://uid:cn@172.16.10.129?did=did&model=chuangmi.camera.055a02 caller=github.com/AlexxIT/go2rtc/internal/streams/producer.go:170

AlexxIT:
`cs2+udp` should be improved in latest master version.

AlexxIT:
https://github.com/AlexxIT/go2rtc/releases/tag/v1.9.14

PR #2333, opened 2026-07-03
fix(cs2): enable TCP keepalive for faster WiFi drop detection

INEEDBUG:
## Problem

When a Xiaomi camera is connected over WiFi, brief WiFi drops of 1-2 seconds cause go2rtc to lose the stream. Without TCP keepalive, the disconnection is only detected after the 10-second read deadline.

## Changes
**pkg/xiaomi/miss/cs2/conn.go:** Enable TCP keepalive with 3s interval
**pkg/xiaomi/miss/producer.go:** Reduce read deadline from 10s to 5s

## Related
#2294 #2026 #2057

## Testing
Tested with Xiaomi CW400 (isa.camera.hlc8) over WiFi

### DIFF
--- a/pkg/xiaomi/miss/producer.go
+++ b/pkg/xiaomi/miss/producer.go
@@ -131,7 +131,7 @@ func (p *Producer) Start() error {
 	var audioTS uint32
 
 	for {
-		_ = p.client.SetDeadline(time.Now().Add(10 * time.Second))
+		_ = p.client.SetDeadline(time.Now().Add(5 * time.Second))
 		pkt, err := p.client.ReadPacket()
 		if err != nil {
 			return err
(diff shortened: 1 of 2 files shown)


pkg/xiaomi/miss/producer.go
```diff
@@ -131,7 +131,7 @@ func (p *Producer) Start() error {
 	var audioTS uint32
 
 	for {
-		_ = p.client.SetDeadline(time.Now().Add(10 * time.Second))
+		_ = p.client.SetDeadline(time.Now().Add(5 * time.Second))
```
skrashevich: Magic numbers - bad. changing one magic number to another without comment - even worse

INEEDBUG:
Addressed in 1dfd962 — reverted the deadline change entirely. TCP keepalive (3s interval) alone handles WiFi drop detection properly. The 10s read deadline was fine for preventing hung connections and was never the problem.

Thanks for the review @skrashevich!

PR #2334, opened 2026-07-05
Add AV1 codec support for WebRTC

duckysownyou:
## What

Adds AV1 as a negotiable video codec for WebRTC consumers.

- `pkg/webrtc/api.go` — register AV1 in the pion MediaEngine (PT 105, same
  RTCP feedback set as the other video codecs), and request larger socket
  buffers on the WebRTC UDP mux listeners (see below).
- `pkg/av1/rtp.go` (new) — `RTPDepay`/`RTPPay` handlers mirroring the existing
  H264/H265 structure, built on `pion/rtp`'s `AV1Depacketizer`/`AV1Payloader`:
  incoming AV1 RTP is reassembled into temporal units, then re-fragmented at
  MTU 1200 (RTSP sources are often TCP-interleaved and can carry RTP packets
  larger than the UDP/TURN-safe MTU).
- `pkg/webrtc/consumer.go` — wire the AV1 case into `AddTrack`.

## Why the pacing + socket buffers

A 4K AV1 keyframe temporal unit spans hundreds of RTP packets. Writing them
in one tight loop overflowed UDP socket buffers — ~50% packet loss measured on
a gigabit LAN with the kernel-default (~208KB) send buffer. Two mitigations:

- light pacing in the AV1 payloader (2ms per 16 packets); each go2rtc sender
  runs on its own goroutine, so this never delays the producer or other
  consumers,
- `SetWriteBuffer(4MB)` / `SetReadBuffer(1MB)` on the UDP mux listeners (this
  also benefits large H265 keyframes).

## Tested

- UniFi Protect G6 PTZ, 4K AV1 via `rtspx://` → Chrome 148: AV1 negotiated
  first in the answer, 3840x2160 frames decoded, direct and through a coturn
  TURN relay.
- H264/H265/audio paths untouched; regression-checked against an H265 camera
  (MSE + WebRTC negotiation unchanged).

## Not included

MSE/fMP4 AV1 (`av01` muxing) — separate concern, tracked in #2260. Browsers
that can't receive AV1 over WebRTC still get the "codecs not matched" error
as before.

### DIFF
--- a/pkg/av1/rtp.go
+++ b/pkg/av1/rtp.go
@@ -1,0 +1,101 @@
+package av1
+
+import (
+	"time"
+
+	"github.com/AlexxIT/go2rtc/pkg/core"
+	"github.com/AlexxIT/go2rtc/pkg/h264"
+	"github.com/pion/rtp"
+	"github.com/pion/rtp/codecs"
+)
+
+// RTPDepay - depacketize AV1 RTP packets (https://aomediacodec.github.io/av1-rtp-spec/)
+// into temporal units in the AV1 low overhead bitstream format (OBUs with obu_size_field).
+// One output packet = one temporal unit (same convention as h264/h265 access units).
+func RTPDepay(handler core.HandlerFunc) core.HandlerFunc {
+	depack := &codecs.AV1Depacketizer{}
+
+	buf := make([]byte, 0, 512*1024) // 512K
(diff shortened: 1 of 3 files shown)


skrashevich:
@AlexxIT the magic of codecs is beyond my competence. I can add to dev for testing after your positive review

PR #1669, opened 2025-03-22
feat(nest): add retry logic for 429 and 409 errors with exponential backoff

hnws:

PR #2335, opened 2026-07-06
fix(nest): per-stream session state and re-arming stream extend timer

### LINKED ISSUE #2319: nest: nil-pointer panic in StartExtendStreamTimer (race with StopExtendStreamTimer) crash-loops Nest WEB_RTC sources
### Describe the bug

The `nest:` source crash-loops with a nil-pointer dereference panic in `StartExtendStreamTimer`. It became reproducible after Google migrated Nest accounts to the Google Home app: `StreamExpiresAt` now reports a very short remaining lifetime, so the extend-stream timer churns rapidly and reliably loses a race between the timer goroutine and `StopExtendStreamTimer`.

### Version

1.9.14 — also present on current `master` (`pkg/nest/api.go` is unchanged).

### Panic

```
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation]
goroutine N [running]:
github.com/AlexxIT/go2rtc/pkg/nest.(*API).StartExtendStreamTimer.func1()
        .../pkg/nest/api.go:474
```

### Root cause

```go
func (a *API) StartExtendStreamTimer() {
    if a.extendTimer != nil {
        return
    }
    a.extendTimer = time.NewTimer(time.Until(a.StreamExpiresAt) - time.Minute)
    go func() {
        <-a.extendTimer.C   // (api.go:474) reads a.extendTimer when the goroutine runs
        ...
    }()
}

func (a *API) StopExtendStreamTimer() {
    if a.extendTimer != nil {
        a.extendTimer.Stop()
        a.extendTimer = nil  // niled concurrently with the goroutine above
    }
}
```

The goroutine evaluates `a.extendTimer.C` when it runs. If `StopExtendStreamTimer` has already set `a.extendTimer = nil` (stream teardown / re-arm), the expression dereferences a nil `*time.Timer` and panics. With the shortened post-migration expiry, start/stop cycles happen constantly, so the race fires almost every time and the source never stays up.

### Steps to reproduce

1. Add a Nest source over WebRTC: `nest:?client_id=...&client_secret=...&refresh_token=...&project_id=...&device_id=...&protocols=WEB_RTC`
2. Consume it (e.g. via RTSP/WebRTC). The producer starts, then panics within a few extend cycles and crash-loops.

### Suggested fix

Capture the timer channel locally so the goroutine never dereferences `a.extendTimer`:

```go
a.extendTimer = time.NewTimer(time.Until(a.StreamExpiresAt) - time.Minute)
ch := a.extendTimer.C
go func() {
    <-ch
    ...
}()
```

A done-channel variant additionally avoids a goroutine leak when the timer is stopped before it fires (the goroutine would otherwise block on the channel forever). Happy to open a PR.

### Related issues

- #1827 — Nest RTSP reconnects at ~5 min because the refreshed streamToken isn't applied. Same stream-extension path: this panic is in `StartExtendStreamTimer`, the timer that drives that pre-expiry refresh — related symptom in the same subsystem (reconnect vs. crash). See also #1253.
- #2311 — Google SDM `GenerateWebRtcStream` returns 400 in 2026 (tightened offerSdp validation). Same `nest:` module and same mid-2026 Google breakage wave, but a distinct failure mode (WebRTC offer rejected at the API vs. an internal nil-pointer) — almost certainly a separate root cause.

FariAmrid:
## Problem

Nest WebRTC/RTSP streams die shortly after start (often within a minute) and never recover. With several cameras on the same credentials it is worse: only the most recently started stream can be extended at all. This became much more visible recently as Nest media sessions got shorter-lived (see #2319), but the underlying defects are older and likely also explain long-standing reports like #2108 and #723.

## Root causes (`pkg/nest/api.go`)

1. **All streams with the same credentials share one `*API` instance.** `NewAPI` returns the cached object itself, and `ExchangeSDP` / `GenerateRtspStream` store per-session state (`StreamProjectID`, `StreamDeviceID`, `StreamSessionID`, `StreamExpiresAt`, …) on it. Every new stream overwrites the previous stream's session, so `ExtendStream` targets whichever camera connected last — extensions for every other camera are lost. The `extendTimer != nil` guard also means only one timer ever exists for all streams.

2. **The extend timer is single-shot and never re-arms.** `StartExtendStreamTimer` schedules one extend at `expiresAt − 1 min` and the goroutine exits afterwards, so a session survives at most two expiry windows even when everything works. When `expiresAt` is less than a minute away, the computed delay is ≤ 0, the timer fires immediately, and the only extend is spent right away — the session then dies at the next expiry.

3. **Race + goroutine leak (#2319).** `StopExtendStreamTimer` sets `a.extendTimer = nil` while the goroutine dereferences `a.extendTimer.C` → nil-pointer panic. And because `Timer.Stop` does not close the channel, a stopped timer leaves that goroutine blocked forever — one leaked goroutine per stream start/stop cycle.

## Fix

- `NewAPI` returns a fresh per-caller `API` carrying the cached token (token caching behavior is unchanged), so each stream owns its session state and extend loop. `refreshToken` now uses a stored credentials key instead of the reverse cache lookup by token — that lookup also failed whenever the cached token had rotated.
- `StartExtendStreamTimer` starts a loop that extends the session 30 s before every expiry (minimum 5 s delay), re-arms after each success using the refreshed `expiresAt`, and retries a failed extend up to 3 times, 2 s apart, before giving up.
- The loop is controlled by a stop channel guarded by a mutex: no shared timer pointer, no nil dereference, no leaked goroutine. `StopExtendStreamTimer` stays idempotent.

No logging added, consistent with the logging removal in #1669.

## Testing

Equivalent patch (same logic, plus debug logging) running in production since 2026-07-01 on linux/arm64 Docker with 4 WebRTC-only cameras (Nest Cam battery/wired gen 2 + Nest Doorbell, `supportedProtocols=[WEB_RTC]`), feeding ffmpeg recorders through the RTSP restream:

- **before:** every producer died ~45–60 s after start; consumers received an initial ~2 s burst of video and then nothing (`Could not find codec parameters`, empty recordings);
- **after:** sessions are created with a 5-minute expiry and streams survive well past the previous death point; motion-event recordings complete and save correctly.

Fixes #2319

### DIFF
--- a/pkg/nest/api.go
+++ b/pkg/nest/api.go
@@ -26,7 +26,10 @@ type API struct {
 	StreamToken          string
 	StreamExtensionToken string
 
-	extendTimer *time.Timer
+	credsKey string
+
+	extendMu   sync.Mutex
+	extendStop chan struct{}
 }
 
 type Auth struct {
@@ -49,8 +52,10 @@ func NewAPI(clientID, clientSecret, refreshToken string) (*API, error) {
 	key := clientID + ":" + clientSecret + ":" + refreshToken
 	now := time.Now()
 
+	// Each caller gets its own API instance (sharing only the cached token),
+	// so concurrent streams don't overwrite each other's session state.
 	if api := cache[key]; api != nil && now.Before(api.ExpiresAt) {
-		return api, nil
+		return &API{Token: api.Token, ExpiresAt: api.ExpiresAt, credsKey: key}, nil
 	}
 
 	data := url.Values{
@@ -89,7 +94,7 @@ func NewAPI(clientID, clientSecret, refreshToken string) (*API, error) {
 
 	cache[key] = api
 
-	return api, nil
+	return &API{Token: api.Token, ExpiresAt: api.ExpiresAt, credsKey: key}, nil
 }
 
 func (a *API) GetDevices(projectID string) ([]DeviceInfo, error) {
@@ -228,23 +233,12 @@ func (a *API) ExchangeSDP(projectID, deviceID, offer string) (string, error) {
 }
 
 func (a *API) refreshToken() error {
-	// Get the cached API with matching token to get credentials
-	var refreshKey string
-	cacheMu.Lock()
-	for key, api := range cache {
-		if api.Token == a.Token {
-			refreshKey = key
-			break
-		}
-	}
-	cacheMu.Unlock()
-
-	if refreshKey == "" {
+	if a.credsKey == "" {
 		return errors.New("nest: unable to find cached credentials")
 	}
 
 	// Parse credentials from cache key
-	parts := strings.Split(refreshKey, ":")
+	parts := strings.SplitN(a.credsKey, ":", 3)
 	if len(parts) != 3 {
 		return errors.New("nest: invalid cache key format")
 	}
@@ -464,23 +458,74 @@ type Device struct {
 	} `json:"parentRelations"`
 }
 
+// StartExtendStreamTimer keeps the stream session alive by extending it
+// before every expiry until StopExtendStreamTimer is called. Sessions can
+// be short-lived, so a single extend at expiry minus one minute is not
+// enough: the delay may be negative and the session still dies one expiry
+// later.
 func (a *API) StartExtendStreamTimer() {
-	if a.extendTimer != nil {
+	a.extendMu.Lock()
+	defer a.extendMu.Unlock()
+
+	if a.extendStop != nil {
 		return
 	}
 
-	a.extendTimer = time.NewTimer(time.Until(a.StreamExpiresAt) - time.Minute)
-	go func() {
-		<-a.extendTimer.C
-		if err := a.ExtendStream(); err != nil {
-			return
-		}
-	}()
+	stop := make(chan struct{})
+	a.extendStop = stop
+
+	go a.extendLoop(stop)
 }
 
 func (a *API) StopExtendStreamTimer() {
-	if a.extendTimer != nil {
-		a.extendTimer.Stop()
-		a.extendTimer = nil
+	a.extendMu.Lock()
+	defer a.extendMu.Unlock()
+
+	if a.extendStop != nil {
+		close(a.extendStop)
+		a.extendStop = nil
+	}
+}
+
+func (a *API) extendLoop(stop chan struct{}) {
+	const margin = 30 * time.Second
+	const minDelay = 5 * time.Second
+	const retryDelay = 2 * time.Second
+	const maxFailures = 3
+
+	for {
+		delay := time.Until(a.StreamExpiresAt) - margin
+		if delay < minDelay {
+			delay = minDelay
+		}
+		if !sleep(stop, delay) {
+			return
+		}
+
+		failures := 0
+		for {
+			if err := a.ExtendStream(); err == nil {
+				break
+			}
+
+			failures++
+			if failures >= maxFailures {
+				return
+			}
+			if !sleep(stop, retryDelay) {
+				return
+			}
+		}
+	}
+}
+
+func sleep(stop chan struct{}, d time.Duration) bool {
+	timer := time.NewTimer(d)
+	defer timer.Stop()
+	select {
+	case <-stop:
+		return false
+	case <-timer.C:
+		return true
 	}
 }


night4me:
PR #2335 completely fixes the Nest session expiration issue on my system. Thank you!

I noticed one minor startup behavior.

After a cold start of Frigate, the initial ffmpeg processes sometimes start before the Nest producer has received the first video keyframe, resulting in one expected startup error:

Could not find codec parameters for stream 0 (Video: h264, none): unspecified size

The watchdog immediately restarts ffmpeg, after which both Nest cameras run normally for hours without any further issues.

So this is no longer a stability problem—only a cosmetic startup race. It might be possible to eliminate the initial log messages by delaying the RTSP restream until the producer is fully initialized, or by waiting until the first keyframe has been received before Frigate attempts to consume the stream.

ajplotkin:
For anyone landing here: **#2351 fixes the same two defects** — the shared `*API` instance clobbering per-stream session state, and the one-shot extend timer that guarantees death at ~10 minutes. Two independent diagnoses of one bug, which is worth knowing when deciding what to merge.

I've been running the equivalent of both fixes in a fork since mid-July across four Nest cameras on one Device Access project. Before them, streams died on a ~5–10 minute cycle and only the most recently connected camera got extended at all; after, sessions run for hours. So this is real and the diagnosis in both PRs matches what I see.

@AlexxIT, the practical question is only which of the two to take; they overlap almost entirely.

One caveat from running it: fixing the extension does **not** eliminate producer re-dials. I still see 1–3 per day per camera with no logged cause — and because `producer.go` only logs when `conn.Start()` returns an *error*, a clean end is completely silent (the next line is just `retry=0`). So if anyone tests either PR and still sees periodic drops, the extend timer may not be the remaining cause, and there's currently no way to tell from the log.

zyntax9999:
<html><body>
<html><head></head><body><p>Confirming this fixes the panic, with about a day of runtime on arm64.</p>
<p><strong>Before (v1.9.14, revision b5948cf)</strong></p>
<p>Repeated crashes with a Nest source:</p>
<pre><code>panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0x6aadd4]

goroutine 138 [running]:
github.com/AlexxIT/go2rtc/pkg/nest.(*API).StartExtendStreamTimer.func1()
	github.com/AlexxIT/go2rtc/pkg/nest/api.go:474 +0x24
created by github.com/AlexxIT/go2rtc/pkg/nest.(*API).StartExtendStreamTimer in goroutine 137
	github.com/AlexxIT/go2rtc/pkg/nest/api.go:473 +0xac
</code></pre>
<p>Two of these within thirteen minutes at one point, and generally one every hour
or two. Each one takes the process down and the consumer loses video until
go2rtc restarts and reconnects.</p>
<p><strong>After (this PR, 56fa5fd)</strong></p>
<p>Built for linux/arm64 from <code>docker/Dockerfile</code> and run in the same place: no
panics at all.</p>
<p>The second effect was the more visible one. My consumer is a proxy that presents
the stream to UniFi Protect as an ONVIF camera, and I log every RTSP connection
it receives:</p>

  | reconnections
-- | --
v1.9.14 | ~7 per minute, sustained
this PR | 2 in 87 minutes


<p>That matches the re-arming fix rather than the nil check: sessions now survive
instead of expiring and being rebuilt constantly. It matters more than it might
look, because the SDM API limits cameras to 100 commands per hour, and every
stream start or extension counts against it. At the old rate that budget was
exhausted in about fifteen minutes.</p>
<p><strong>Setup</strong></p>
<ul>
<li>go2rtc in a Home Assistant add-on, Raspberry Pi 4, HAOS 18.2</li>
<li>Nest Cam (battery), mains powered, via the SDM API</li>
<li>One consumer pulling continuously (UniFi Protect through an ONVIF proxy), plus
a UniFi AI Port doing detections on the same stream</li>
</ul>
<p>Happy to run anything else against it if that would help.</p></body></html>
</body>
</html>

ajplotkin:
We hit the same two problems on a Nest SDM setup (four cameras, one credential) and ended up at the same structural conclusions independently — per-stream `API` instances rather than the shared cached one, and an extend loop that re-arms instead of firing once. Our version has been running continuously for months on a fork of 1.9.14; no panics, and stream sessions hold.

Since the approaches converged, two things ours does that this PR does not, both of which bit us:

**1. The OAuth token has to be refreshed inside the extend loop.**

`ExtendStream()` sends `Authorization: Bearer a.Token` and returns `nest: wrong status: …` on non-200 — it never refreshes. The Google OAuth token lives about an hour; a held session outlives it. So once the token expires, every `ExtendStream()` in `extendLoop` 401s, burns `maxFailures` at `retryDelay` (~6 s), and the loop returns. The session then expires and the stream stalls — roughly an hour into any long-held stream.

We refresh before extending when the token is close to expiry:

```go
case <-timer.C:
    // The OAuth token lives ~1 hour, sessions can live longer
    if time.Now().After(a.ExpiresAt.Add(-30 * time.Second)) {
        if err := a.refreshToken(); err != nil {
            if fails++; fails <= 3 {
                timer.Reset(15 * time.Second)
                continue
            }
            return
        }
    }
    if err := a.ExtendStream(); err != nil { ... }
```

**2. `http.Client{Timeout: time.Second * 5000}` is 5000 *seconds*.**

That appears six times in `pkg/nest/api.go` and looks like it was meant to be milliseconds. It matters here specifically: a hung SDM request blocks `extendLoop` for ~83 minutes, so `maxFailures`/`retryDelay` never engage — the loop isn't retrying, it's parked. We set all six to 10 s. It's outside this PR's scope, but it undercuts the retry design the PR adds.

One difference that is just a tuning choice rather than a defect: we clamp the re-arm floor at 30 s rather than 5 s, and space retries 15 s rather than 2 s, on the theory that SDM's 100 commands/hour/device budget is the scarcer resource. Either is defensible.

Happy to open a PR for the token refresh against this branch or after it merges, whichever is more useful — or to just leave the notes here if you'd rather keep the change tight.

Issue #1290, opened 2024-08-02
Allow to add go2rtc as a network camera in Windows


felipecrs:
Windows has this feature of network cameras:

![image](https://github.com/user-attachments/assets/9139587d-5d1e-4c08-9289-7de5b4b0cbdb)

It would be amazing if Windows could recognize go2rtc as a network camera.

But that would require go2rtc to implement the ONVIF WS-Discovery protocol:

https://learn.microsoft.com/en-us/windows-hardware/drivers/stream/network-cameras#discovery

Do you think that would be possible?

My use case is very niche:

When running Microsoft Teams under Azure Virtual Desktop, it will only recognize cameras that are "physically" connected to the client computer and not virtual cameras like OBS Studio or NVIDIA Broadcast.

With go2rtc running on Windows, I can input the camera with `ffmpeg`, and then if I was able to add go2rtc back to Windows as a network camera, I would be able to use it in Teams.

felipecrs:
Another nice use case would be to allow using any arbitrary camera from go2rtc, like cameras plugged in a different computer, in any video conferencing application like Teams or Zoom.

memphisdj:
I too am interested in having go2rtc act as ONVIF compliant source and to expose the streams in it over ONVIF. My use case is to add a stream from go2rtc into Hikvision DVR, which allows only ONVIF IP Cameras to be added.

AlexxIT:
ONVIF server supported long time ago

felipecrs:
https://github.com/AlexxIT/go2rtc?tab=readme-ov-file#source-onvif:~:text=to%20go2rtc%20config-,Hass%20%3E%20Settings,-%3E%20Integrations%20%3E%20Add%20Integration

memphisdj:
> https://github.com/AlexxIT/go2rtc?tab=readme-ov-file#source-onvif:~:text=to%20go2rtc%20config-,Hass%20%3E%20Settings,-%3E%20Integrations%20%3E%20Add%20Integration

In first place sorry for hijacking your issue, but the doc you linked is if i would like to interogate my camera from go2rtc through onvif. That is not what i want. I mean each camera in go2rtc to be able to be added to another device as onvif source for it. As pictured above: IP Cam already in go2rtc -> go2rtc exposing its stream info as onvif source -> Hikvision DVR (or any device for that matter).

felipecrs:
Ops, sorry, wrong link. I meant this:

[...]

PR #2336, opened 2026-07-06
feat: add WS-Discovery probe responder

lrh3321:
Add a WS-Discovery module that listens for multicast Probe requests and responds with ONVIF device service ProbeMatch messages.

Also register the module at startup and guard concurrent probe handling with a fixed worker limit and panic recovery.

#1290


AlexxIT:
This is simply a function of the ONVIF module. There's no point in creating a separate module for it.

Issue #908, opened 2024-01-27
Stream dies with EOF error every 10 minutes to the second


andonevris:
I have a stream that dies every 10 minutes exactly from when go2rtc launches.  It is a Tuya camera with its stream exposed using the "expose_camera_stream" hacs integration.

Refreshing the webpage brings the stream back up.  

As you can see from the logs it happens at precise 10 minute intervals

```
1/27/2024, 12:32:03 PM	info	go2rtc version 1.8.5 linux/amd64
1/27/2024, 12:32:03 PM	info	[rtsp] listen addr=:8554
1/27/2024, 12:32:03 PM	info	[api] listen addr=:1984
1/27/2024, 12:32:03 PM	info	[webrtc] listen addr=:8555
1/27/2024, 12:42:10 PM	warn	undefined error=EOF url=echo:bash /config/custom_components/expose_camera_stream_source/get_stream.sh camera.xxxxxxx caller=github.com/AlexxIT/go2rtc/internal/streams/producer.go:171
1/27/2024, 12:52:10 PM	warn	undefined error=EOF url=echo:bash /config/custom_components/expose_camera_stream_source/get_stream.sh camera.xxxxxxx caller=github.com/AlexxIT/go2rtc/internal/streams/producer.go:171
1/27/2024, 1:02:11 PM	warn	undefined error=EOF url=echo:bash /config/custom_components/expose_camera_stream_source/get_stream.sh camera.xxxxxxx caller=github.com/AlexxIT/go2rtc/internal/streams/producer.go:171
```

I have other streams from different cameras that don't exhibit this issue.

Using the Tuya app I can watch the stream indefinitely so it's not Tuya cutting the stream off.

AlexxIT:
Why do you think Tuya app uses same protocol?

andonevris:
> Why do you think Tuya app uses same protocol?

I don't, I'm just putting the info there for you in case it helps with diagnosis. 

I considered it could well be Tuya but the stream restarts at 10 minute intervals from when go2rtc first launches so thought it could also be possible its go2rtc somehow.

AlexxIT:
You do.

> Using the Tuya app I can watch the stream indefinitely so it's not Tuya cutting the stream off.

I think Tuya API from Hass integration, that you using, provide link only for 10 minutes. It's usual situation. Nest gives link only for 5 minutes

andonevris:
> You do.
> 
[...]

PR #2351, opened 2026-07-13
nest: keep extending stream sessions and stop sharing session state between streams

tillo:
Hi Alex, thank you for go2rtc — it has been rock solid for me across many camera types. I believe I found the reason Nest streams drop every 5–10 minutes (#723, #908) and would like to offer a fix.

**What I observed.** Two Nest cameras (same Device Access project) on the `nest:` source reconnect every ~5 or ~10 minutes, both at the same moment. Each renegotiation starts a stream with fresh H264 parameter sets, so downstream RTSP consumers (Frigate's ffmpeg in my case) wedge on `non-existing PPS 0 referenced / no frame!` until their watchdog restarts them — about 70 short outages a day.

**Cause 1 — the extension timer is one-shot.** `StartExtendStreamTimer` arms a single timer, calls `ExtendStream` once and the goroutine exits. Since Google grants ~5-minute sessions, every stream inevitably dies at ~10 minutes (initial 5 + one extension), even though `ExtendStream` already stores the new `expiresAt` from the response. Fixed by looping: after each successful extension the timer is re-armed from the new expiry, and the OAuth token (1 h lifetime) is refreshed shortly before it expires so long-lived sessions keep extending past the first hour. A stop channel ends the loop cleanly from `StopExtendStreamTimer`.

**Cause 2 — session state is shared between streams.** `NewAPI` caches the `*API` object per credentials, but `ExchangeSDP`/`ExtendStream` keep per-stream session state (`StreamSessionID`, `StreamExpiresAt`, `StreamToken`, …) on that same object. With several cameras on one project the sessions overwrite each other, the `if a.extendTimer != nil` guard leaves only one timer for all of them (extending only the last-connected camera's session), and `Stop()` on any one stream kills that shared timer for the rest. Fixed by caching only the OAuth token: every `Dial` now gets its own `API` instance (the token is still reused across streams, so no extra token requests). `refreshToken` now uses the stored credentials key instead of scanning the cache for a matching token value, which also survives the token being rotated by another instance.

**Result.** With both fixes my two cameras have held a single uninterrupted session each (previously: a renegotiation every 5–10 minutes, matching the reports in #723/#908 to the second).

Happy to adjust anything — naming, commit message, or dropping the token-refresh part if you would prefer it as a separate change.

### DIFF
--- a/pkg/nest/api.go
+++ b/pkg/nest/api.go
@@ -26,7 +26,10 @@ type API struct {
 	StreamToken          string
 	StreamExtensionToken string
 
+	key string // credentials cache key, used to refresh the OAuth token
+
 	extendTimer *time.Timer
+	extendStop  chan struct{}
 }
 
 type Auth struct {
@@ -49,8 +52,12 @@ func NewAPI(clientID, clientSecret, refreshToken string) (*API, error) {
 	key := clientID + ":" + clientSecret + ":" + refreshToken
 	now := time.Now()
 
+	// The cache only stores the OAuth token. Each caller gets its own API
+	// instance, because the Stream* fields hold per-stream session state -
+	// multiple cameras sharing one instance overwrite each other's session
+	// and only the last one gets extended.
 	if api := cache[key]; api != nil && now.Before(api.ExpiresAt) {
-		return api, nil
+		return &API{Token: api.Token, ExpiresAt: api.ExpiresAt, key: key}, nil
 	}
 
 	data := url.Values{
@@ -89,7 +96,7 @@ func NewAPI(clientID, clientSecret, refreshToken string) (*API, error) {
 
 	cache[key] = api
 
-	return api, nil
+	return &API{Token: api.Token, ExpiresAt: api.ExpiresAt, key: key}, nil
 }
 
 func (a *API) GetDevices(projectID string) ([]DeviceInfo, error) {
@@ -228,23 +235,12 @@ func (a *API) ExchangeSDP(projectID, deviceID, offer string) (string, error) {
 }
 
 func (a *API) refreshToken() error {
-	// Get the cached API with matching token to get credentials
-	var refreshKey string
-	cacheMu.Lock()
-	for key, api := range cache {
-		if api.Token == a.Token {
-			refreshKey = key
-			break
-		}
-	}
-	cacheMu.Unlock()
-
-	if refreshKey == "" {
+	if a.key == "" {
 		return errors.New("nest: unable to find cached credentials")
 	}
 
-	// Parse credentials from cache key
-	parts := strings.Split(refreshKey, ":")
+	// Parse credentials from the cache key
+	parts := strings.SplitN(a.key, ":", 3)
 	if len(parts) != 3 {
 		return errors.New("nest: invalid cache key format")
 	}
@@ -469,11 +465,30 @@ func (a *API) StartExtendStreamTimer() {
 		return
 	}
 
-	a.extendTimer = time.NewTimer(time.Until(a.StreamExpiresAt) - time.Minute)
+	// Google expires sessions after ~5 minutes; each successful extension
+	// returns a new expiresAt, so keep extending until the stream stops.
+	timer := time.NewTimer(time.Until(a.StreamExpiresAt) - time.Minute)
+	stop := make(chan struct{})
+	a.extendTimer = timer
+	a.extendStop = stop
+
 	go func() {
-		<-a.extendTimer.C
-		if err := a.ExtendStream(); err != nil {
-			return
+		for {
+			select {
+			case <-timer.C:
+				// The OAuth token lives ~1 hour, sessions can live longer
+				if time.Now().After(a.ExpiresAt.Add(-30 * time.Second)) {
+					if err := a.refreshToken(); err != nil {
+						return
+					}
+				}
+				if err := a.ExtendStream(); err != nil {
+					return
+				}
+				timer.Reset(time.Until(a.StreamExpiresAt) - time.Minute)
+			case <-stop:
+				return
+			}
 		}
 	}()
 }
@@ -482,5 +497,9 @@ func (a *API) StopExtendStreamTimer() {
 	if a.extendTimer != nil {
 		a.extendTimer.Stop()
 		a.extendTimer = nil
+	}
+	if a.extendStop != nil {
+		close(a.extendStop)
+		a.extendStop = nil
 	}
 }


benholtz:
Confirming this fix in production. Setup: 4 Nest cameras on the same Device Access credentials, go2rtc 1.9.14 feeding Frigate over RTSP, cherry-picked this PR onto v1.9.14.

Before: exactly the symptom described. Only the most recently started stream kept its session extended, the others renegotiated every ~5 minutes, and Frigate's ffmpeg consumers wedged on `non-existing PPS 0` after each renegotiation. Roughly one forced ffmpeg restart every 4 minutes per affected camera, around the clock.

After 6+ hours soak: zero session-expiry renegotiations, zero PPS errors from this cause, all 4 streams stable. @AlexxIT would love to see this merged.

One adjacent note for anyone with multiple RTSP consumers per nest stream: SPS/PPS still only arrive at dial time (they're in Google's SDP but not repeated in-band), so a consumer that attaches mid-stream never gets parameter sets. Worked around it by giving the second consumer an ffmpeg re-encode child. Wiring sprop-parameter-sets from the Nest SDP into go2rtc's RTSP SDP would fix that properly.

ajplotkin:
Running this in production for a few weeks serving 3–4 Nest cameras through a single warm go2rtc stream to Homebridge/HomeKit — this PR fixes a real bug for me. Before it, every session died at the ~10-minute mark (one-shot timer) and the shared `*API` meant only the last-connected camera actually got extended; the others renegotiated constantly (`non-existing PPS 0 referenced / no frame!`). Looping the timer + per-`Dial` instance both fixed it. Thanks for this.

One thing I hit in production that this PR doesn't cover: the extend goroutine treats **any** `ExtendWebRtcStream`/token-refresh error as terminal and returns, so a single transient blip (a 5xx from Google, a momentary network drop, an OAuth refresh that times out) permanently stops extension — and because WebRTC media keeps flowing for a while after, `producer.go` never sees a connection error and never reconnects. The stream just silently freezes ~10 min later. In my fork I made the loop retry transient extend/refresh failures a few times (~15s apart) before giving up, and clamped the timer reset to a floor (≥30s) so a zero/past `expiresAt` from Google can't hot-loop `ExtendStream` into a 429. Happy to send that as a follow-up on top of this if it's useful — but the core fix here is the important part.

bhamiltoncx:
@AlexxIT , I'd love to see this land, this is biting me every few minutes with a bunch of Nest cameras.

donparlor:
I’ve been testing a follow-up on top of #2351 for the remaining failure mode discussed here: after an extension/refresh failure exhausts its retries, simply returning from the extension goroutine can still leave the WebRTC transport alive. Media may keep flowing temporarily and then freeze when the Nest session expires, while Producer.worker() never sees Start() return and therefore never reconnects.

My current patch escalates an unrecoverable Nest extension failure to the owning transport: WebRTC is closed / RTSP is stopped, which makes Start() return and lets the existing Producer.reconnect() path create a fresh Nest session.

It also adds:

retry classification for network/deadline errors and HTTP 408/409/429/5xx;
a forced OAuth refresh on HTTP 401 that bypasses the still-valid local token cache;
15s HTTP operation timeouts;
context-cancellable retry/backoff;
a mutex + done-channel runner lifecycle so normal Stop and fatal recovery are explicitly serialized and idempotent.

I added tests for timer boundaries, retries, forced OAuth refresh, cancellation, concurrent stops and both Stop-vs-fatal orderings. go test -race ./pkg/nest is clean, including repeated Stop/fatal race tests.

I deployed the linux/arm64 build in production today on a Nest Doorbell feeding Frigate. Initial state is healthy (one producer, H264 present). I’m doing a 72-hour production soak before claiming runtime validation and will report the result here.

If this proves stable, I’d be happy to turn the recovery portion into a focused follow-up PR rather than duplicating the existing #2351/#2335 work.

ajplotkin:
@donparlor the escalation is the right direction. Leaving the transport up after
the extension goroutine gives up is exactly how you get media that flows and then
freezes with nothing calling `reconnect()`. Two things about where recovery
lands.

`Producer.reconnect()` can sever the consumers you carried over, without a log
line. The loop at `internal/streams/producer.go:206` matches at most one receiver per
media and `break`s. If `p.receivers` holds two H264 receivers for the same media,
the second is never `Replace()`d, and `p.conn.Stop()` at :238 closes it and
detaches its consumer senders. No log line at any level. Filed as #2387.

Two H264 receivers for one media is not a corner case on Nest, it is #2389:
Google answers with two payload types and streams on the second, so a consumer
attached at dial time binds to the PT 96 receiver while the transmitted PT 98
creates a second one. A cold consumer and a warm consumer end up on different
receivers, and a reconnect keeps whichever the media matches first.

It is the `break` rather than the two `continue`s above it. Neither of those can
fire here: `webrtc.Conn.GetTrack` returns a nil error on every
non-panicking branch, and a fresh Nest session answers with the same H264/Opus
medias, so `MatchCodec` matches like-for-like.

#2480 is what makes your recovery path safe, so it should land first. The receiver
`reconnect()` pre-seeds is created through `Conn.GetTrack` with the same `*Media`
that `OnTrack` later resolves, so #2480's reuse loop adopts the transmitted
payload type onto that receiver and the carried-over consumers are fed. It also
stops a second receiver being created for a media, which removes the precondition
for the `break` severing anything.

Without it, every reconnect re-rolls #2389: the pre-seed runs before the first
RTP packet, `OnTrack` misses the pointer, a second receiver appears, and the
consumers you carried over sit on the one that is never fed. Your change makes
that path run far more often than it does today.

On #2351 alone, each forced recovery should cost one
Frigate ffmpeg restart: no video on the carried-over consumer, the watchdog
fires, the replacement attaches warm and binds to the fed receiver. If you see
that, it is #2389 rather than your patch. With #2480 on top the same recovery
should carry the consumer over with no restart.

Frozen `bytes_send` alone will not tell you which happened.
`consumers[].senders[].parent` will: a receiver that exists but has no bytes when
starved, an id that no longer appears in `producers[].receivers` when severed.

I would force an extension failure rather than wait for a natural one. And
"one producer, H264 present" was your initial-state check rather than a soak
criterion, so you may already have this covered.

I have been running a Nest-gated variant of the receiver-reuse fix in production
since late July, on four cameras, which is what surfaced #2387 and #2389.

donparlor:
Thanks — this is very useful, especially the correction on the reconnect path.

One important detail I hadn’t mentioned: my 2ca8a77 build is not based on plain master. The base already includes #2375, so the Nest/WebRTC receiver-reuse behavior is present before my extension-recovery patch.

I checked the live graph after deployment:

producer id=163
H264 receiver id=164 bytes=268221872 childs=[165]
consumer sender id=165 parent=164 bytes=268221872

There is currently a single H264 receiver, and the consumer sender is parented to that same fed receiver. So the initial graph is not in either the #2389 starved-receiver state or the #2387 severed-parent state.

I agree that #2480 is still the cleaner/s safer evolution of #2375 because it adds the codec compatibility guard and adopts the actually transmitted codec.

Your point about validating an actual reconnect is well taken. A 72-hour soak alone only proves stability if a fatal extension failure actually occurs during the window. I’m considering a controlled reconnect/failure test, but I want to avoid introducing a fault-injection change into the production build just to manufacture the condition.

If I do force one, I’ll capture producer.id, receiver IDs/children, sender parent, bytes_recv, receiver bytes and sender/consumer bytes immediately before and after recovery.

ajplotkin:
Your base has #2375, so the sever I described does not apply. That fix already
returns the existing receiver for the media, so `OnTrack` cannot create a second
one for Nest's PT 96/98 pair, `p.receivers` never holds two live H264 receivers
for one media, and the `break` at `producer.go:222` has nothing to orphan. Your
single fed receiver with the sender parented to it is the expected state.

What #2480 adds on top of #2375:

- the codec guard. `track.Media == media` alone returns an H264 receiver to a VP8
  request on a media carrying both. `pkg/webrtc/api.go` registers H264 for video
  and Opus for audio on separate medias, so Nest does not reach it.
- `track.Codec = codec`. Without the adoption `track.Codec` stays PT 96 while the
  wire carries PT 98, so the pointer-keyed cache in
  `internal/streams/producer.go` misses for a consumer attaching after
  `media.Codecs` is narrowed, and `p.receivers` gains a second entry pointing at
  the same receiver. That duplicate is inert on reconnect: the first entry is
  `Replace()`d, `MoveNode` moves every child and nils the source's, and
  `Connection.Stop()` iterates the conn's own receiver list, which has no
  duplicate. The visible effect is a stale `receivers[].codec`.

For forcing the condition without a fault-injection build: `pkg/nest` calls
`smartdevicemanagement.googleapis.com` for `ExtendStream`. Match the SNI rather
than the address, since that host and `www.googleapis.com` resolve to the same
front-end addresses and an address rule would take the token endpoint with it.
Hostname rules also resolve once at insert, so DNS rotation during the retry
window can let a retry through.

```
# host netns: OUTPUT. bridge-networked container: DOCKER-USER.
sudo iptables  -I OUTPUT 1 -p tcp --dport 443 -m string --algo bm \
  --string "smartdevicemanagement.googleapis.com" -j REJECT --reject-with tcp-reset
sudo ip6tables -I OUTPUT 1 -p tcp --dport 443 -m string --algo bm \
  --string "smartdevicemanagement.googleapis.com" -j REJECT --reject-with tcp-reset
# pull both as soon as the fatal path fires: the reconnect dials the same host,
# and its backoff reaches 60s after 20 failed attempts
sudo iptables -D OUTPUT 1 && sudo ip6tables -D OUTPUT 1
```

Go's `dialParallel` starts the other address family as soon as the first fails,
so a v4-only rule gets bypassed over IPv6 and the failure never fires. REJECT
gives an immediate ECONNREFUSED, which hits your network-error classification
without waiting out the 15s timeout. The "after" capture has to come after the
rules are pulled.

Blocking the token host will not exercise the forced-refresh-on-401 branch. It
only makes the refresh itself fail with a network error, and nothing at the
network layer makes SDM answer 401.

One correction to my capture suggestion: `/api/streams` does not expose the
transmitted payload type anywhere. `Codec.MarshalJSON` emits name, type, profile,
level, sample rate and channels. On a #2375 build `receivers[].codec` also keeps
the pre-seed codec after a reconnect. Receiver `bytes` and the sender `parent`
are the two that carry the answer.

donparlor:
Thanks — the forced test produced a useful result.

I blocked SDM for both IPv4/IPv6 from the Frigate container while keeping the running go2rtc process intact. The existing producer was 163 with one fed H264 receiver and a correctly parented consumer sender.

During the block, media stopped advancing at ~14:03:08. Frigate’s ffmpeg watchdog then entered ~11s restart cycles. After the SDM block was removed at ~14:04:23, there was one final failed ffmpeg cycle at 14:04:31, after which the stream recovered automatically.

go2rtc itself never restarted — same PID 129, same 2ca8a77 process throughout.

Post-recovery the producer changed from 163 to 179, with:

producer 179
H264 receiver 180, childs=[188]
sender 188, parent=180
consumer 176

Over a 5s sample after recovery:

producer bytes_recv  +669040
receiver bytes       +638482
sender bytes         +638482
consumer bytes_send  +668152

So the recovered graph is healthy: one fed H264 receiver, sender parented to that receiver, and end-to-end counters advancing.

One caveat: Frigate’s ffmpeg consumer was restarted during the outage (consumer 162 → 176), so this demonstrates successful automatic system recovery and a fresh Nest producer, but it does not by itself prove that the consumer survived the fatal transport escalation transparently. Since I intentionally avoided logging the fatal Nest error to prevent the raw source URL/credentials from reaching Producer.worker()’s warning path, I also don’t have a direct timestamp for the fatal callback itself.

I’m going to leave this build running now and use the 72h period as a stability soak rather than repeat the disruptive fault injection.

ajplotkin:
The producer 163 -> 179 graph is healthy, and your caveat is the one that
matters: consumer 162 -> 176 means Frigate rebuilt its side, so nothing there
speaks to whether an existing consumer survives the escalation.

A second consumer with no watchdog of its own settles that without repeating the
injection. From inside the Frigate container:

```
curl -sN -o /dev/null 'http://127.0.0.1:1984/api/stream.mp4?src=<stream>' &
```

The mp4 handler stops only on request-context cancellation or an explicit
`duration=`, and `pkg/mp4/consumer.go` has no timers, so it sits through a media
gap. Record its consumer id, its sender id, that sender's `parent`, and confirm
the sender's `bytes` keep advancing after any recovery.

It survives a producer reconnect if `reconnect()` carries it:
`receiver.Replace(track)` is `MoveNode`, which moves every child to the new
receiver and sets `child.parent`, so the sender object persists and its `parent`
changes from the old receiver id to the new one. A severed consumer instead keeps
a `parent` that no longer appears in `producers[].receivers`, with `bytes`
frozen. Left attached for the 72h soak it records the answer if a fatal fires on
its own.

For the timestamp, log the error at the point in `pkg/nest` where you decide the
failure is unrecoverable, without the source string. What reaches the line 170
warning is only whatever `Start()` returns, so your own line gives the
fatal-callback time while that path stays quiet. Nothing under `pkg/` logs today,
so treat it as a soak-build line rather than part of the PR.

One detail to pin down when you have the Frigate log: whether the final failed
ffmpeg cycle at 14:04:31 was the watchdog racing a producer that had only just
re-dialed (the block lifted at 14:04:23, and a Nest dial is an OAuth call plus an
SDP exchange) or a second failure. The ~11s rhythm is Frigate's
`ffmpeg.retry_interval`, default 10s; go2rtc's reconnect runs 1s, 5s, 10s then
60s and never restarts ffmpeg, so it is not in that cycle.

donparlor:
Thanks — I went ahead and completed the controlled test with the passive MP4 consumer attached, and it gave a pretty definitive result.

Baseline after the unrelated HAOS power outage/restart was:

Nest producer: 56
H264 receiver: 57
MP4 consumer: 60
H264 sender: 61
sender parent: 57

Before forcing anything, an interesting natural failure occurred: H264 stopped completely while Opus continued advancing. I observed that state for another 10 minutes:

producer: 56
H264 receiver: 57
MP4 consumer: 60
MP4 sender: 61

H264 bytes:      unchanged
MP4 sender bytes: unchanged
Opus bytes:      continued increasing

So the PeerConnection remained alive enough to carry audio, but video was dead, and no Producer.reconnect() occurred.

I then used the controlled SDM failure specifically to test your consumer-transfer hypothesis. The next ExtendWebRtcStream attempt was blocked, the retry/fatal path was allowed to run, SDM was released, and the producer reconnected.

Result:

producer:      56 -> 566
H264 receiver: 57 -> 567

MP4 consumer:  60 -> 60
MP4 sender:    61 -> 61
sender parent: 57 -> 567

Post-reconnect 10s deltas:

producer +889470
receiver +774581
sender   +774581
consumer +841173

So receiver.Replace(track) / MoveNode() behaved exactly as expected: the existing consumer and sender survived, the sender parent moved to the new H264 receiver, and video resumed.

go2rtc itself stayed on the same PID throughout, the passive curl consumer stayed on the same PID, and all temporary firewall rules were removed afterward.

That settles the consumer-survival question for me.

The separate finding is the natural H264-only stall: audio continued normally for >10 minutes, H264 remained frozen, and the producer never reconnected. That appears to be a different failure mode from the ExtendStream failure fixed in my current test build. My next step is to look at whether Nest-specific video-liveness detection is needed so that a prolonged H264 stall can close the transport and reuse the same reconnect path we just validated.

ajplotkin:
consumer 60 -> 60, sender 61 -> 61, parent 57 -> 567 with bytes advancing is the
clean result. `MoveNode` carries the consumer across the reconnect, so the
escalation recovers an attached consumer transparently, and the earlier Frigate
restart was its own watchdog rather than a severed graph.

On the H264-only stall: receiver 57's `bytes` moves only from `track.WriteRTP` at
the bottom of the per-track read loop in `pkg/webrtc/conn.go`, and that loop
returns with no log on a `remote.Read` error or an `rtp.Packet.Unmarshal` error.
A frozen H264 receiver beside a live Opus receiver fits either Google no longer
sending video or the video goroutine having exited, and `/api/streams` cannot
tell those apart. Neither changes PeerConnection state, so `reconnect()` has no
trigger. Closing the transport covers both.

I run a Nest-gated version of that check in `pkg/webrtc/conn.go`. It is in
https://github.com/ajplotkin/nest-homekit-cameras/blob/main/patches/go2rtc-nest.patch
if the code is useful. Per video track it stores the time of the last non-padding
video packet and the last IDR, ticks every second, and closes the connection on
either branch:

```go
// video flowing, no keyframe for 60s: permanently dry session
if videoAge <= 8*time.Second && idrAge > 60*time.Second { c.Close() }
// no real video at all for 8s: dead session or sustained outage
if videoAge > 8*time.Second { c.Close() }
```

Three things that cost me time and are not obvious from the outside:

The keyframe deadline started at 4s and that was wrong. Over 48h it fired 23
times, and 13 of those killed a session whose video had arrived under 2s earlier
(median 1.28s, several at 40 to 120ms). Healthy Nest jitters keyframes into the 3
to 4s range routinely, so 4s sits inside normal behaviour, and every kill forces
a re-dial. It manufactured the condition it was meant to catch.

The keyframe branch has to be gated on `videoAge`. Since a keyframe is also a
video packet, `idrAge >= videoAge` always, so without the guard the fast trigger
shadows the full-stall branch and turns a 5s network blip into a teardown.

Padding and bandwidth-probe packets must not refresh the video timestamp. A
camera that keeps probing through a drought otherwise looks alive indefinitely.

Nest also withholds video for roughly 12 to 24s after a motion event while it
uploads its own clip, and ignores FIR during that window (idr_age climbed past
11s with FIR sent five times). A re-dialed session opens with an IDR and recovers
in 2 to 4s, so a fresh session is the only thing that ends it. That is why the
full-stall threshold is 8s rather than something longer.

Two counters to avoid when you build it: the webrtc producer's `bytes_recv` sums
every track, so it stays healthy while one track is dead, and an RTSP consumer's
`bytes_send` increments only when `writeInterleavedData` fails. `receivers[].bytes`
and `senders[].bytes` are per track. Nothing in `pkg/core` records when a counter
last moved, so the check needs its own timestamps.

One logging note: Nest assigns SSRC 7777 to every camera, so SSRC cannot
distinguish them. I carry a connection id in the log lines instead.

donparlor:
Thanks — this was very helpful. I audited your Nest watchdog patch read-only against my current 2ca8a77 tree, and the overall architecture lines up very well with what we observed.

One edge case stood out that I wanted to run by you.

In the current patch, the per-track watchdog is tied to:

stallDone := make(chan struct{})
defer close(stallDone)

while the remote.Read() and rtp.Packet.Unmarshal() error paths still just return.

That seems to mean:

if Google simply stops delivering video and remote.Read() remains blocked, the watchdog does exactly what we want;
but if the H264 goroutine exits because remote.Read() returns an error, or because RTP unmarshal fails, the deferred close(stallDone) also terminates the watchdog immediately.

In that second case, if Opus keeps the PeerConnection alive, we could end up back in the same state: no H264, PC still connected, and nothing closing the transport.

My current plan is therefore to keep the watchdog, but for Nest/H264 explicitly call c.Close() on those two error paths before returning. I’d also make the watchdog lifetime tied to the connection close/stop rather than only to the OnTrack goroutine.

I’m also leaning conservative on the first no-video threshold:

no video RTP : 30s
no IDR       : 60s, only while video is otherwise recent
no video ever: 30s

mainly because of your 12–24s Nest event-upload drought observation. The 60s IDR threshold makes complete sense to me given your 48h data.

I’m leaving 2ca8a77 untouched until its current 72h soak completes, then I’ll make the H264 recovery a separate commit on top of it.

Have you ever caught one of your H264 stalls ending through an actual remote.Read() error or RTP unmarshal error, rather than the read simply staying blocked? That would help distinguish how important that explicit-close path is in practice.

ajplotkin:
Agreed on the defer, and the gap is wider than that.

What I had not thought through is that nothing else picks it up. The connect watchdog I added in
`OnConnectionStateChange` is a one-shot `time.Sleep(30 * time.Second)` at connect that reads
`nestVideoSeen`, a one-way latch set by the first video track. After one video packet it can never
fire again for that connection. So in your scenario the stall watchdog is gone via the defer and
the connect watchdog is latched off.

The case that is covered is Google tearing the session down: ICE goes disconnected or failed and
upstream closes at `conn.go:140-143`. What is not covered is the transport staying up while the
video goroutine is gone, and that does not need Opus to hold it open. pion marks the selected pair
disconnected only when nothing has arrived from the remote candidate for `disconnectedTimeout`, 5s
by default and not overridden by go2rtc (`agent.go:824` in ice v4.2.0), and any inbound packet from
that address refreshes it, STUN or not (`agent.go:1422`, `candidate_base.go:283`). Google's consent
checks alone keep the PeerConnection `connected` with no media at all.

Both returns are reachable by a single packet in the versions go2rtc pins, and neither logs:

- `TrackRemote.Read` calls `checkAndUpdateTrack` on every packet, and it returns an error for a
  payload type that is not in the negotiated list (webrtc v4.2.3, `track_remote.go:163-176`).
- `Packet.Unmarshal` returns `errTooSmall` when the padding-count byte exceeds the packet
  (rtp v1.10.0, `packet.go:237-248`). That byte is inside the SRTP-encrypted portion, and neither
  srtp nor the default interceptors look at it, so this is the first check it meets.

I have not seen Nest do either and I would expect both to be rare. The reason the close-on-error
still belongs in is what happens when one lands: the goroutine is gone, the watchdog with it, the
transport stays up, and nothing logs.

On whether I have seen it: no, and this data cannot rule it out. The window is 7 days, 62 watchdog
arms across three container runs, 39 fires, all on the no-video-RTP branch. A fire proves the
`OnTrack` goroutine had not returned, so none of those 39 went out through either error path. That
is all it proves. From the watchdog lines alone your case and a clean shutdown look identical:
ticks stop, no fire line. The producer's lifecycle lines would separate them, since a `Close()`
makes `Start()` return and `reconnect()` logs a retry, while a zombie logs neither. A log line on
those two returns would settle it directly and is worth adding either way.

On the thresholds. In my last comment I argued for 8s on the grounds that a re-dial ends a drought
faster than waiting it out. That still holds for a real drought. What I had not checked is how many
of the 8s kills were droughts at all.

The debug tick logs whenever `video_age` passes 3s. Counting episodes where it did, 65 recovered on
their own against the 39 the watchdog killed. The recovered ones by how deep they got:

| peak `video_age` | episodes |
|---|---|
| 3-4s | 56 |
| 4-6s | 6 |
| 6-7s | 1 |
| 7-8s | 2 |

The last two were sampled at 7.64s and 7.95s. The watchdog samples at 1Hz, so a 7.95s reading with
no fire on the next tick means video arrived somewhere between 7.95s and 8.95s. The 39 kills were
read at 8.01s to 8.98s. A gap ending anywhere in that overlap is killed or survives depending only
on where the tick lands. What I cannot see is how long the killed ones would have run, because the
kill censors them, so this says 8s is inside the recoverable tail and says nothing about where
between 9s and 30s the line belongs. 30s is above the 12-24s post-motion droughts I measured before
the watchdog existed, which is the reason you gave for it. The 7.95s session was still running 23
hours later.

That is the same mistake as the 4s keyframe deadline I already wrote a comment about in that
function. I fixed that branch and left this one at 8s without checking where its distribution ends.

60s on the keyframe branch has never fired here. The longest keyframe gap I can watch grow from the
start with video still arriving is 9.6s. I had a 17.9s reading and threw it out: the watchdog armed
19s earlier and its first tick already reported 16.9s, with no ticks in between, which is a clock
step rather than a gap, since the ages come from `time.Unix` and carry no monotonic reading. So I
have nothing between 10s and 60s, and cannot argue 60s over 30s from data, only that nothing
observed comes near either.

The one live sample of this failure either of us has is your producer 56, ten minutes of Opus
advancing with H264 frozen and no reconnect. Nothing in `/api/streams` says whether the video
goroutine was still blocked in `remote.Read` or had already returned, which is where a log line on
those two returns pays off. The next one tells you which it was, and that decides whether raising
the thresholds alone would have caught it or whether these are two independent bugs with the same
symptom. Either way the close-on-error goes in.

donparlor:
Thanks — this is very useful, especially the 7-day distribution.

I agree that the current build cannot distinguish the two cases yet:

the H264 remote.Read() remains blocked with no video arriving, or
the track goroutine has already exited through remote.Read() / RTP unmarshal while the PeerConnection stays connected.

That also means there is not much value in forcing another failure on my current 2ca8a77 build, because those return paths are still silent. I am keeping that build unchanged through its current 72h soak.

Your recovered-stall distribution also convinces me not to use 8s for the first no-video threshold. The 7.95s case makes it clear that 8s is inside the recoverable tail, and the 12–24s post-motion droughts argue for a more conservative initial value.

My next separate commit will therefore focus on:

safe logging on the Nest/H264 remote.Read() and RTP decode exits;
explicit transport close on a fatal track exit;
a cancellable no-video watchdog for the blocked-read case;
monotonic elapsed-time tracking rather than UnixNano() reconstruction;
a conservative initial no-video threshold, likely around 30s;
keeping the IDR path separately conservative.

Once that instrumentation is running, the next natural stall should tell us directly whether the video goroutine was still blocked or had already exited.

For now I’ll let 2ca8a77 finish its soak unchanged and report back with the next implementation separately.

ajplotkin:
One detail from making the monotonic change on my side, in case it saves you a step.

The natural form is a `time.Time` per track, and that does not fit in an `atomic.Int64`, which
is what the read loop writes and the watchdog goroutine reads. `atomic.Pointer[time.Time]` fits
but allocates on every packet. I settled on a base captured once per track, with the atomics
holding elapsed time against it:

```go
nestEpoch := time.Now() // carries a monotonic reading; an int64 cannot

// read loop
lastVideoNS.Store(int64(time.Since(nestEpoch)))

// watchdog tick
elapsed := time.Since(nestEpoch)
videoAge := elapsed - time.Duration(lastVideoNS.Load())
```

That is in `nestfix-1.9.14-13` on my fork.

One correction to what I implied earlier about the consequences. The 17.9s reading I threw out
came on the first Nest dial after a reboot, on a Pi with no RTC, and it showed up only on the
keyframe age: `video_age` on the same tick was 43ms. Every packet refreshes that one, and
packets arrive far more often than the 1s tick, so a step landing during flowing video is
erased before the watchdog looks at it. The hazard is narrower than a step simply exceeding a
deadline: the step adds itself to whichever gap is already open when it lands, so a modest real
drought plus a step can cross a deadline that neither would reach alone. Worth having at any
threshold, and it does not argue for one over another.

donparlor:
Thanks — that saves a step. The per-track monotonic epoch + atomic elapsed duration is exactly the direction I was leaning toward, and I agree the clock-step issue is a robustness problem rather than evidence for one timeout value over another.

I’ll use that model in the separate H264 recovery commit after the current 2ca8a77 soak completes.

donparlor:
@ajplotkin — following up on my last note, the 72-hour 2ca8a77 soak is complete, and I’ve now built and deployed the separate H264 recovery patch. Your observations and the monotonic-timing suggestion were particularly useful in shaping it.

What the soak established

The controlled SDM failure had already demonstrated that the extension fatal path could close the transport, create a fresh Nest session, and carry the passive MP4 consumer across reconnect (consumer 60 → 60, sender 61 → 61, parent 57 → 567, with video bytes advancing).

The longer soak exposed a distinct limitation: H264 eventually stopped advancing while Opus continued. I confirmed the same producer and H264 receiver remained in place for over four additional hours, with frozen H264 and MP4 sender counters, live audio, and no go2rtc process restart or producer reconnect. The graph was still correctly attached. So 2ca8a77 validated the extension-recovery mechanism, but did not solve H264-only liveness.

The separate H264 patch

I kept this deliberately Nest-specific and smaller than the full watchdog in your fork:

A continuous 30-second no-valid-H264 watchdog, covering both no initial video and a stall after video has started.
A monotonic epoch plus atomic elapsed time, following your suggestion.
Watchdog lifetime tied to the connection, not to the OnTrack read-loop goroutine. This closes the defer close(stallDone) gap we discussed.
An explicit transport close when Nest/H264 remote.Read() returns an error.
For rtp.Unmarshal() errors, I chose to count and discard the individual malformed packet without terminating the read loop. Persistent absence of valid video still reaches the 30-second watchdog. This is a deliberate difference from my earlier plan to close immediately on both errors.
Safe recovery diagnostics distinguishing read_error, no_video_initial, and stalled, without logging the Nest source URL or credentials.

I left the IDR watchdog out of this first commit. Our demonstrated failure is a complete H264 RTP stall, and your 60-second IDR branch had not fired in the observations you shared. I’d rather validate one recovery mechanism before introducing a separate keyframe-liveness policy.

Testing and deployment

The local tests cover the timeout boundaries, initial-video absence, live-video non-regression, concurrent recovery/close, and Conn.Start() unblocking. During race testing, we also uncovered and fixed a pre-existing core.Waiter result-publication race in a separate commit.

A real two-peer Pion integration test confirmed that H264 RTP reaches OnTrack, the simulated stale-video condition triggers the watchdog, Conn.Start() returns, and the PeerConnection reaches closed. That integration test passed three times with -race; the full core, webrtc, and nest package race tests also pass.

Local commits, on top of 2ca8a77:

3c55fe6 fix(core): synchronize Waiter result publication
ea1bfde fix(nest): recover stalled WebRTC H264 video

ea1bfde was deployed to my Frigate/Nest Doorbell installation on September 19 at 05:55 UTC. The running go2rtc process reports revision=ea1bfde, version 1.9.10, Linux ARM64.

Initial production validation is healthy: one Nest producer, active video and audio receivers, and the passive MP4 consumer attached to the same fed receiver as Frigate. Over 10 seconds, the video receiver and MP4 sender each advanced by 1,530,336 bytes. The new passive MP4 probe is being left attached to observe the next natural reconnect.

What remains unproven: the new build has not yet encountered and recovered from a natural H264-only stall in production. The local Pion test establishes that the watchdog can close a stalled connection; the next real event will tell us whether the full Nest redial and consumer migration behave as expected under that failure.

Thanks again for sharing the seven-day stall distribution and the monotonic-clock approach. Both directly influenced this implementation. I’ll report the first natural H264 recovery with the producer/receiver/consumer IDs, sender parent, video-byte deltas, and recovery reason rather than treating a healthy startup as the final result.

ajplotkin:
Both of your fixes are now in my fork as `nestfix-1.9.14-16`, and I have the stall counters
for you after all.

The `core.Waiter` race reproduces here. I wrote a test against the unpatched file and `-race`
flags the read at `return w.err` in `Wait()` against the write in `Done()`. The sequencing is
that `WaitGroup.Done()` releases a blocked waiter immediately and `w.err` is only assigned
afterwards, while `Wait()` reads it without holding `w.mu`, so a waiter can be released and
read the previous value. In our case that means a connection that ended with an error can be
reported as having ended cleanly. I moved the assignment ahead of the `WaitGroup.Done()`,
which also gives the unlocked read a happens-before edge through the WaitGroup rather than
needing a lock on the `Wait()` side. Good find, and it is a pre-existing bug rather than
anything either of us introduced.

Your `rtp.Unmarshal` decision is better than mine and I have taken it. I was closing the
transport on any unmarshal error. Your reasoning holds: persistent absence of valid video
reaches the watchdog anyway, and the path is reachable on a single malformed packet, since
`errTooSmall` comes from the padding-count byte and that byte lives inside the SRTP-encrypted
portion where nothing upstream validates it.
Tearing down a healthy session over one packet is the same false-positive teardown my 4s
keyframe deadline used to cause. It now counts and discards and keeps reading, and the stall
watchdog still catches a genuine stop, so patience costs nothing. The `remote.Read` path still
closes.

Leaving the IDR watchdog out looks right on my data too, not just as sequencing. Across the
seven days I shared, all 39 fires were the no-video-RTP branch and the keyframe branch fired
zero times. The longest keyframe gap I can actually watch grow from the start, with video
still arriving, is 9.6s, so 60s has a wide margin and nothing has approached it. The two days
since add nothing to that: no keyframe fire, and the longest gap with video flowing was 7.8s.

One thing for anyone reproducing your core race test on master rather than on your tree:
`pkg/core` does not compile its tests there. `core_test.go` still calls `StripUserinfo`, which
`de157eb` (2026-01-21, two days after the v1.9.14 tag) removed from `pkg/core/helpers.go`
without touching the test, so `go vet ./pkg/core/` on `c245815` fails with
`undefined: StripUserinfo`. Stock v1.9.14 has the function at `helpers.go:103` and v1.9.10 at
`helpers.go:88`. The version string your build reports comes from the constant in `main.go`,
which puts your tree before the removal, and that fits your core race tests compiling. I had
to move that file aside to run mine.

On your original question, whether a stall ends through a `remote.Read` error or the read
simply staying blocked. Two days on the instrumented build, no restarts:

| | count |
|---|---|
| `closing stalled stream (no video RTP)` | 9 |
| video track read failed (other than `io.EOF`, the teardown path) | 0 |
| RTP unmarshal failed | 0 |

So nine real stalls, and every one of them ended with the read still blocked. None went out
through either error return. The 39 from the seven-day window do not add to that count: each
of those fires proved the goroutine was alive, but that build logged nothing on either error
return and its watchdog died with the goroutine, so a session that exited that way would have
left no fire and no line. These two days are the only window in which an error exit could have
shown up at all, and none did.

That makes the defer bug latent here rather than the thing actually biting, which is the
answer I would have bet against. It does not make the fix optional, since the failure it
guards against leaves no trace and I would not know if it had happened. But on this hardware
the thresholds are the live issue and the error paths are insurance. The no-video branch is
doing all the work here, and on my build that is still 8s rather than your 30s, so these
numbers say nothing about how 30s would have behaved. I am leaving the collection running to
the 23rd in case a longer window turns an error exit up.

donparlor:
@ajplotkin — excellent follow-up, and thank you for independently reproducing the core.Waiter race and incorporating both fixes into nestfix-1.9.14-16.

Your new counters answer the question I was particularly interested in. Nine no-video watchdog fires with zero unexpected remote.Read errors and zero RTP unmarshal errors point to the blocked-read/no-video path as the observed failure mode in that instrumented window. The explicit read-error recovery remains useful protection against the silent goroutine-exit case, but the continuing no-video watchdog is clearly the mechanism doing the work in your current observations.

I also agree with your distinction between our thresholds: your nine interventions at 8 seconds do not establish how many would have persisted to my 30-second deadline. That remains something my production build needs to establish independently.

Your alternative core.Waiter fix is interesting. Mine synchronizes the final read with the existing mutex; yours publishes the error before WaitGroup.Done(), establishing the required ordering through the WaitGroup itself. Both address the race we reproduced, so I’m keeping my already-tested build unchanged during the current production observation period.

I'm glad the RTP unmarshal distinction proved useful in your fork. It seems preferable to preserve an otherwise healthy Nest session after one malformed packet and reserve transport teardown for an actual loss of valid video.

My ea1bfde build remains deployed with the passive MP4 consumer attached. I'll report the first natural H264 recovery with the recovery reason, producer/receiver IDs, sender-parent migration and byte deltas. Looking forward to your additional observations through September 23.

ajplotkin:
@donparlor the longer window, same build (`b203b4e`, `nestfix-1.9.14-15`) and the same container run with no restarts: 2026-09-16 23:30 UTC to 2026-09-23 14:05 UTC, about 6.6 days on four cameras, 67 watchdog arms. The Pi is still on `-15`, so the unmarshal row below is the close-on-error version, not your count-and-discard change that is in `-16`.

| | count |
|---|---|
| `closing stalled stream (no video RTP)` | 40 |
| `closing stalled stream (no keyframe)` | 3 |
| video track read failed (other than `io.EOF`) | 0 |
| RTP unmarshal failed | 0 |

The result from the 19th holds: 43 watchdog fires, and in every one the read loop had not returned when the watchdog fired. No session in the window went out through either logged error return. The limit is the other 24 arms: four are the sessions still running, and 20 ended without a fire. An `io.EOF` read exit is silent by design, so that is where an unlogged exit could sit. The window also has 202 clean session ends against the 67 arms. The sessions that never armed never had a video track reach `OnTrack`, so they had no video read loop to exit. As before, all of this is at 8s and says nothing about what those reads would have done by 30s.

I need to correct something I told you on the 19th. I said the keyframe branch had never fired and that the longest keyframe gap I could watch with video arriving was 9.6s. Both changed on the 21st, during one event.

At 09:37:01 UTC video stopped on all four streams within 40 ms of each other. For the next 5m40s every dial hit the 10s client timeout on `GenerateWebRtcStream`. I can't tell from these logs whether that was my network or Google. When sessions came back at 09:42:58, all four sessions got video for 14 to 21s, with keyframes only in the first 10 to 12s, then RTP stopped on all four a second time and the no-video branch closed them at 09:43:22 to 09:43:28.

Then the keyframe branch fired three times in a row on one camera, the front door, which is a Nest doorbell. Each re-dial (conns 47, 49, 50) was closed after 60s without a keyframe, at 09:44:29, 09:45:32 and 09:46:34. `idr_age` is measured from the arm on the first packet. On two of the three it tracked the arm time to within a millisecond, so no keyframe arrived after the opening packet, and the logs can't say whether that packet was itself an IDR. The third got one keyframe 0.25s in and none in the 60s after. The next re-dial got a keyframe after 21s.

For two of the three, RTP was still arriving at the fire (`video_age` 0.8s and 1.6s), with gaps of up to 7s along the way, so an 8s no-video watchdog would not have closed them then and a 30s one would not have come close. The third had `video_age` 7.9s at the fire, one tick from the no-video branch.

Other sessions in the same recovery ran with video flowing and no keyframe for over 30s ten times, across four sessions on three cameras, the longest 58, 57, 54 and 45s, before one arrived. Every one of those gaps ended with a keyframe, the last at about 09:47:18. Outside that event the longest such gap in the window is 13.8s. Since your build has no IDR branch, the 35 to 58s gaps that recovered on their own are probably the more useful numbers. This is one event at one site.

The four sessions that were live when it started are not evidence either way for your question. They ended 5 to 7s after the last video packet, before the 8s watchdog could fire, with `producer ended without error` and no read-error line. Peer-connection state changes aren't logged for the Nest producer (upstream doesn't log them either), so from these logs an ICE-side close and a read that returned `io.EOF` look the same. pion's default 5s disconnected timeout fits the timing, but the logs don't show the state.

donparlor:
@ajplotkin — thank you for following through with the longer collection window. The September 21 event changes the IDR discussion in an important way, and I appreciate you correcting the earlier conclusion rather than treating the previous zero-fire result as settled.

The three 60-second no-keyframe closures on your Nest doorbell establish that the IDR branch can encounter a real condition that the no-video-RTP watchdog does not cover: RTP was still arriving during at least two of those interventions. At the same time, the 45–58-second gaps that recovered spontaneously make me reluctant to introduce a 60-second IDR teardown into my currently stable build without first observing this behavior locally.

I also have some production results to report since my September 19 update.

The 30-second H264 watchdog has now recovered natural stalls. In the retained logs I examined on September 23, I found eight reason=stalled recoveries, with video ages around 30.2–30.7 seconds, and two reason=read_error recovery entries. Those read-error entries need additional context before I attribute them to the specific silent-exit failure we originally discussed. I have not observed an RTP-unmarshal-related recovery.

More importantly, the passive MP4 consumer gave us the continuity evidence we were looking for. It remained attached through natural producer reconnects without needing to recreate the consumer or its sender. Near the end of that observation period, the existing MP4 consumer was still 7, its H264 sender was still 8, and the sender had migrated to the current receiver 1061. Over a 15-second comparison, that receiver, the MP4 sender, and Frigate’s H264 sender each advanced by exactly 2,238,690 bytes. That is useful independent confirmation of consumer migration under natural recovery, not just the earlier controlled SDM test.

There is still an unresolved failure mode on my installation: I found two periods of Frigate FFmpeg/PPS or no-frame errors on September 23 without corresponding [nest/h264] recovery events. That does not establish that RTP was arriving without decodable IDRs—the old build did not expose enough information to make that distinction.

So I took a deliberately diagnostic next step rather than adding another reconnect policy.

I have just deployed a separate instrumentation commit, ed4cd81, on top of ea1bfde. It observes IDR/SPS/PPS NAL units in Nest H264 RTP, including single NAL, STAP-A, and FU-A start packets, and reports prolonged periods of RTP activity without a recently observed IDR at 30/60/90/120 seconds. These are diagnostic thresholds only: they do not close the transport. The existing continuous 30-second no-valid-H264 watchdog remains unchanged.

The new build passed the relevant package tests and race tests and started in Frigate at 2026-09-24 00:45 UTC. I reattached the passive MP4 probe. Its first 15-second baseline is healthy:

One Nest producer, one Frigate MPEG-TS consumer, and one passive MP4 consumer.
H264 receiver: +2,224,005 bytes / +1,970 packets.
Frigate H264 sender: +2,224,005 bytes / +1,970 packets.
MP4 H264 sender: +2,224,005 bytes / +1,970 packets.
Opus receiver: +30,504 bytes.

There were no [nest/h264] diagnostic events in the first few minutes, which is simply our starting baseline—not evidence yet about the unresolved FFmpeg episodes.

Your September 21 traces are exactly why I wanted this separation. If my doorbell encounters RTP-without-IDR periods, I should now be able to measure their duration and compare them with FFmpeg failures before deciding whether a second, IDR-specific recovery trigger is justified. I also want to preserve the distinction between an IDR NAL observed on the wire and a successfully decoded video frame.

Thanks again for the detailed multi-camera timeline and for keeping the -15 versus -16 distinction explicit. I’ll report back once the new instrumentation catches something meaningful.

Issue #2365, opened 2026-07-17
webrtc: consumers attaching mid-stream to dialed sources (nest) never get a keyframe; Google ignores PLI but honors FIR


benholtz:
## Problem

For dialed (active producer) WebRTC sources such as `nest:`, any consumer that attaches to an already-running stream never receives a keyframe and decodes reference-broken P-frames indefinitely (grey/smeared output, or ffmpeg consumers failing with `non-existing PPS 0 referenced`).

Google Nest cameras send SPS/PPS+IDR only once at dial time and never repeat them in-band, so the first consumer (attached during the dial) decodes fine, while every later consumer joins mid-GOP and has nothing to bootstrap from. With two consumers on one nest stream (in my case Frigate record + detect on the same URL) whichever connects second is permanently broken until the whole stream is restarted.

## Root cause

`pkg/webrtc/conn.go` starts a periodic PLI loop in `OnTrack`, but only for `ModePassiveProducer`. Active producers (all dialed WebRTC sources) never request keyframes upstream, so the source has no reason to emit another IDR.

Two additional findings from testing against Google's servers:

1. Extending the existing PLI loop to active producers is NOT enough. Google ignores bare PLI: with a 5s PLI loop running I measured 285 frames over 12s with zero keyframes on a freshly attached consumer.
2. Sending FIR (RFC 5104, with an incrementing sequence number) works. With PLI+FIR every 5s a freshly attached consumer receives an IDR within ~5s and decodes cleanly. Both feedback types are already negotiated in the offer (`videoRTCPFeedback` in `pkg/webrtc/api.go`), so no SDP changes are needed.

## Fix I'm running in production

Replaces the `ModePassiveProducer`-only PLI block in `OnTrack` (`pkg/webrtc/conn.go`):

```go
if remote.Kind() == webrtc.RTPCodecTypeVideo &&
	(c.Mode == core.ModePassiveProducer || c.Mode == core.ModeActiveProducer) {
	// Periodic keyframe requests keep a recent IDR in the stream so
	// consumers that attach mid-stream can start decoding. Passive
	// producers (browser push) always got a 2s PLI. Active producers
	// (dialed sources, e.g. nest:) sent no keyframe requests at all.
	// Send BOTH PLI and FIR: both are negotiated in the offer, and
	// senders differ in which they honor (Google Nest ignores PLI,
	// honors FIR). FIR carries an incrementing sequence number per
	// RFC 5104 so repeated requests aren't deduplicated.
[...]

PR #2375, opened 2026-07-22
to-dev: fix(webrtc): reuse existing Receiver when remote switches payload type mid-session

misecichin:
## Bug

Some WebRTC producers start sending media on a different RTP payload type than the one negotiated at connection setup, without a full SDP renegotiation. I hit this consistently with a Nest doorbell (`nest:` source, `protocols=WEB_RTC`).

`Conn.GetTrack` (`pkg/webrtc/producer.go`) deduplicates `Receiver`s by exact `*Codec` pointer identity:

```go
for _, track := range c.Receivers {
    if track.Codec == codec {
        return track, nil
    }
}
```

When the remote peer switches payload type for the same `Media`, `getMediaCodec` resolves a *different* `*Codec` pointer (matching the new payload type), so this lookup misses and a brand-new, unwired `Receiver` is created — permanently starving every already-connected consumer (RTSP/WebRTC senders) of video, even though the producer is actively receiving valid frames on the new payload type.

## Reproduction

go2rtc negotiates H264 on payload type 96 with the Nest media server. A few seconds into the session, Nest begins transmitting on payload type 98 instead (same media, no renegotiation). Before this fix, `/api/streams` shows:

```json
"receivers": [
  {"id": 3, "codec": {"codec_name": "h264"}, "childs": [4]},                    // wired to the RTSP sender, 0 bytes forever
  {"id": 7, "codec": {"codec_name": "h264"}, "bytes": 787424, "packets": 730}   // orphan, no consumer, real data
]
```

ffmpeg consuming the RTSP restream fails with:

```
Could not find codec parameters for stream 0 (Video: h264, none): unspecified size
```

I reproduced this identically on v1.9.0 and v1.9.10, and also with an unmodified `alexxit/go2rtc:latest` — so it isn't a version-specific regression, and adding explicit `video=h264&audio=opus` hints to the source URL doesn't change the behavior either.

## Fix

Reuse the existing `Receiver` for the same `Media` when no exact codec-pointer match is found, instead of creating an orphan:

```go
case core.ModePassiveProducer, core.ModeActiveProducer:
    for _, track := range c.Receivers {
        if track.Media == media {
            return track, nil
        }
    }
```

This is safe because `Receiver.Input` just forwards raw RTP payload bytes to its children — it doesn't depend on which declared payload type the packet arrived on, and outgoing payload-type stamping is handled separately by the `Sender` side. I don't see simulcast/multi-encoding selection logic in this producer path that this would conflict with, but happy to adjust if I'm missing a case this affects.

## Verification

After the fix, `/api/streams` shows a single `Receiver` correctly wired to the consumer, and ffmpeg reports a fully valid stream:

```
Stream #0:0: Video: h264 (High), yuv420p(progressive), 384x512, 30 fps, 15 tbr, 90k tbn
Stream #0:1: Audio: opus, 48000 Hz, stereo, fltp
```

15MB+ of real video flowing through in a 12s test, versus 0 bytes on the wired receiver before the fix. Also validated against a real Frigate deployment (ffmpeg 7.0) pulling the RTSP restream in production, stable with no drops over multiple minutes, versus crash-looping every ~20s before.

### DIFF
--- a/pkg/webrtc/producer.go
+++ b/pkg/webrtc/producer.go
@@ -34,6 +34,22 @@ func (c *Conn) GetTrack(media *core.Media, codec *core.Codec) (*core.Receiver, e
 		// Passive producers: OBS Studio via WHIP or Browser
 		// Active producers: go2rtc as WebRTC client or WebTorrent
 
+		// Some remote peers (observed with Nest/SDM WebRTC) start sending
+		// media on a different payload type than the one originally wired
+		// to consumers (e.g. switch from PT 96 to PT 98 mid-session for the
+		// same video media, without a renegotiation). Since each payload
+		// type is a distinct *Codec pointer, the exact-codec lookup above
+		// misses and a brand-new, unwired Receiver would otherwise be
+		// created for it, silently starving every already-connected
+		// consumer. Reuse the existing Receiver for this Media instead, so
+		// already-wired Senders keep receiving packets regardless of which
+		// declared payload type the remote actually transmits on.
+		for _, track := range c.Receivers {
+			if track.Media == media {
+				return track, nil
+			}
+		}
+
 	default:
 		panic(core.Caller())
 	}


misecichin:
Current docker with the fix which i'm using now for the Nest Doorbell 2:
misecichin/go2rtc-nest:1.9.10-nestfix

ajplotkin:
Independent confirmation, from a different deployment and a different consumer path: this is real, and your diagnosis matches mine exactly.

I hit the same thing on a Nest Doorbell (wired) via `nest:` on go2rtc 1.9.14, with an RTSP consumer rather than Frigate. Same mechanism — `Conn.GetTrack` dedupes by `*Codec` pointer, Nest starts transmitting on a payload type that resolves to a different `*Codec`, the lookup misses, and a second unwired `Receiver` is created while every already-attached consumer starves. `/api/streams` shows two video receivers, one with `childs` and no bytes, one with bytes and no children.

What made it maddening to diagnose is that it is **dial-order dependent**: a cold dial mis-binds essentially every time, while a consumer attaching to an already-warm producer binds correctly. So it presents as "recording works if I open the camera first," which sends you looking in entirely the wrong place.

I filed it as #2389 before finding this PR — happy for that to be closed as a duplicate if you'd rather consolidate. **#2386** is a third independent report (Nest Doorbell Gen 2 via Frigate, also stuck in a restart loop), and **#2365** is arguably a fourth party hitting a downstream symptom of the same starvation. Cross-linking them here so it is clear this is not a one-off.

## One suggestion on the patch itself

The lookup matches on `Media` alone:

```go
for _, track := range c.Receivers {
    if track.Media == media {
        return track, nil
    }
}
```

`core.Media.Codecs` is a list, and `getMediaCodec` (`pkg/webrtc/conn.go`) selects among those codecs *by the transmitted payload type*. So "same Media" does not imply "compatible codec" — a video m-line that offered H264 and VP8 is ordinary WebRTC, and a remote moving from the H264 PT to the VP8 PT would, with this patch, be handed the receiver still described as H264. Consumers would then get VP8 RTP labelled H264, which fails in a much more confusing way than the bug being fixed.

Adding the compatibility check costs nothing and bounds it to the case you actually observed:

```go
for _, track := range c.Receivers {
    if track.Media == media && track.Codec.Match(codec) {
        return track, nil
    }
}
```

`Codec.Match` compares name, clock rate and channels (and ignores `FmtpLine`), so a genuine PT-switch-within-the-same-codec still matches — which is the Nest behaviour — while a switch to a different codec falls through to the existing create path rather than being silently mis-wired.

## The variant I have been running in production

For what it is worth, mine updates the receiver's codec rather than returning it as-is, and is gated to the Nest source to bound the blast radius:

```go
if c.FormatName == "nest/webrtc" {
    for _, recv := range c.Receivers {
        if recv.Media == media && recv.Codec != codec && recv.Codec.Match(codec) {
            recv.Codec = codec
            break
        }
    }
}
track, err := c.GetTrack(media, codec)
```

Updating `recv.Codec` keeps the receiver's description in step with what is actually arriving, which matters for anything reading the codec downstream (the RTSP server copies `FmtpLine` into `DESCRIBE`). Your version is simpler and covers more sources, which I think is the better shape for upstream — I mention mine only because the two differ in that one respect and it may be worth a deliberate choice rather than an accident.

Running since 2026-07-31 across four Nest cameras: cold restarts now produce one video receiver with bytes flowing, zero starved receivers, and the "recording delivered audio but no video" failures are gone.

phob-eng:
Tested this PR against a real Nest Doorbell (SDM `nest:` source, `protocols=WEB_RTC`) and can confirm it fixes the empty-receiver starvation described in #2386.

**Setup:** built the PR head (2eb363f) for linux/amd64, deployed as a drop-in replacement for the HAOS go2rtc add-on (dev, 9cec93c), consumed by Home Assistant 2026.8.1.

**Before (both 1.9.14 stable and dev 9cec93c):**
- Any ffmpeg/PyAV RTSP consumer received the SDP but **zero media packets, ever** — `Could not find codec parameters for stream 0 (Video: h264, none): unspecified size` regardless of `-analyzeduration`/`-probesize` (tested up to 60s), while parallel WebRTC/MSE consumers of the same producer played fine. Frigate's record/detect ffmpeg crash-looped identically (frigate#9250 symptom).
- go2rtc's own HLS died with `[hls] can't get init`, and `/api/stream.mp4` returned an init-only 657-byte file.

**After (this PR):**
- Home Assistant's stream worker (PyAV over RTSP) produces a filled HLS playlist within ~8 seconds of a cold start.
- WebRTC/MSE consumers unaffected (still working).

One note: the WS `{type:"mp4"}` request still replies with an `mse`-typed init message and never sends media, before and after this PR — that appears to be a separate issue.

Thanks for the fix — happy to test further builds on this hardware.

PR #2377, opened 2026-07-23
Fix xiaomi/cs2 stream restart on pop buffer overflow

DongbinNie:
## Problem

When the media channel pop buffer (hardcoded capacity 100 packets) fills up — e.g. the consumer pipeline is temporarily slower than the camera (CPU throttling, slow downstream client) — `dataChannel.Push` returns a fatal error. This kills the whole camera connection and forces a stream restart with re-probing. Users see repeated warnings:

    miss: read media: cs2: pop buffer is full

and periodic video interruptions (in my setup every ~10 seconds).

## Fix

- Drop the oldest queued packet instead of failing the connection. Consumers already tolerate packet loss (RTP sequence gaps) and decoders resync on the next keyframe — far better than restarting the whole stream.
- Make the pop buffer size configurable via the `buffer` URL param (default 256, previously hardcoded 100), so setups with slow consumers can trade memory for fewer drops, e.g. `xiaomi://...&buffer=1024`.

Tested with a chuangmi.camera.81ac1 (cs2+tcp): the stream no longer restarts under consumer-side slowdowns.

### DIFF
--- a/internal/xiaomi/README.md
+++ b/internal/xiaomi/README.md
@@ -62,3 +62,10 @@ You can use a second channel for dual cameras: `channel=2`.
 streams:
   xiaomi1: xiaomi://***&channel=2
 ```
+
+You can change the media buffer size (in packets) if your consumers are slow: `buffer=256` (default).
+
+```yaml
+streams:
+  xiaomi1: xiaomi://***&buffer=1024
+```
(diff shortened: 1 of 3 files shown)


wesertyki:
подтверждаю такую же проблему.Что только не перепробывал:и буфер увеличивал и перекодировку аддоном ставил.Бестолку.Когда это исправят в следующей версии?

PR #2399, opened 2026-08-04
Add Wyze Video Doorbell v1 (WYZEDB3) support to wyze source

### LINKED ISSUE #2097: Wyze v1 doorbell?
I've got a probe output for a v1 doorbell.  Can that be used to incorporate it?  I know it's TUTK only.  Worked in wyze bridge and I'd like to abandon the bridge for this!

```
{
  "producers": [
    {
      "id": 44,
      "format_name": "wyze",
      "protocol": "wyze/dtls",
      "remote_addr": "192.168.1.56:52880",
      "source": "wyze://192.168.1.56?dtls=true\u0026enr=W%2FzuuG0fJ6XICYTJ\u0026mac=2CAA8E68578A\u0026model=WYZEC1-JZ\u0026uid=DVAVTN3PSGD72P64111A",
      "medias": [
        "video, recvonly, H264",
        "audio, recvonly, PCMU/8000/1",
        "audio, sendonly, PCMU/8000/1"
      ],
      "receivers": [
        {
          "id": 45,
          "codec": {
            "codec_name": "h264",
            "codec_type": "video",
            "level": 40,
            "profile": "Main"
          },
          "childs": [
            46
          ]
        },
        {
          "id": 47,
          "codec": {
            "channels": 1,
            "codec_name": "pcm_mulaw",
            "codec_type": "audio",
            "sample_rate": 8000
          },
          "childs": [
            48
          ],
          "bytes": 2560,
          "packets": 8
        }
      ],
      "senders": [
        {
          "id": 52,
          "codec": {
            "channels": 1,
            "codec_name": "pcm_mulaw",
            "codec_type": "audio",
            "sample_rate": 8000
          },
          "parent": 50
        }
      ],
      "mixer": {
        "id": 50,
        "codec": {
          "channels": 1,
          "codec_name": "pcm_mulaw",
          "codec_type": "audio",
          "sample_rate": 8000
        },
        "parents": [
          49
        ],
        "childs": [
          52
        ]
      }
    }
  ],
  "consumers": [
    {
      "id": 43,
      "format_name": "probe",
      "protocol": "http",
      "remote_addr": "192.168.1.250:51856",
      "user_agent": "Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/144.0.0.0 Safari/537.36",
      "medias": [
        "video, sendonly, ALL",
        "audio, sendonly, ALL",
        "audio, recvonly, ANY"
      ],
      "bytes_send": 9059,
      "receivers": [
        {
          "id": 49,
          "codec": {
            "codec_name": "ANY",
            "codec_type": ""
          },
          "childs": [
            50
          ]
        }
      ],
      "senders": [
        {
          "id": 46,
          "codec": {
            "codec_name": "h264",
            "codec_type": "video",
            "level": 40,
            "profile": "Main"
          },
          "parent": 45,
          "bytes": 5859,
          "packets": 2
        },
        {
          "id": 48,
          "codec": {
            "channels": 1,
            "codec_name": "pcm_mulaw",
            "codec_type": "audio",
            "sample_rate": 8000
          },
          "parent": 47,
          "bytes": 3200,
          "packets": 10
        }
      ]
    }
  ]
}
```

welovejeff:
## Summary

Adds support for the Wyze Video Doorbell v1 (`WYZEDB3`) to the `wyze://` source. This is a cleaned-up version of #2249 — same three fixes, with the CRLF/whitespace churn removed (that PR shows ~1,300 changed lines; the real diff is 51 lines across 3 files), `gofmt` applied, and **confirmed working against real WYZEDB3 hardware**.

Credit for the original diagnosis and fixes goes to @jdeath (#2249). This PR exists because that one is difficult to review through the line-ending noise and its author no longer has hardware to test on.

## The three fixes

**1. `pkg/tutk/dtls/dtls.go` — DTLS handshake failure (the main bug)**

The client config sets only `CustomCipherSuites` (the custom `TLS_ECDHE_PSK_WITH_CHACHA20_POLY1305_SHA256`) and leaves `CipherSuites` nil. pion then falls back to `defaultCipherSuites()` — six certificate-based ECDHE suites — and because the connection uses PSK with no certificate, pion's auth-type filter drops all six. The resulting ClientHello advertises exactly one cipher suite. Older-firmware cameras such as the WYZEDB3 only accept `TLS_PSK_WITH_AES_128_CBC_SHA256`, find no overlap, and abort with fatal `handshake_failure(40)`:

```
wyze: connect failed: dtls: client handshake failed: handshake error: alert: Alert Fatal: HandshakeFailure
```

Fix: also set `CipherSuites = [TLS_PSK_WITH_AES_128_CBC_SHA256]` on the client path. pion prepends `CustomCipherSuites` to ID-selected suites (`cipher_suite.go`, "Put CustomCipherSuites before ID selected suites"), so newer cameras still negotiate ChaCha20 first — no regression for currently-working models.

**2. `pkg/tutk/frame.go` — 48-byte FrameInfo**

Standard cameras mark the trailing FrameInfo block with `0x0028` (40 bytes); the doorbell uses `0x0030` (48 bytes). The size marker is now carried through `PacketHeader.FrameInfoSize` and used when trimming the payload, instead of assuming 40. Tested behavior confirms the standard 40-byte layout sits at the front of the 48-byte block (codec detection works unchanged).

**3. `pkg/wyze/client.go` — doorbell resolution**

Sends `frameSize=3, bitrate=0xB4` in the K10052 command for `WYZEDB3`. Note: this hardcodes the resolution for this model, so the hd/sd subtype distinction is inert for the doorbell.

## Hardware confirmation

Tested against a wired Wyze Video Doorbell v1 (`WYZEDB3`), firmware `4.25.1.333`:

- DTLS handshake completes (`protocol: wyze/dtls` in `/api/streams`)
- Video negotiates as H264 Main, level 51, 480×640; audio PCM 8 kHz (recvonly + sendonly both present)
- Live video decodes correctly end-to-end (verified via MSE in-browser and via an ffmpeg RTSP consumer; image intact, no artifacts, correct codec detection on every frame)
- Before the patch, the identical config fails 100% of the time with the handshake error above
- Regression check: a Wyze Cam V3 (`WYZE_CAKP2JFUS`) on the same instance continues to stream normally with the patch applied

Happy to run additional tests on this hardware if useful for review.

## Notes

- Base: current `master`. The three touched files are unchanged between v1.9.14 and master, so this applies cleanly to both.
- `gofmt` clean, `go vet ./pkg/tutk/... ./pkg/wyze/...` clean.
- Closes #2097 for `WYZEDB3` specifically (same handshake failure reported there); supersedes #2249 if preferred, though keeping both open until this is verified by a second tester is reasonable.

### DIFF
--- a/pkg/tutk/dtls/dtls.go
+++ b/pkg/tutk/dtls/dtls.go
@@ -28,7 +28,6 @@ func dialDTLS(ctx context.Context, channel uint8, addr net.Addr, writeFn func([]
 
 	var conn *dtls.Conn
 	var err error
-
 	if isServer {
 		conn, err = dtls.Server(adapter, addr, buildDTLSConfig(psk, true))
 	} else {
@@ -68,6 +67,7 @@ func buildDTLSConfig(psk []byte, isServer bool) *dtls.Config {
 	if isServer {
 		config.CipherSuites = []dtls.CipherSuiteID{dtls.TLS_PSK_WITH_AES_128_CBC_SHA256}
 	} else {
+		config.CipherSuites = []dtls.CipherSuiteID{dtls.TLS_PSK_WITH_AES_128_CBC_SHA256}
 		config.CustomCipherSuites = CustomCipherSuites
 	}
 
--- a/pkg/tutk/frame.go
+++ b/pkg/tutk/frame.go
@@ -78,11 +78,18 @@ func (fi *FrameInfo) Channels() uint8 {
 }
 
 func ParseFrameInfo(data []byte) *FrameInfo {
-	if len(data) < frameInfoSize {
+	return parseFrameInfo(data, frameInfoSize)
+}
+
+func parseFrameInfo(data []byte, fiSize int) *FrameInfo {
+	if fiSize < frameInfoSize {
+		fiSize = frameInfoSize
+	}
+	if len(data) < fiSize {
 		return nil
 	}
 
-	offset := len(data) - frameInfoSize
+	offset := len(data) - fiSize
 	fi := data[offset:]
 
 	return &FrameInfo{
@@ -112,14 +119,15 @@ type Packet struct {
 }
 
 type PacketHeader struct {
-	Channel      byte
-	FrameType    byte
-	HeaderSize   int
-	FrameNo      uint32
-	PktIdx       uint16
-	PktTotal     uint16
-	PayloadSize  uint16
-	HasFrameInfo bool
+	Channel       byte
+	FrameType     byte
+	HeaderSize    int
+	FrameNo       uint32
+	PktIdx        uint16
+	PktTotal      uint16
+	PayloadSize   uint16
+	HasFrameInfo  bool
+	FrameInfoSize int // actual FrameInfo size in bytes (may differ from standard 40)
 }
 
 func ParsePacketHeader(data []byte) *PacketHeader {
@@ -150,8 +158,9 @@ func ParsePacketHeader(data []byte) *PacketHeader {
 		hdr.PayloadSize = binary.LittleEndian.Uint16(data[16:])
 		hdr.FrameNo = binary.LittleEndian.Uint32(data[24:])
 
-		if pktIdxOrMarker == 0x0028 && (IsEndFrame(frameType) || hdr.PktTotal == 1) {
+		if isFrameInfoMarker(pktIdxOrMarker, hdr.PktTotal, frameType) {
 			hdr.HasFrameInfo = true
+			hdr.FrameInfoSize = int(pktIdxOrMarker)
 			if hdr.PktTotal > 0 {
 				hdr.PktIdx = hdr.PktTotal - 1
 			}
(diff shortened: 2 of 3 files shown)


welovejeff:
Both review comments addressed:

- `isFrameInfoMarker` now matches `0x0030` explicitly rather than relying on the `>= PktTotal` heuristic, which would indeed have failed for frames spanning more than 48 packets. This also mirrors the existing explicit handling of `0x0028`.
- The doorbell resolution debug log is now gated behind `verbose`, consistent with the other `[Wyze]` logs in the file.

MaybeSomeMoor:
@welovejeff I deleted my comment after I noticed that the stream started working if I leave it running for about 40 seconds.
Your recent patch makes the stream work instantly and smoothly. With the exception of a few things, the resolution is at 640x480, ~~the fps is at 40~~, ~~and the stream stops working after 40 seconds~~.

welovejeff:
@MaybeSomeMoor Thanks for the follow-up — good to hear it comes up instantly now.

The 640x480 is expected, and it isn't your setup. The doorbell branch of `SetResolution` was setting `frameSize`/`bitrate` for WYZEDB3 unconditionally, so the requested quality was ignored and every DB3 got the same stream. That pair is the one combination I found that produces a working stream on mine; the generic frame sizes didn't.

I've changed it so the override only applies on auto — an explicit quality now reaches the camera, so there's something to experiment with if a higher mode works on yours. I've only run it at the default, so I can't say what the firmware accepts beyond that.

MaybeSomeMoor:
@welovejeff Thank you for working on this.
Using a web browser, everything works fine. But when trying RTSP, the resolution is always at 640x480

MaybeSomeMoor:
Adding `&subtype=2` fixed the resolution issue.

PR #2412, opened 2026-08-09
webrtc: prevent duplicate active producer receivers

ucbmckee:
Active WebRTC startup can create duplicate receivers when consumer attachment and negotiated track arrival select equivalent codecs with different RTP payload types. Downstream consumers may remain attached to the idle receiver while packets are delivered to the other one.

- Reuses the existing same-media receiver when codecs differ only by payload type.
- Serializes receiver selection per connection without changing passive producers or packet delivery.

Tests: Added focused regression tests covering both receiver-selection orders, concurrent selection, codec immutability, and media/codec isolation. Verified with `go test -count=1 ./pkg/webrtc` and `go test -race -count=20 ./pkg/webrtc`.

This addresses the duplicate-receiver failure observed when the Nest transport establishes or replaces a WebRTC session with equivalent H.264 payload types.

### DIFF
--- a/pkg/webrtc/conn.go
+++ b/pkg/webrtc/conn.go
@@ -4,6 +4,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"strings"
+	"sync"
 	"time"
 
 	"github.com/AlexxIT/go2rtc/pkg/core"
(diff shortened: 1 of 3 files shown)


ucbmckee:
Related: #2375 reports the same duplicate-receiver failure with Nest WebRTC.

This PR uses a narrower compatibility boundary based on the concern raised in that discussion: it reuses a receiver only for the same media when every codec field except `PayloadType` is equal, rather than matching media alone. It also serializes active-producer receiver selection to prevent concurrent duplicates, keeps the published receiver codec immutable, and adds focused regression coverage for both selection orders, concurrency, codec immutability, and media/codec isolation.

PR #2413, opened 2026-08-10
Add native Reolink/Baichuan support

ucbmckee:
** This was generated by a mix of AI suggestions and human editing **

This PR adds a native `reolink://` source using Reolink's Baichuan protocol over direct LAN TCP or local UID-discovered LAN-P2P. It carries video, audio, and two-way audio directly from the camera, without using RTSP, RTMP, HTTP-FLV, or FFmpeg as the camera input.

One configured source can feed recording, live view, and other go2rtc consumers through the normal fanout. Users configure the camera once, then choose its streams, lenses, audio, and talkback as native features.

## Why it exists

Without a native source, this camera setup needed a continuously running FFmpeg process for each configured input to bridge Reolink's HTTP-FLV feed into go2rtc. Direct Baichuan input removes those processes, reduces CPU use, and makes Reolink streams and talkback part of go2rtc's normal source model.

It also addresses the failures that motivated this work: malformed or incomplete keyframes, timing discontinuities, and downstream decoder/FFmpeg crashes when using full-resolution streams. The adapter validates access units and codec parameter sets, suppresses invalid video, resumes only on a valid keyframe with established parameter sets, and keeps RTP timestamps moving forward across reconnects. In final testing, 4K HEVC segments had the expected frame and keyframe cadence, complete PTS/DTS, no DTS regressions or non-positive video durations, and clean FFmpeg 8.1 strict decoding. FFmpeg 8.1 was used only to validate the output; the source does not invoke FFmpeg or require a particular FFmpeg version. The observed keyframe, timing, and transport-related crash symptoms did not recur in the tested production workload.

## Performance

The prior non-Baichuan setup used go2rtc's generic HTTP-FLV support through one source-side FFmpeg process per configured source, then loopback RTSP. Native Baichuan removes that process. Matched tests on the same cameras and Intel Core i7-12700H host preserved codec, resolution, frame rate, keyframe cadence, and strict-decode results:

| Camera workload | Prior HTTP-FLV/FFmpeg path | Native Baichuan | Attributable change |
|---|---:|---:|---:|
| H.264 sub, 640x360/10 | 2.4 CPU points | 1.8 | 25% lower |
| H.264 extern, 896x512/20 | 3.8 CPU points | 2.9 | 25% lower |
| HEVC main, 3840x2160/20 | 2.4 CPU points | 1.0 | 56% lower |

CPU points are Frigate-reported process CPU attributable to one source path, not whole-host percentages. Results may differ elsewhere, but every matched profile used less CPU.

## Validation

Hardware testing covers E1 Zoom/E340 cameras and both lenses of dual-lens TrackFlex cameras: main and sub streams, TCP and same-LAN UID/LAN-P2P connections, H.264, H.265, AAC, camera ADPCM, and repeated talkback setup and teardown. Audible two-way audio was verified through Frigate on TrackFlex and E1 Zoom cameras. A production system has exercised 16 profiles from six cameras.

Verification includes focused normal and race tests, malformed-input tests, nine fuzz targets with about 12 million final executions, vet, strict FFmpeg decode and timestamp checks, and Linux, Windows, macOS, and FreeBSD builds.

## This implementation versus [fzurita](https://github.com/AlexxIT/go2rtc/pull/2263)

The fzurita branch showed that native Baichuan support in go2rtc is practical. This PR covers the same core streaming and talkback use case while keeping Reolink work isolated from the rest of go2rtc. It also adds stricter safeguards and recovery behavior. The implementations differ at the protocol, lifecycle, and shared-code boundaries, so this is submitted as an independent alternative rather than a patch stack.

The implementation lives in three new packages:

- protocol-only `pkg/baichuan`
- the go2rtc adapter in `pkg/reolink`
- registration in `internal/reolink`

Outside those packages, existing files change only to register and document the source. Shared H.264, H.265, RTP, WebRTC, and stream-lifecycle code is untouched; the pinned fzurita branch changes several of those shared paths. That limits the blast radius: Reolink protocol and recovery changes stay on Reolink camera paths. Configuration, camera sharing, media conversion, talkback, lifecycle, and health checks are separated for review and maintenance. This follows go2rtc's usual thin internal registration plus protocol/adapter package split.

This PR also adds safeguards that are absent from the current fzurita branch. The protocol safeguards do not restrict supported resolutions, bitrates, codecs, or normal go2rtc fanout.

| Area | This PR | fzurita at `fcbd6109` | Why it matters |
|---|---|---|---|
| Camera data | Validates camera-supplied sizes before allocating memory | Allocates camera-declared message sizes; parser memory may grow indefinitely | Bounds memory exposure from malformed or corrupt size fields |
| Network I/O | Tracks requests safely; completes writes with deadlines; reports corruption | Requests may collide; one write without a deadline; drops or resets bad data | Bounds hangs, completes commands, and makes stream damage visible |
| Stream routing | Routes by channel and stream before parsing | Sends each message to every preview before filtering by stream | Reduces extra work and cross-stream interference |
| Credentials | Keeps credentials private and redacted; discards the raw source URL | Public credential fields; retains the parsed source URL | Reduces accidental credential exposure |
| Source lifecycle | Keeps configured sources independent, leaves capacity decisions to the camera, starts talkback on the selected source's existing connection, and waits for cleanup | No cross-source coordination; talkback shares the preview connection | Avoids guessed model limits and cross-source teardown interference |

## Performance versus fzurita

For a direct Baichuan comparison, both implementations were built with Go 1.24 and identical Linux build flags, then tested against the same 3840x2160/20 HEVC camera. Each ran alone in a Debian 13 Proxmox LXC with 8 assigned cores and 16 GiB RAM on the same host.

| 180-second workload | This PR | fzurita |
|---|---:|---:|
| One RTSP-copy viewer | 1.1% CPU; 23.8 MiB RSS | 1.0% CPU; 26.3 MiB RSS |
| Two RTSP-copy viewers | 1.1% CPU; 26.8 MiB RSS | 1.2% CPU; 28.3 MiB RSS |

CPU use was close enough to treat as equivalent at this load. This PR used 2.6 MiB less memory with one viewer and 1.4 MiB less with two. Both produced matched recordings with complete, forward-moving timestamps and no strict-decode errors. These are short, single-camera measurements rather than fleet-wide performance claims.

The feature and engineering differences—not a claimed universal speed advantage—are the main reasons for this implementation.

## Implemented feature coverage

The comparison below uses the current [`fzurita/reolink_protocol@fcbd6109`](https://github.com/fzurita/go2rtc/tree/fcbd6109b1a582f5fbf002b6278f5cec5056299d). This PR remains experimental; checks indicate implemented behavior, not universal camera or firmware support.

| Feature | This PR | fzurita |
|---|:---:|:---:|
| Connect directly by camera IP over LAN TCP | ✅ | ✅ |
| Connect by Reolink UID using LAN-P2P over reliable UDP | ✅ | ✅ |
| Choose the network interface and broadcast address for UID discovery | ✅ | ❌ |
| `main`, `sub`, and `extern` camera streams | ✅ | ✅ |
| Select NVR channels or individual lenses | ✅ | ✅ |
| Detect H.264 or H.265 from the selected stream | ✅ | ✅ |
| Receive AAC or camera ADPCM audio | ✅ | ✅ |
| Two-way audio using common go2rtc audio formats | ✅ | ✅ |
| Turn video or incoming audio off for one source | ✅ | ✅ |
| Create a talkback-only source | ✅ | ✅ |
| Combine main and lower-resolution sources for browser codec fallback | ✅ | ✅ |
| Share one configured source across recording and live viewers | ✅ | ✅ |
| Start talkback without opening another camera connection | ✅ | ✅ |
| Show model, firmware, observed channels, and query status in diagnostics | ✅ | ❌ |
| Suppress invalid video and resume on the next valid keyframe | ✅ | ❌ |
| Reconnect when only one advertised audio or video track stalls | ✅ | ❌ |
| Reconnect when audio remains far below its advertised rate | ✅ | ❌ |

## Unsupported (so far)

- WAN or cloud P2P, relay, NAT traversal, and UPnP; UID/LAN-P2P is local only
- Battery-camera wake and cloud lifecycles
- PTZ, presets, lights, sirens, camera settings, and event APIs
- Automatic RTSP, RTMP, HTTP-FLV, or FFmpeg fallback

Baichuan should be used only on a trusted local network.

## AI disclosure

This implementation was developed with assistance from OpenAI Codex and Anthropic Claude. Its architecture, acceptance criteria, and engineering standards were human-directed. The final branch was reviewed, tested, benchmarked, and exercised on physical cameras. The code was repeatedly reduced and reorganized for reviewability without removing the protocol, safety, lifecycle, or test coverage described above.

### DIFF
--- a/README.md
+++ b/README.md
@@ -194,6 +194,7 @@ A summary table of all modules and features can be found [here](internal/README.
 - [`kasa`](internal/kasa/README.md) - [TP-Link Kasa](https://www.kasasmart.com/) cameras.
 - [`multitrans`](internal/multitrans/README.md) - Two-way audio for Chinese version of [TP-Link](https://www.tp-link.com.cn/) cameras.
 - [`nest`](internal/nest/README.md) - [Google Nest](https://developers.google.com/nest/device-access/supported-devices) cameras through user-unfriendly and paid APIs.
+- [`reolink`](internal/reolink/README.md) - Experimental native Baichuan LAN TCP/UID streaming and two-way audio for Reolink cameras.
 - [`ring`](internal/ring/README.md) - Ring cameras with two-way audio support.
 - [`roborock`](internal/roborock/README.md) - [Roborock](https://roborock.com/) vacuums with cameras with two-way audio support. 
 - [`tapo`](internal/tapo/README.md) - [TP-Link Tapo](https://www.tapo.com/) cameras with two-way audio support.
--- a/internal/README.md
+++ b/internal/README.md
@@ -52,6 +52,7 @@ Some formats and protocols go2rtc supports exclusively. They have no equivalent 
 | [`multitrans`] | `rtp`           | `tcp`            |       |        |        | yes     |
 | [`nest`]       | `srtp`          | `rtsp`, `webrtc` | yes   |        |        | no      |
 | [`onvif`]      | `rtp`           | *                | yes   | yes    |        |         |
+| [`reolink`]    | `baichuan`      | `tcp`, `udp`     | yes   |        |        | yes     |
 | [`ring`]       | `srtp`          | `webrtc`         | yes   |        |        | yes     |
 | [`roborock`]   | `srtp`          | `webrtc`         | yes   |        |        | yes     |
 | [`rtmp`]       | `flv`           | `rtmp`           | yes   | yes    | yes    |         |
@@ -96,6 +97,7 @@ Some formats and protocols go2rtc supports exclusively. They have no equivalent 
 [`ngrok`]: ngrok/README.md
 [`onvif`]: onvif/README.md
 [`pinggy`]: pinggy/README.md
+[`reolink`]: reolink/README.md
 [`ring`]: ring/README.md
 [`roborock`]: roborock/README.md
 [`rtmp`]: rtmp/README.md
--- a/internal/reolink/README.md
+++ b/internal/reolink/README.md
@@ -1,0 +1,135 @@
+# Reolink
+
+Experimental native Baichuan source for Reolink cameras on a trusted local network.
+
+- H.264/H.265 video from `main`, `sub`, and `extern` profiles
+- AAC audio or camera ADPCM converted to PCMA
+- PCMA/PCMU/PCM/PCML two-way audio converted to camera ADPCM
+- Direct LAN TCP or UID-discovered LAN-P2P over reliable UDP
+- Explicit NVR and multi-lens channels
+
+No FFmpeg, RTSP, or HTTP-FLV input bridge is required.
+
+## Compatibility
+
+The adapter discovers the selected profile's codecs from live media and probes talkback directly. Model identity, observed channels, and capability-query status are diagnostic; they do not select a model-specific media path.
+
+Hardware testing covers:
+
+- E1 Zoom/E340 `main` and `sub` over TCP
+- dual-lens TrackFlex `main` and `sub` on both channels over TCP and UID/LAN-P2P
+- H.264, H.265, AAC, and talkback across those tests
+
+Current Reolink specifications place [E1 Outdoor Pro](https://reolink.com/us/product/e1-outdoor-pro/), [Duo 3 PoE](https://support.reolink.com/c/duo-3-poe/), and [TrackMix PoE](https://reolink.com/us/product/reolink-trackmix-poe/) inside the same codec, profile, and talkback envelope. Other likely compatibility candidates include Duo 2 WiFi, Video Doorbell, TrackMix WiFi 6, and RLC-1212A. These are compatibility candidates, not tested support claims for this adapter.
+
+## Configuration
+
+```yaml
+streams:
+  camera_main: reolink://${REOLINK_USER}:${REOLINK_PASSWORD}@camera.local/main
+  camera_sub: reolink://${REOLINK_USER}:${REOLINK_PASSWORD}@camera.local/sub?audio=0&backchannel=0
+  camera_channel: reolink://${REOLINK_USER}:${REOLINK_PASSWORD}@camera.local/main?channel=1
+  camera_uid: reolink://${REOLINK_USER}:${REOLINK_PASSWORD}@${REOLINK_UID}/main?transport=uid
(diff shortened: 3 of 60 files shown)


quenthal:
"Use a separate camera connection for talkback"

Maybe I misintrepet this, but is this beneficial? Less simultaneous connections equals more stable Reolink camera.

ucbmckee:
> "Use a separate camera connection for talkback"
> 
> Maybe I misintrepet this, but is this beneficial? Less simultaneous connections equals more stable Reolink camera.

Good question. A separate connection is not required by the protocol; I chose it to isolate talk setup and teardown from preview, and it permits talk alongside main-stream preview in testing. However, I have not established that it is more stable than sharing the authenticated preview connection and it consumes an additional transient camera session, as you said. Your connection-limit concern is valid. This was me prematurely optimizing. I'm going to test shared-client talk while retaining an independent logical talk lifecycle; unless that exposes a firmware-specific conflict, shared should be the default.

ucbmckee:
> > "Use a separate camera connection for talkback"
> > Maybe I misintrepet this, but is this beneficial? Less simultaneous connections equals more stable Reolink camera.
> 
> Good question. A separate connection is not required by the protocol; I chose it to isolate talk setup and teardown from preview, and it permits talk alongside main-stream preview in testing. However, I have not established that it is more stable than sharing the authenticated preview connection and it consumes an additional transient camera session, as you said. Your connection-limit concern is valid. This was me prematurely optimizing. I'm going to test shared-client talk while retaining an independent logical talk lifecycle; unless that exposes a firmware-specific conflict, shared should be the default.

Change pushed, thanks for the suggestion.

WiredLife:
very nice, but i see one problem after switching from fzurita's version to yours
i'm using the /extern stream for camera substream because of better image quality
with your version the change from main (or on the live page when a still image becomes live substream) to sub causes to a very pixelated image in the beginning

i'm using this settings

> go2rtc:
  streams:
    erdgeschoss:
      - reolink://admin:admin@192.168.11.22/main
      - ffmpeg:erdgeschoss#audio=opus
    erdgeschoss_sub:
      - reolink://admin:admin@192.168.11.22/extern
      - ffmpeg:erdgeschoss_sub#audio=opus

PR #2423, opened 2026-08-14
fix ffmpeg binary path with spaces on Windows

paresy:
The ffmpeg source builds a flat command string ("exec:" + Args.String()) that internal/exec re-parses with shell.QuoteSplit. Since Args.String() wrote the binary path unquoted, a path containing spaces (for example C:\Program Files\ffmpeg\ffmpeg.exe) was split at the first space:

    [api.mjpeg] add consumer error=streams: exec: "C:\Program": executable file not found in %PATH%

Quote the path in Args.String() when it contains a space, the same way filters are already quoted, so QuoteSplit keeps it as one argument.

I tested and verified the fix on Windows and also added a small test.

### DIFF
--- a/pkg/ffmpeg/ffmpeg.go
+++ b/pkg/ffmpeg/ffmpeg.go
@@ -54,7 +54,15 @@ func (a *Args) HasFilters(filters ...string) bool {
 func (a *Args) String() string {
 	b := bytes.NewBuffer(make([]byte, 0, 512))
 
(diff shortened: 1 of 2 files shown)


paresy:
I just wanted to ask quickly if i can help land this small fix somehow? Any more tests or verification steps i can do or supply?

Thanks for looking into this!

Issue #2232, opened 2026-04-25
RTSP authentication fails


johnchikov:
go2rtc 1.19.14 (arm\linux)

IP Camera HikVision DS-2CD1148-I/B
RTSP Authentication - Basic

```
go2rtc config:
	streams:
		camera1: "rtsp://user:password@192.168.1.64/Streaming/Channels/101"
```

go2rtc can't authentication.

TCP dump:

```
DESCRIBE rtsp://192.168.1.64/Streaming/Channels/101 RTSP/1.0
CSeq: 1
Accept: application/sdp
User-Agent: go2rtc/1.9.14

RTSP/1.0 401 Unauthorized
CSeq: 1
WWW-Authenticate: Digest realm="b4a38201a73f", nonce="739ee6c7ea93532ef87d5f319b155537", stale="FALSE"
WWW-Authenticate: Basic realm="b4a38201a73f"
Date:  Sat, Apr 25 2026 00:05:48 GMT

DESCRIBE rtsp://192.168.1.64/Streaming/Channels/101 RTSP/1.0
Authorization: Digest username="user", realm="b4a38201a73f", nonce="739ee6c7ea93532ef87d5f319b155537", uri="rtsp://192.168.1.64/Streaming/Channels/101", response="a05e3b9c5e302373ff280255389337a2"
Accept: application/sdp
User-Agent: go2rtc/1.9.14
CSeq: 2

RTSP/1.0 401 Unauthorized
CSeq: 2
WWW-Authenticate: Digest realm="b4a38201a73f", nonce="2a42927df1739be4565e0f8253075f46", stale="FALSE"
WWW-Authenticate: Basic realm="b4a38201a73f"
Date:  Sat, Apr 25 2026 00:05:48 GMT
```

I wrote JS-script and calc **response** and I get another result.

As I understand it, go2rtc uses the ready-made built-in net/http module.
Why isn't the **response** being calculated correctly?

When running ffmpeg (without go2rtc) using the same address, it connects correctly.

AlexxIT:
As far as I can see, there's an incorrect protocol implementation on the camera side. There are two Authenticate headers, which shouldn't be there.

johnchikov:
I apologize, the problem was caused by incorrect username/password entered in go2rtc.ymal

PR #2451, opened 2026-08-25
Add RFC 7616 algorithm negotiation to RTSP digest auth

david-saint:
RTSP digest authentication currently reads only the first `WWW-Authenticate` header (`Header.Get`) and always hashes HA1/HA2/response with MD5. It never parses the challenge's `algorithm=` directive.

RFC 7616 allows a server to offer MD5, MD5-sess, SHA-256, and SHA-256-sess, including as separate `WWW-Authenticate` values. This change:

- Reads every `WWW-Authenticate` challenge instead of only the first header
- Selects the strongest supported algorithm
- Hashes HA1/HA2/response with that algorithm, including `-sess` and `qop=auth` when offered
- When the challenge omits `algorithm` and `qop`, emits the previous five-field `Authorization` header unchanged

`go test ./pkg/tcp/...` covers MD5 and SHA-256 vectors, algorithm ranking, dual-header selection in both orders, and the legacy header shape.

#2232 independently noted two Authenticate headers on the wire; that report itself was misconfigured credentials, not this defect.

### DIFF
--- a/pkg/tcp/auth.go
+++ b/pkg/tcp/auth.go
@@ -2,19 +2,35 @@ package tcp
 
 import (
 	"crypto/md5"
+	"crypto/sha256"
 	"encoding/base64"
 	"encoding/hex"
 	"fmt"
 	"net/url"
+	"regexp"
 	"strings"
+
+	"github.com/AlexxIT/go2rtc/pkg/core"
 )
 
 type Auth struct {
-	Method  byte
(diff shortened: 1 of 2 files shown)


bobvandevijver:
I tried this implementation with my Shelly Camera with authentication (#2476), and it seems to work 👍🏻

PR #2480, opened 2026-09-10
webrtc: reuse a compatible Receiver when the remote transmits on another payload type (Nest cold-consumer starvation)

### LINKED ISSUE #2386: Nest source: duplicate H264 receiver track on every WebRTC negotiation, RTSP output wired to the empty one
## Environment
  - go2rtc version 1.9.10 (df95ce3) linux/amd64
  - Bundled inside `ghcr.io/blakeblackshear/frigate:stable`, Debian 12 base
  - Host: Unraid, container network mode `host`
  - Camera: Nest Doorbell Gen 2 (wired), connected via the built-in `nest:` source (Google SDM API / Device Access), no Home Assistant
  involved
  - Consumer: Frigate's ffmpeg pulls from go2rtc over local RTSP (`preset-rtsp-restream`, `-rtsp_transport tcp`)

  ## Symptom
  Every single WebRTC negotiation between go2rtc and the Nest SDM API for this doorbell results in **two separate H264 receiver
  tracks** inside the same producer:
  
  - One receiver has `childs` (i.e. it is wired to the RTSP consumer's sender) but **never receives any bytes** (the
  `bytes`/`packets` fields are absent, i.e. zero, for the entire life of the producer).
  - A second, separate H264 receiver **is not wired to any consumer** but receives all the real incoming video data (bytes/packets
  grow continuously).
  
  Because the RTSP output is bound to the empty receiver, Frigate's ffmpeg never gets usable video and fails with:
  [rtsp @ ...] Could not find codec parameters for stream 0 (Video: h264, none): unspecified size
  Consider increasing the value for the 'analyzeduration' (0) and 'probesize' (5000000) options
  [out#0/segment @ ...] Output file does not contain any stream
  This triggers Frigate's watchdog to kill and restart ffmpeg every ~20s, which in turn causes go2rtc to tear down and
  re-negotiate a brand new WebRTC session with Nest every ~20s — a continuous reconnect loop that has apparently been running for
  weeks.
  
  ## Reproduction rate
  Observed 6 consecutive fresh negotiations (passively, driven by Frigate's ffmpeg reconnect loop, no manual intervention) over
  ~90 seconds: **6/6 showed the same wrong wiring**. Also reproduced immediately after a manual restart of the go2rtc service
  (2/2). This looks fully deterministic for this device, not a rare race.
  
  ## Example `/api/streams` output (credentials redacted)
  ```json
  {
    "doorbell_nest": {
      "producers": [{
        "id": 194, 
        "format_name": "nest/webrtc",
        "protocol": "http+udp",
        "remote_addr": "74.125.247.225:19305 host",
        "url": "nest:?client_id=REDACTED&client_secret=REDACTED&refresh_token=REDACTED&project_id=REDACTED&device_id=REDACTED",
        "medias": ["video, recvonly, H264", "audio, recvonly, OPUS/48000/2"],
        "receivers": [
          {"id": 195, "codec": {"codec_name": "h264", "codec_type": "video"}, "childs": [196]},
          {"id": 197, "codec": {"channels": 2, "codec_name": "opus", "codec_type": "audio", "sample_rate": 48000}, "bytes": 1694,
  "packets": 7},
          {"id": 198, "codec": {"codec_name": "h264", "codec_type": "video"}, "bytes": 9152, "packets": 9}
        ],
        "bytes_recv": 11230
      }],
      "consumers": [{
        "id": 193, 
        "format_name": "rtsp",
        "protocol": "rtsp+tcp",
        "senders": [{"id": 196, "codec": {"codec_name": "h264", "codec_type": "video"}, "parààà``ent": 195}]
      }]
    }
  }
```
  Note receiver 195 (parent of the RTSP sender 196) never appears with a bytes field across the entire session, while sibling
  receiver 198 (same producer, same codec, not wired to anything) accumulates bytes/packets continuously. This pattern repeats
  identically on every negotiation, only the IDs change.

## Suspected root cause

  Reading pkg/webrtc/conn.go, track binding in OnTrack matches purely on media.ID == transceiver.Mid() plus codec.PayloadType == 
  remote.PayloadType(). My guess is Nest's SDP answer includes a retransmission (RTX) SSRC/payload type alongside the primary H264
  payload for the same m=video line (standard WebRTC reliability mechanism, a=fmtp:<rtx-pt> apt=<primary-pt>), and this matching
  logic doesn't correlate the RTX SSRC back to the primary track — so Pion fires two independent OnTrack events, and go2rtc
  creates two independent receivers instead of one. Whichever one is created/bound first ends up wired to the consumer, and
  empirically that's always the wrong one for this device (I have not confirmed which of the two is the RTX stream vs. primary
  from the logs alone — happy to capture raw SDP if that helps debug).

### LINKED ISSUE #2389: WebRTC producer creates two receivers for one video codec; consumers attach to the one that is never fed
> **Update:** [#2375](https://github.com/AlexxIT/go2rtc/pull/2375) (@misecichin) already fixes this upstream, and is the better place for it — it makes `GetTrack` fall back to matching on `Media` when the `*Codec` pointer lookup misses, which is the same root cause described below. I filed this before finding that PR. [#2386](https://github.com/AlexxIT/go2rtc/issues/2386) is a duplicate report of the same behaviour. Leaving this open only for the analysis and the production traces; close it if you'd rather consolidate on #2375.

## Summary

A WebRTC producer can end up with **two receivers for the same video codec**, only one of which is ever fed, while consumers are attached to the other. The stream then delivers audio and no video, indefinitely, with nothing logged.

`getMediaCodec` resolves the codec from the payload type actually being transmitted, but `media.Codecs` is narrowed to that entry *after* `GetTrack` runs. A consumer that wired up earlier matched the **first** codec in the answer, and `GetTrack` compares receivers by `*Codec` **pointer** identity — so the two disagree and a second receiver is created.

## Where

`pkg/webrtc/conn.go`, in `pc.OnTrack` (line numbers from master `c245815`):

```go
media, codec := c.getMediaCodec(remote)   // codec = the PayloadType actually transmitted
if media == nil {
    return
}

track, err := c.GetTrack(media, codec)    // pointer-identity lookup -> may create a 2nd receiver
if err != nil {
    return
}

switch c.Mode {
case core.ModePassiveProducer, core.ModeActiveProducer:
    // replace the theoretical list of codecs with the actual list of codecs
    if len(media.Codecs) > 1 {
        media.Codecs = []*core.Codec{codec}   // narrowing happens AFTER GetTrack
    }
}
```

`GetTrack` here is `Conn.GetTrack` (`pkg/webrtc/producer.go:8`), which shadows `core.Connection.GetTrack`; both compare by pointer:

```go
for _, track := range c.Receivers {
    if track.Codec == codec {
        return track, nil
    }
}
```

Meanwhile `Codec.Match` (`pkg/core/codec.go:126`) compares by name/clock-rate/channels and deliberately ignores payload type — so the codebase matches by **value** in one place and by **pointer** in another.

## Observed

Nest producer (`nest/webrtc`), from `/api/streams` on a live instance:

```
receiver 34852  childs=true   bytes=0        profile=<nil>   <- RTSP consumer's video sender is parented here
receiver 34856  childs=false  bytes=3546841  profile=Main    <- OnTrack feeds this one
rtsp video sender  parent=34852  bytes=None
rtsp audio sender  parent=34854  bytes=55349
```

The `profile=<nil>` receiver is the one built from the negotiated SDP entry; the `profile=Main` one is built by `OnTrack` from the actual track. 37 samples over 90s showed the same split on two different cameras.

Audio is unaffected because there is a single Opus payload type and therefore no ambiguity. That is what makes the failure so quiet for an ffmpeg consumer: audio keeps arriving, so a socket timeout never fires, and `-movflags frag_keyframe` cannot cut a fragment without video keyframes. The reader stays alive and silent with nothing on stderr.

## It is deterministic, not intermittent

A **cold** dial reliably mis-binds: the consumer's `AddConsumer` → `MatchMedia` → `GetTrack` runs microseconds after `Dial` returns, while `OnTrack` fires only once the first RTP packet arrives — so the consumer picks `media.Codecs[0]` before the narrowing.

A **warm** attach reliably binds correctly, because by then `media.Codecs` has been narrowed to the transmitted codec.

That matches the field behaviour exactly: a camera is healthy for hours, then breaks the moment everything re-dials at once (a restart), and one snapshot request against the stream before the consumer attaches is enough to make the next attach correct.

`Producer.reconnect` (`internal/streams/producer.go:206`) pre-seeds the new connection's receiver with `MatchCodec`, which also picks the first name-matching entry — so a reconnect can re-roll the same mis-bind.

## Possible fix

Adopting the transmitted codec into the receiver consumers are already on, before `GetTrack`, removes the duplicate:

```go
for _, recv := range c.Receivers {
    if recv.Media == media && recv.Codec != codec && recv.Codec.Match(codec) {
        recv.Codec = codec
        break
    }
}
```

`GetTrack` then finds that receiver by identity and feeds the consumers attached to it. The reverse (moving senders onto the newly created receiver) looks wrong: that receiver is absent from `Producer.receivers`, so `stopProducers` (`internal/streams/stream.go`) would see no attached senders and stop a producer that still has live consumers.

Measured on the deployment above, cold start with no warm-up:

| | video receivers | consumer bound to | result |
|---|---|---|---|
| before | 2 | the starved one (`bytes=0`) | no video, reader killed on a ~22s loop |
| after | 1 | the fed one (16,394,290 bytes) | video flowing, zero kills |

I am running this on a fork, scoped to `nest/webrtc` to bound the blast radius. I have **not** proposed it as a PR because the gate is wrong for upstream — the pointer-vs-value mismatch is generic to any producer whose answer keeps several codecs on one media — and because placing it inside the `OnTrack` closure makes it awkward to test. If you would like it as a PR, I am happy to rework it ungated (perhaps inside `Conn.GetTrack` so it also covers the reconnect pre-seed path) and with a test.

I may also be missing a reason the two receivers are intended, in which case the narrowing ordering is probably the more interesting question.

## Environment

- Observed on v1.9.14; the code path is unchanged on master `c245815` (verified against source).
- Producer `nest/webrtc`; consumers RTSP (ffmpeg) and the internal preload.
- Linux/arm64, Docker, Raspberry Pi 4.

rkj:
## Problem

A remote peer may answer one media with several payload types and then transmit on one that is not the first. Google Nest (SDM) does this for H264: go2rtc's offer lists three H264 payload types (96, 97, 98); Google's answer keeps 96 and 98 and the camera streams on 98 (High profile, `profile-level-id=64001f`).

`Conn.GetTrack` deduplicates receivers by `*Codec` pointer. A consumer that attaches **before the first RTP packet** already holds a `Receiver` for the first codec of the media. When `OnTrack` fires, `getMediaCodec` resolves the transmitted payload type to a different `*Codec`, the exact lookup misses, and a second `Receiver` is created that no consumer is wired to. Every cold consumer starves forever; a consumer that attaches later binds to the fed receiver and works.

This is #2389 and #2386 (and the "open it in VLC first, then Frigate works" reports in the Frigate discussions). It is dial-order dependent, which is why it looks intermittent.

## Fix

In the producer branch of `Conn.GetTrack`, before creating a new `Receiver`, reuse an existing one for the same `Media` whose codec is compatible (`Codec.Match`: same name, clock rate and channels) and adopt the transmitted codec on it. A switch to a genuinely different codec (e.g. H264 to VP8) still falls through to the create path, as today. This is the same mechanism as #2375, with the codec-compatibility guard suggested in that PR's review, and it also updates the receiver's codec so `/api/streams` and later consumers see the payload type actually in use.

## Measurements (first-generation Nest Doorbell, wired, `nest:` source, host networking)

Test: consumer A (`/api/stream.mp4`) attaches at dial time and holds 40 s; consumer B attaches 12 s later for 20 s.

| build | A (cold) | B (warm) | video receivers in `/api/streams` |
|---|---|---|---|
| v1.9.10 unpatched | 0 bytes in 40 s | 3.1 MB, first byte 1.2 s | two: id 3 with `childs` and no bytes, id 6 with all bytes and no children |
| master (c245815) + this patch | 2.3 MB, first byte 6.2 s | 0.9 MB, first byte 0.13 s | one, with the consumer as its child |

The 6 s first byte on the cold consumer is ICE setup plus the wait for the first keyframe; this camera sends SPS+IDR from the first packet and every 2 to 3 s after, so no keyframe request was needed for this test.

Nothing else changes for producers that answer with a single payload type: the exact-pointer lookup still hits first.

Fixes #2389, fixes #2386.

### DIFF
--- a/pkg/webrtc/producer.go
+++ b/pkg/webrtc/producer.go
@@ -33,6 +33,21 @@ func (c *Conn) GetTrack(media *core.Media, codec *core.Codec) (*core.Receiver, e
 	case core.ModePassiveProducer, core.ModeActiveProducer:
 		// Passive producers: OBS Studio via WHIP or Browser
 		// Active producers: go2rtc as WebRTC client or WebTorrent
+		//
+		// The remote may answer one media with several payload types and then
+		// transmit on one that is not the first (Google Nest/SDM answers an
+		// H264 offer with two payload types and streams on the second). A
+		// consumer that attached before the first RTP packet already owns a
+		// Receiver for the first codec, so OnTrack must reuse it instead of
+		// creating a second Receiver that no consumer is wired to. Only a
+		// compatible codec is adopted (same name, clock rate and channels);
+		// a switch to a different codec still gets its own Receiver.
+		for _, track := range c.Receivers {
+			if track.Media == media && track.Codec.Match(codec) {
+				track.Codec = codec
+				return track, nil
+			}
+		}
 
 	default:
 		panic(core.Caller())


ajplotkin:
Confirming from a second deployment: this is the mechanism I have been running
(gated to `nest/webrtc`, inside `OnTrack`) since late July, with the same result.

One correction to #2375's framing, since it matters for how this reads. On
master, `Conn.GetTrack` matches `track.Codec == codec` by pointer. A cold
consumer goes through `MatchMedia`, which returns the first producer codec
(PT 96), while pion fires the handler once after peeking the first packet, by
which point the track carries PT 98. The pointer misses, a second receiver is
created, and the consumer binds to the one that never receives. #2375 describes
this as Nest switching PT mid-session; pion handles a genuine mid-stream payload
change in place on the same `TrackRemote` (`track_remote.go`,
`checkAndUpdateTrack`) without a second `OnTrack`, and go2rtc's read loop
resolves the track once, so that path would keep feeding the same receiver. The
fault is bind order. Your loop sits exactly at the miss.

The `track.Codec = codec` adoption also keeps `Producer.GetTrack`'s pointer cache
hitting for consumers that arrive after `media.Codecs` is narrowed; without it
each of those re-enters `conn.GetTrack` and appends the same receiver to
`p.receivers` again. Harmless, but the description could mention it.

#2375 targets `dev` and this targets `master`; both insert at the same lines of
`producer.go`, so whichever lands second will conflict. Saying in the description
that this supersedes #2375 (the guard from that review, plus the adoption from
#2389) would save the maintainer working it out.

A test, if you want it — I offered one on #2389 and never wrote it, and I'd
rather send it against your branch than open a third PR. Build a `Conn` in
`ModeActiveProducer` with one `Media` (`DirectionRecvonly`, or `core.Assert`
panics) carrying two H264 codecs, call `GetTrack(media, codecs[0])` then
`GetTrack(media, codecs[1])`, and assert one receiver with
`Receivers[0].Codec == codecs[1]`. That assertion fails on master. A VP8 negative
case asserting two receivers passes on master and guards the `Match` bound. The
producer branch touches only `c.Mode` and `c.Receivers`, so it needs no
`PeerConnection`.

Separately: `track.Codec` is written from the `OnTrack` goroutine while
`Producer.GetTrack` reads it under `p.mu`. That race predates this change
(`c.Receivers = append(...)` already has it), so I wouldn't block on it, but it
should be mentioned.

